PR #1, opened 2024-08-19
"Hello world" Preservation and Storage API

donaldgray:
Mostly bootstrapping for now. Storage and Preservation API only (UI will be added shortly).

Overview:
* Single dotnet solution with 4 projects, as detailed in readme (+ test projects)
* Preservation API has a `/storage` endpoint that will call Storage API and return with a boolean to show success/failure
* Both apps have:
  * Serilog configured
  * Correlation Id - first service called will append if not found + propagate to downstream requests.
  * Basic test projects to configure `WebApplicationFactory` for integration tests.


donaldgray:
Merging now, this is boilerplate with little functionality.

PR #15, opened 2024-12-11
Add Disable Auth Feature Flag

frank-kerrigan:
Add Auth0 feature flag so it can run without Auth.

The feature flag does not have to added to appsettings as it will default to Auth Enabled.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.UI/Program.cs
+++ b/src/DigitalPreservation/DigitalPreservation.UI/Program.cs
@@ -19,8 +19,6 @@ Log.Information("Application starting..");
 
 try
 {
-
-
     var builder = WebApplication.CreateBuilder(args);
     builder.Host.UseSerilog((hostContext, loggerConfiguration)
         => loggerConfiguration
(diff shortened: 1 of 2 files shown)


src/DigitalPreservation/DigitalPreservation.UI/Program.cs
```diff
@@ -29,24 +27,32 @@
...
-    builder.Services.AddRazorPages().AddMvcOptions(options =>
-    {
-        var policy = new AuthorizationPolicyBuilder()
-            .RequireAuthenticatedUser()
-            .Build();
-        options.Filters.Add(new AuthorizeFilter(policy));
-    }).AddMicrosoftIdentityUI();
+
+        // <ms_docref_add_default_controller_for_sign-in-out>
+        builder.Services.AddRazorPages().AddMvcOptions(options =>
```
donaldgray, line 48: ~Do you still need to call `.AddRazorPages()` regardless of whether there is auth or not?~

Edit - `.AddRazorPages()` is called at line 70 - might be worth the call in 1 place only?

src/DigitalPreservation/DigitalPreservation.UI/Program.cs
```diff
@@ -88,9 +94,12 @@
     app.MapControllers();
     app.UseHealthChecks("/health");
     //Authentication
-    app.UseAuthentication();
-    app.UseAuthorization();
-   
+    if (useAuthFeatureFlag)
+    {
+        app.UseAuthentication();
+        app.UseAuthorization();
```
donaldgray: `.UseAuthorization()` is called regardless as it's on line 91

donaldgray (approved):

src/DigitalPreservation/DigitalPreservation.UI/Program.cs
frank-kerrigan: fixed

src/DigitalPreservation/DigitalPreservation.UI/Program.cs
frank-kerrigan: yes, removed line

PR #16, opened 2024-12-17
Feature/auth all the way

tomcrane:
This writes `callerIdentity` into Fedora.

We now have potentially more than one RDF statement to make in a request body, so there's a new extension method `AppendRdf` tested here:

https://github.com/uol-dlip/digital-preservation/blob/844a8c8343a8686ac6e5ca7cd33990eb958eeaf4/src/DigitalPreservation/Storage.API.Tests/Fedora/RdfTests.cs#L5

At the moment, the IFedoraClient write operations take an explicit `callerIdentity` parameter, but everywhere else the ClaimsPrincipal is pulled out of either the ASP.NET Core `User` or `ClaimsPrincipal.Current` (e.g., in a Mediatr handler).

This extension method pulls out our single "slug" string to use for the caller identity.

https://github.com/uol-dlip/digital-preservation/blob/844a8c8343a8686ac6e5ca7cd33990eb958eeaf4/src/DigitalPreservation/DigitalPreservation.Core/Auth/ClaimsPrincipalX.cs#L7

However, nothing is actually propagating that identity up the chain yet, so the implementation is fake.

An additional complication - when in import job is executed, it's pulled off a queue by a separate process. So  `ClaimsPrincipal.Current` is not going to be the user who created the job. In this case, the callerIdentity is obtained from the stored import job:

https://github.com/uol-dlip/digital-preservation/blob/844a8c8343a8686ac6e5ca7cd33990eb958eeaf4/src/DigitalPreservation/Storage.API/Features/Import/Requests/ExecuteImportJob.cs#L26

This PR is a halfway house - still need to get the real identity moving up the API call chain!


src/DigitalPreservation/DigitalPreservation.Core/Web/Headers/PropagateCorrelationIdHandler.cs
```diff
@@ -46,6 +46,16 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
             AddCorrelationId(request, headerValue);
         }
 
+
+        //Pass Bearer token to downstream API 
+        var bearerToken = contextAccessor.HttpContext.TryGetHeaderValue("Authorization");
```
donaldgray, line 51: This logic feels like it'd be better suited to a different location, maybe a specific `DelegatingHandler` that's only added to `HttpClients` that make requests to either the Storage or Preservation API - is that what `AuthTokenInjector` is doing?)

The `PropagateCorrelationIdHandler` is configured to run on _every_ outgoing http request so could end up sending bearer token to somewhere that shouldn't receive it.

src/DigitalPreservation/DigitalPreservation.UI/Infrastructure/SessionTimeoutAsyncPageFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+            var session = context.HttpContext.Session;
+            if (!session.Keys.Any())
+            {
+                //try and refresh the request path if possible
+                context.HttpContext.Response.Redirect($"/Account/RefreshLogin?path={context.HttpContext.Request.Path}");
+                return;
+            }
+        }
+        catch (Exception e)
+        {
```
donaldgray, line 25: Worth logging anything here?

src/DigitalPreservation/DigitalPreservation.UI/Pages/Browse.cshtml.cs
```diff
@@ -4,20 +4,26 @@
...
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Web;
 
 namespace DigitalPreservation.UI.Pages;
 
-public class BrowseModel(IMediator mediator) : PageModel
+public class BrowseModel(IMediator mediator, ITokenAcquisition tokenAcquisition) : PageModel
```
donaldgray: Unused arg?

src/DigitalPreservation/Preservation.Client/AuthTokenInjector.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        try
+        {
+            var scopes = tokenScope?.ScopeUri?.Split(" ") ?? [];
+            var accessToken = await tokenAcquisition.GetAccessTokenForUserAsync(scopes);
+            return accessToken;
+        }
+        catch (Exception e)
+        {
+            logger.LogError(e, $"AuthTokenInjector: Failed to to Bearer Token: {e.Message}");
```
donaldgray: Serilog is the configured logger, which supports structured logging so best to pass parameters rather than interpolated strings

```cs
logger.LogError(e, "AuthTokenInjector: Failed to to Bearer Token: {Message}", e.Message);
```

Although I think this could be shortened:
* The `e` passed as first arg will already be logging the `.Message` along with the `.Stacktrace`
* `"AuthTokenInjector:"` prefix is redundant as it'll already be in log as Serilog is configured with `.Enrich.FromLogContext()` and has `{SourceContext}` in the output template (the `{SourceContext}` will be the calling class)

src/DigitalPreservation/Storage.API/Features/Import/Requests/ExecuteImportJob.cs
```diff
@@ -22,6 +23,7 @@ public class ExecuteImportJobHandler(
     public async Task<Result<ImportJobResult>> Handle(ExecuteImportJob request, CancellationToken cancellationToken)
     {
         var importJob = request.ImportJob;
+        var callerIdentity = importJob.CreatedBy!.GetSlug()!;
```
donaldgray, line 26: `callerIdentity` is passed a lot of times - I wonder if there could be something that could wrap that (but would just be easier to pass the name)?

```cs
var transaction = await fedoraClient.BeginTransaction();

using var scopedFedoraClient = fedoraClient.WithIdentity(callerIdentity);
// maybe .WithTransaction(transaction); too??

var fedoraContainerResult = await scopedFedoraClient.CreateContainerWithinArchivalGroup(...);

var fedoraPutBinaryResult = await scopedFedoraClient.PutBinary(...);
```

src/DigitalPreservation/Storage.API/Features/Import/Requests/QueueImportJob.cs
```diff
@@ -21,6 +22,11 @@ public class QueueImportJobHandler(
 {
     public async Task<Result<ImportJobResult>> Handle(QueueImportJob request, CancellationToken cancellationToken)
     {
+        if (request.ImportJob.CreatedBy == null)
+        {
```
donaldgray, line 26: Log this? Or will it be logged higher up?

src/DigitalPreservation/Storage.API/Fedora/Http/RdfBodyBuilder.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System.Net.Http.Headers;
+using System.Text;
+using System.Text.RegularExpressions;
+using DigitalPreservation.Utils;
+
+namespace Storage.API.Fedora.Http;
+
+public static partial class RdfBodyBuilder
```
donaldgray, line 8: 💪

src/DigitalPreservation/Storage.API/Fedora/Http/RdfBodyBuilder.cs
```diff
@@ -0,0 +1,65 @@
...
+        requestMessage.Content = new StringContent(content, Turtle);
+        
+    }
+
+    private static string FormatPrefix(string prefix, string uri)
+    {
+        return $"PREFIX {prefix}: <{uri}>";
+    }
+
+    [GeneratedRegex("PREFIX ([a-z]*): <(http.*)>", RegexOptions.Multiline)]
```
donaldgray, line 63: TIL `[GeneratedRegex]`, nice

src/DigitalPreservation/Storage.API/Features/Import/Requests/ExecuteImportJob.cs
tomcrane: I did think it could get it from httpContext.request, but there won't always be an httpContext (e.g., running in a process picking up jobs). I will look into the `.WithIdentity` approach though... what does that return? a subclass?

donaldgray: Yeah a subclass that has an identity prop OR the same class where `.WithXX` sets a prop and it's cleared on Dispose? I hadn't given the actual implementation a huge amount of thought, was more thinking if it would save passing the same value to a number of methods that are all called in the same scope. When I've done similar in the past I've returned something that implements `IDisposable` - doesn't necessarily need to do something on dispose but it's just to keep it scoped to current set of operations.

```cs
// Completely different class that's scoped with identity
public class IdentifyingFedoraClient(string callerIdentity) : IFedoraClient, IDisposable
{
  public async Task<Result<Binary?>> PutBinary(Binary binary, Transaction transaction)
  {
     // this is the bit in existing code that uses callerIdentity
    var patchReq = MakeHttpRequestMessage(metadataUri, HttpMethod.Patch)
        .InTransaction(transaction);
    var isCreation = binaryResponse.CreatedBy == fedoraOptions.Value.AdminUser;
    patchReq.AsInsertTitlePatch(binary.Name!, this.callerIdentity, isCreation);
  }
}

public class FedoraClientX
{
    public async IFedoraClient WithIdentity(string callerIdentity) => 
        new IdentifyingFedoraClient(callerIdentity, other, params,...);
}
```

OR 

```cs
public class FedoraClient(..existing..) : IFedoraClient, IDisposable
{
  // Just set a prop then clear it down on Dispose
  public FedoraClient WithIdentity(string identity)
  {
    this.callerIdentity = identity;
    return this;
  }

  public void Dispose()
  {
    this.callerIdentity = null;
  }

  public async Task<Result<Binary?>> PutBinary(Binary binary, Transaction transaction)
  {
     // this is the bit in existing code that uses callerIdentity
    var patchReq = MakeHttpRequestMessage(metadataUri, HttpMethod.Patch)
        .InTransaction(transaction);
    var isCreation = binaryResponse.CreatedBy == fedoraOptions.Value.AdminUser;
    patchReq.AsInsertTitlePatch(binary.Name!, this.callerIdentity, isCreation);
  }
}
```

It might not be worth it just to avoid passing a string to X number of methods, maybe the effort would only be worth it if more than `identity` was scoped - only other example I could think of was Transaction. 

If this approach was used it'd maybe require a `FedoraClientFactory` or something to generate the class that's used.

src/DigitalPreservation/DigitalPreservation.UI/Infrastructure/SessionTimeoutAsyncPageFilter.cs
tomcrane: I have added a debug logging message

src/DigitalPreservation/DigitalPreservation.Core/Web/Headers/PropagateCorrelationIdHandler.cs
tomcrane: To be addressed

src/DigitalPreservation/DigitalPreservation.UI/Pages/Browse.cshtml.cs
tomcrane: Something odd about this. It appears as a greyed-out unused arg in Rider at first, but after a while it changes (I didn't make any code changes)

image

My other question is - if this is needed here, isn't it needed in every other PageModel and Controller that calls the API?

tomcrane: I'll try re-running tomorrow (and not on the train)

src/DigitalPreservation/Preservation.Client/AuthTokenInjector.cs
tomcrane: (edited)

src/DigitalPreservation/Storage.API/Features/Import/Requests/ExecuteImportJob.cs
tomcrane: I want to come back to this, I have created an issue for it.

src/DigitalPreservation/Storage.API/Features/Import/Requests/QueueImportJob.cs
tomcrane: logged as an error

src/DigitalPreservation/DigitalPreservation.Core/Web/Headers/CorrelationIdMiddleware.cs
```diff
@@ -1,4 +1,4 @@
-namespace DigitalPreservation.Core.Web.Headers;
+﻿namespace DigitalPreservation.Core.Web.Headers;
```
tomcrane, line 1: The unit tests on this are failing

src/DigitalPreservation/DigitalPreservation.UI/Pages/Browse.cshtml.cs
frank-kerrigan: opps that was me in early stages of development. That ref has been removed.

donaldgray (approved):

PR #22, opened 2025-03-10
Feature/storage activity stream

tomcrane:
This PR adds an Activity Stream to the Storage API that reports on completed import jobs, and an Activity Stream to the Preservation API that uses the first stream to publish an Archival Group activity stream.

Background:
https://iiif.io/api/discovery/1.0/

We are using these in Preservation to publish events about ImportJobs and ArchivalGroups, rather than IIIF Manifests. 

The classes in DigitalPreservation.Common.Model.ChangeDiscovery namespace could be replaced by referencing a very similar implementation in iiif-net nuget package.

src/DigitalPreservation/DigitalPreservation.Common.Model/ChangeDiscovery/Reader/ActivityStreamReader.cs is a very light implementation of the IIIF change discovery spec algorithm.

The Storage API gains three extra columns in its ImportJobs table, to record start and end time of a job in the DB row rather than only in the serialised JSON - we need to be fast to generate the activity stream data.

The Preservation API gains a new table to log the events that it learns from the Storage stream, and generate its own ArchivalGroup Activity Stream - src/DigitalPreservation/Preservation.API/Data/Entities/ArchivalGroupEvent.cs.

The latter has one row of seed data to provide a start date. We read the storage API stream back until the start date.

Both API projects have a new ActivityController, with similar looking implementations, that generates the OrderedCollection and OrderedCollectionPage responses.
src/DigitalPreservation/Preservation.API/Features/Activity/ActivityController.cs
src/DigitalPreservation/Storage.API/Features/Activity/ActivityController.cs

src/DigitalPreservation/Preservation.API/Features/Activity/Readers/StorageImportJobsProcessor.cs is what turns raw info from ImportJobs from the storage API stream into Archival Group events. It runs in a backgroundService, every minute (or every 30 mins if it's failing): 
src/DigitalPreservation/Preservation.API/Features/Activity/Readers/StorageImportJobsService.cs


src/DigitalPreservation/Storage.API/Features/Import/S3/ImportJobResultStore.cs
```diff
@@ -18,7 +18,17 @@ public class ImportJobResultStore(
 {
     private readonly AwsStorageOptions options = options.Value;
     private readonly string jobResultsPrefix = "importjobresults/";
-    
+
+    public Task<Result<int>> GetTotalImportJobs(CancellationToken cancellationToken)
```
donaldgray: Can this class go? Thought it odd that there are a couple of `NotImplementedException` so had a look in IDE and looks like it's not used + obsolete

src/DigitalPreservation/DigitalPreservation.Common.Model/ChangeDiscovery/Activity.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.Text.Json.Serialization;
+
+namespace DigitalPreservation.Common.Model.ChangeDiscovery;
+
+public class Activity
```
donaldgray, line 5: Is it worth having `actor` on this if we know who's doing what? Can see why we may not want to include that in stream though.

src/DigitalPreservation/Preservation.API/Features/Activity/ActivityController.cs
```diff
@@ -0,0 +1,25 @@
+﻿using DigitalPreservation.Core.Web;
+using MediatR;
+using Microsoft.AspNetCore.Mvc;
+using Preservation.API.Features.Activity.Requests;
+
+namespace Preservation.API.Features.Activity;
+
+[Route("[controller]")]
+[ApiController]
+public class ActivityController(IMediator mediator) : Controller
```
donaldgray, line 10: Should these set `Content-Type: application/ld+json;profile="http://iiif.io/api/discovery/1/context.json"`

src/DigitalPreservation/Preservation.API/Features/Activity/Requests/GetArchivalGroupsOrderedCollectionPage.cs
```diff
@@ -0,0 +1,101 @@
...
+                    Id = resourceMutator.GetActivityStreamUri($"archivalgroups/pages/{request.Page + 1}")
+                };
+            }
+            page.WithContext();
+            return Result.OkNotNull(page);
+
+        }
+        catch (Exception e)
+        {
+            logger.LogError(e, e.Message);
```
donaldgray: Tiny, `logger.LogError(e)` automatically logs the error message + stacktrace so so maybe better adding something more informative for 2nd arg. e.g. for `throw new ApplicationException("Dummy error");` serilog will log

```
[10:23:28 ERR GetArchivalGroupsOrderedCollectionPageHandler] Dummy Error
System.ApplicationException: Dummy Error
   at ......
```

donaldgray (approved):
lgtm, couple of small comments.

The activity streams would be relatively simple to add integration tests for too

src/DigitalPreservation/Storage.API/Features/Import/S3/ImportJobResultStore.cs
tomcrane: Yeah, I don't think I'll go back to an S3 impl

src/DigitalPreservation/Preservation.API/Features/Activity/ActivityController.cs
tomcrane: Probably yes, and def yes for the IIIF-CS one, but I'm inclined to leave as plain JSON for Preservation.

Technically we haven't defined the types "ArchivalGroup" etc so we need an extra @context.

PR #27, opened 2025-03-18
Add machine header auth if using machine token

frank-kerrigan:
See https://github.com/uol-dlip/digital-preservation/pull/23  for details. 

Putting machine token into branch.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Core/Web/Headers/PropagateCorrelationIdHandler.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Core/Web/Headers/PropagateCorrelationIdHandler.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Extensions.Http;
+﻿using DigitalPreservation.Core.Auth;
+using Microsoft.Extensions.Http;
 using Microsoft.Extensions.Primitives;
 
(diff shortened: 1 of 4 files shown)


src/DigitalPreservation/DigitalPreservation.Core/Web/Headers/PropagateCorrelationIdHandler.cs
```diff
@@ -54,7 +55,12 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
             request.Headers.TryAddWithoutValidation("Authorization", bearerToken);
         }
 
-
+        //Pass Machine Token Downstream
+        var machineToken = contextAccessor.HttpContext.TryGetHeaderValue(AuthFilterIdentifier.MachineHeaderName);
```
donaldgray, line 59: `PropagateCorrelationIdHandler` will add this to every outgoing request for all HttpClients, which seems a bit dangerous. Should this go elsewhere? OR should `PropagateCorrelationIdHandler` be renamed to make it more obvious what it does? (the original intention was for this to just fire `x-correlation-id` out willy nilly as it's not a secret and made sense to add to all requests).

donaldgray (approved):

PR #32, opened 2025-03-27
Add different Auth for ActivityController 

frank-kerrigan:
**Issue**
The PreservationAPI  has a background worker process which calls the SecurityAPI Activity controller.  This controller however does not have have a Bearer Token that it can pass through as its is not called by a user (or machine token).

**Approach**
Implement a separate Authentication method for this (and possibly other controllers).  Generated TOTP tokens from the Presentation API are used for calling the Activity stream via a custom authorisation filter. This has an advantage over a fixed APIKey as the token will constantly change and is shorted lived (300 seconds set in code). Also APIKey management is not required as the token exchange uses the same private seed. 

**Testing***
Looked at calls in code between the layers.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Common.Model/ChangeDiscovery/Reader/ActivityStreamReader.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Common.Model/ChangeDiscovery/Reader/ActivityStreamReader.cs
@@ -1,4 +1,6 @@
 ﻿using System.Net.Http.Json;
+using DigitalPreservation.Common.Model.Constants;
+
 
 namespace DigitalPreservation.Common.Model.ChangeDiscovery.Reader;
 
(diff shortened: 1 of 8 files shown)


src/DigitalPreservation/DigitalPreservation.Common.Model/ChangeDiscovery/Reader/ActivityStreamReader.cs
```diff
@@ -8,17 +10,26 @@ namespace DigitalPreservation.Common.Model.ChangeDiscovery.Reader;
 /// <param name="httpClient"></param>
 public class ActivityStreamReader(HttpClient httpClient)
```
tomcrane, line 11: I am pretty sure that this injected httpClient will never be ALSO used for a regular (pass-through) request to the Storage API - but are we absolutely sure about that?

DI for http client: https://github.com/uol-dlip/digital-preservation/blob/main/src/DigitalPreservation/Storage.Client/ServiceCollectionX.cs#L25

PR #37, opened 2025-04-10
Escape chars in RDF Turtle statements going to Fedora

tomcrane:
Starting adding real content to Fedora and it revealed this bug, created this test for it:

```
    [Fact]
    public void Single_Rdf_Statement_Dodgy_Name()
    {
        // Arrange
        var msg = new HttpRequestMessage();

        // Act
        msg.WithName("Barth Bridge. Original drawing used in \"The Yorkshire Dales\" (1956), page 156");
        
        // Assert
        ((StringContent)msg.Content!).ReadAsStringAsync().Result.Should()
            .Be($"""
                 PREFIX dc: <http://purl.org/dc/elements/1.1/>
                 <> dc:title "Barth Bridge. Original drawing used in \"The Yorkshire Dales\" (1956), page 156" .
                 """);
    }
```

(Note that the \" escape in the final Assert section are LITERAL - they are really there in the string. The $""" escaped string makes it look like the // Act string but it isn't!

image

### DIFF
--- a/src/DigitalPreservation/Storage.API.Tests/Fedora/RequestXTests.cs
+++ b/src/DigitalPreservation/Storage.API.Tests/Fedora/RequestXTests.cs
@@ -1,0 +1,40 @@
+﻿using Storage.API.Fedora.Http;
+
+namespace Storage.API.Tests.Fedora;
+
+public class RequestXTests
+{
+    [Fact]
+    public void Single_Rdf_Statement_Simple_Name()
+    {
+        // Arrange
+        var msg = new HttpRequestMessage();
+
+        // Act
+        msg.WithName("Simple name");
+        
+        // Assert
+        ((StringContent)msg.Content!).ReadAsStringAsync().Result.Should()
+            .Be($"""
+                 PREFIX dc: <http://purl.org/dc/elements/1.1/>
+                 <> dc:title "Simple name" .
+                 """);
+    }
+    
+    [Fact]
+    public void Single_Rdf_Statement_Dodgy_Name()
+    {
+        // Arrange
+        var msg = new HttpRequestMessage();
+
+        // Act
+        msg.WithName("Barth Bridge. Original drawing used in \"The Yorkshire Dales\" (1956), page 156");
+        
+        // Assert
+        ((StringContent)msg.Content!).ReadAsStringAsync().Result.Should()
+            .Be($"""
+                 PREFIX dc: <http://purl.org/dc/elements/1.1/>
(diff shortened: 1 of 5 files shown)


src/DigitalPreservation/Storage.API.Tests/Fedora/RequestXTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    public void Single_Rdf_Statement_Simple_Name()
+    {
+        // Arrange
+        var msg = new HttpRequestMessage();
+
+        // Act
+        msg.WithName("Simple name");
+        
+        // Assert
+        ((StringContent)msg.Content!).ReadAsStringAsync().Result.Should()
```
donaldgray, line 17: I don't _think_ it'll have any practical affect in tests but best practice is to `await` rather than access `.Result`.

src/DigitalPreservation/Storage.API/Fedora/Http/RdfBodyBuilder.cs
```diff
@@ -8,6 +8,44 @@ namespace Storage.API.Fedora.Http;
 public static partial class RdfBodyBuilder
 {
     private static readonly MediaTypeHeaderValue Turtle = MediaTypeHeaderValue.Parse("text/turtle");
+
+    public static string EscapeForLiteralRdf(this string s, bool stripLineBreaks)
```
donaldgray, line 12: Unit tests for this? I think the `RequestXTests` contains tests for 1 permutation but worth verifying tests for all handled chars?

src/DigitalPreservation/Storage.API/Fedora/FedoraClient.cs
```diff
@@ -386,7 +386,11 @@ private static string GetStorageMapCacheKey(Uri archivalGroupUri, string? versio
             req.AsArchivalGroup();
         }
         var response = await httpClient.SendAsync(req);
-        response.EnsureSuccessStatusCode();
+        if (!response.IsSuccessStatusCode)
+        {
+            return Result.Fail<Container>(ErrorCodes.GetErrorCode((int?)response.StatusCode),
```
donaldgray, line 391: General thing for all the `Result.Fail` - are they logged somewhere for tracing through a "why's this failing"? question once deployed?

donaldgray (approved):

src/DigitalPreservation/Storage.API/Fedora/FedoraClient.cs
tomcrane: They should generally be ending up in "error" fields in DB import job results

PR #44, opened 2025-04-22
Fix auth token generation issue

frank-kerrigan:
**Fix**

see: https://dev.azure.com/universityofleeds/Library/_workitems/edit/93159

- Add Session Cookie When Account Not In Cache Events  handler
- Add to step 1 to CookieAuthenticationOptions in builder services
- Change AuthTokenInjector (and TimingHandler) to Scoped DI
- Change AuthTokenInjector so throws exception.

**Test**
Tested locally across 4 sessions


tomcrane (approved):
Approving as we have to test in dev anyway - make sense to me from what I can see

PR #46, opened 2025-04-23
Make Preservation API first class Auth Client

frank-kerrigan:
**Changes**
Add token handler to inject a machine token for calls to storage api from pres api.
Includes mem cache to reduce token fetch. 
Removal of Topt/totp code. 

**Testing**
Testing locally to ensure cache works and refresh and calls activity controller on storage api


src/DigitalPreservation/DigitalPreservation.Core/Web/Headers/PropagateCorrelationIdHandler.cs
```diff
@@ -33,13 +38,19 @@ public Action<HttpMessageHandlerBuilder> Configure(Action<HttpMessageHandlerBuil
...
 
     protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
     {
-        if (contextAccessor.HttpContext == null) return base.SendAsync(request, cancellationToken);
+        if (contextAccessor.HttpContext == null)
+        {
+            SetMachineToken(request);
+
+            return base.SendAsync(request, cancellationToken);
+        }
```
tomcrane, line 53: The thing to watch for will be whether we see things being created by the wrong "user" - "api-call"

...and whether the Storage API extracts that callerIdentity correctly

PR #52, opened 2025-05-07
Add Development, Test and Production for deployments

frank-kerrigan:
Add functionality to deploy to different ECR tags.

How this will affect the works flow.   

**DEV automatic CI/CD** 
There should be no changes as default environment is `development`

**TEST and PROD**
1. Create a Tagged release
2. Run workflow against that tag picking target environment.  

This will allow for older tags releases to be redeployed if roll back is required. 
Releases will contain differences between releases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,15 +2,25 @@ name: Build, Test & Publish dotnet Images
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


donaldgray (approved):
LGTM - only thought is whether it's worth having tagged releases automatically go to test?

PR #59, opened 2025-06-10
use background keep transaction alive Timer

tomcrane:
This PR contains a ragbag of tweaks I made while investigating the issues with very large commits.

## Minor fixes

In trying various fixes I uncovered a couple of issues in unrelated code, so lets get that out of way first:

 - src/DigitalPreservation/DigitalPreservation.Common.Model/ToolOutput/Siegfried/SiegfriedOutput.cs
 - If a Siegfried report is created on Windows it will use `\` as the file path separator; needed normalise this to `/` for processing.

 - src/DigitalPreservation/DigitalPreservation.Core/Web/Handlers/TimingHandler.cs
 - Log the verb of the HTTP request as well as the URI

 - src/DigitalPreservation/Storage.API/Features/Binaries/ContentController.cs
 - Improvements to how the Storage API streams binary content; make sure it only does this for a binary.
 - NB: this is not proxied by the Preservation API except in the special case of 
 
## Long running Fedora calls with no activity on the connection:

 - src/DigitalPreservation/DigitalPreservation.Core/Web/HttpClientBuilderX.cs
 - src/DigitalPreservation/Preservation.Client/ServiceCollectionX.cs
 - src/DigitalPreservation/Storage.API/Fedora/ServiceCollectionX.cs
 - src/DigitalPreservation/Storage.Client/ServiceCollectionX.cs

This implements the suggestions in 

 - https://repost.aws/articles/ARjhCdFMoTTwmxenTc4B7elg/how-to-avoid-network-timeout-issues-when-invoking-long-running-lambda-functions-from-net6-applications-on-linux-platforms
 - https://github.com/dotnet/runtime/issues/77755
 - https://github.com/dotnet/runtime/issues/24917

Issues: to run on Linux we need to use the IP address not the host name. What overhead (if any) does this add? Is it a problem running this tcp keep-alive on all requests within the VPC?

Alternatives: use internal host names rather than public ones. This *probably* avoids the NAT gateway timeout of 350 seconds mentioned above. But it would require work - the applications need to be aware of their public and internal hostnames and translate between them.

## Transaction monitor

This addresses the problem with Fedora not returning - not even attempting to return, so not the same as above - from very long running transaction commit calls. Previously we refreshed the transaction between files if it hadn't been refreshed for 60s, but then assumed that once the commit was called it can take care of itself.

Now, this monitoring happens in a separate thread. This thread can inspect the running transaction and if it detects an issue and commit has already been called, it can cancel that HTTP request, allowing the process to continue.

 - src/DigitalPreservation/Storage.API/Features/Import/Requests/ExecuteImportJob.cs
 - src/DigitalPreservation/Storage.API/Features/Import/Requests/FedoraTransactionMonitor.cs

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Core/Web/HttpClientBuilderX.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Core/Web/HttpClientBuilderX.cs
@@ -1,9 +1,14 @@
-﻿using System.Net.Sockets;
+﻿using System.Net;
+using System.Net.Sockets;
 
 namespace DigitalPreservation.Core.Web;
 
 public static class HttpClientBuilderX
 {
+    // See https://github.com/dotnet/runtime/issues/77755
+    // See https://github.com/dotnet/runtime/issues/24917
+    // https://repost.aws/articles/ARjhCdFMoTTwmxenTc4B7elg/how-to-avoid-network-timeout-issues-when-invoking-long-running-lambda-functions-from-net6-applications-on-linux-platforms
+    
     public static IHttpClientBuilder ConfigureTcpKeepAlive(
         this IHttpClientBuilder builder, 
         bool enabled,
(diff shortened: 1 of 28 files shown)


frank-kerrigan (approved):

src/DigitalPreservation/DigitalPreservation.Core/Web/HttpClientBuilderX.cs
```diff
@@ -0,0 +1,81 @@
...
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, 15);
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveRetryCount, 5);
+            await socket.ConnectAsync(ipAddresses, context.DnsEndPoint.Port, token);
+            return new NetworkStream(socket, ownsSocket: true);
+        }
+        catch
+        {
+            socket.Dispose();
+            throw;
+        }
```
frank-kerrigan, line 56: Maybe move dispose to  finally.

```
finally 
{
   socket.Dispose();
}
```

src/DigitalPreservation/DigitalPreservation.Core/Web/HttpClientBuilderX.cs
```diff
@@ -0,0 +1,81 @@
...
+            socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveRetryCount, 60); // was 10
+            await socket.ConnectAsync(context.DnsEndPoint, token).ConfigureAwait(false);
+
+            return new NetworkStream(socket, ownsSocket: true);
+        }
+        catch
+        {
+            socket.Dispose();
+            throw;
+        }
```
frank-kerrigan, line 78: as above

src/DigitalPreservation/DigitalPreservation.Core/Web/HttpClientBuilderX.cs
tomcrane: What do you think of this discussion?
https://stackoverflow.com/questions/1555567/when-is-finally-run-if-you-throw-an-exception-from-the-catch-block

The code is adapted from an AWS blog post and a couple of MS examples none of which use `finally`.

I would assume that the Microsoft code higher up the stack handles the `throw` and therefore the `finally` will run, but I can't guarantee it.

PR #73, opened 2025-09-12
Fix/add aria labels and suppress error messages

brianmac1967:
Following testing Christine had some feedback which will be fixed in this PR.

Thus it will add aria labels to pipeline job table row cells and will refresh the combined directory before calculating the mismatches as this was incorrectly displaying a METS mismatch message after the pipeline run (the pipeline refreshes the METS file). Also, will disable the "Create diff import job" when the pipeline job is running. All TempData messages will be suppressed whilst the pipeline is running.

Also, removed the PipelineOptions from `appSettings.json` as now applied via Terraform.
Keeping in the `appsettings.Example.json` as will point to the `dlip-pres-local-<queue>` for running locally.
Removed all options from `appsettings.Testing.json` as all IOptions applied via Terraform on testing.

added in arguments to logging statements instead of string interpolation


tomcrane (approved):
Still some compiler warnings (see end of file diff)

PR #86, opened 2025-10-01
Pipeline identity

tomcrane:
This started from documenting the Pipeline features of the Preservation API and feeling uncomfortable about documenting the `runUser` parameter, so I started an experiment with removing that from the public API and relying on callerIdentity (the execution side still uses the concept).

I then did a bit of refactoring as I followed through the execute logic, which doesn't change behaviour but removes a lot of repetitive code blocks and makes the logic flow easier to follow (IMO). Some of the changes are really from the POV of an API consumer who is *not* the Preservation UI; what does it look like as a public API?

I also made some UI changes in Deposit.cshtml.

Summary of changes:

* Mediatr handlers take a `ClaimsPrincipal` rather than `string runUser`
* When starting a pipeline run, the API logs the first pipeline run status as well as sends the SNS message, rather than the caller logging the message. All messages are out of the API client's hands, and the `LogPipelineRunStatus` is a non-documented public API method (for internal use only). See `RunPipeline` and `RunPipelineStatus`. (might want to rename the latter but another time).
* The minting of the identity moves behind the Preservation API rather than in the UI. That is, callers can't supply their own jobId.
* In the Deposit view I have made the inactive action menu items grey - although they were `disabled` they still didn't look visually different
* In the Deposit Razor and model I have moved some code into the Model class (`RunningPipelineJob`) and removed the use of runUser. I renamed PipelineJobsRunning to GetCleanedPipelineJobsRunning and incorporated CleanupPipelineRunsForDeposit into it so that these definitely happen together.
* In ExecutePipelineJob I introduced some new methods: `TryReleaseLock` and various overloads of `UpdateJobStatus`. Although these add some new lines they save about 100 lines of code where the same actions are performed during the flow. This also isolated the point at which a chained Mediatr call happens, which I then refactored to call the preservation API directly so there is no Mediatr chain (see UpdateAnyJobStatus - this used to be many calls to await `mediator.Send(new LogPipelineJobStatus(..))`).
* Ordered pipeline jobs by dateSubmitted (there was no ordering) and did same for importJobs to match.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Common.Model/Error.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Common.Model/Error.cs
@@ -5,14 +5,16 @@ namespace DigitalPreservation.Common.Model;
 public class Error
 {
     /// <summary>
-    /// Id directly to this version
+    /// Id directly to this version -optional
     /// </summary>
     [JsonPropertyName("id")]
     [JsonPropertyOrder(1)]
     public Uri? Id { get; set; }
     
     /// <summary>
-    /// 
+    /// The error text
     /// </summary>
+    [JsonPropertyName("message")]
+    [JsonPropertyOrder(2)]
     public required string Message { get; set; }
 }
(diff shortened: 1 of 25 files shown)


brianmac1967 (approved):
LGTM

brianmac1967 (approved):

PR #102, opened 2025-10-16
Get content/mime type from server side for less known file extensions like .msg

brianmac1967:
Get content/mime type from server side for less known file extensions like .msg
Also, get some Metadata properties from the `FileInDeposit` object if the associated `DepositFileFormatMetadata` properties are null.

Added in changes to show more information in METS mismatch banner.

Ticket:
https://dev.azure.com/universityofleeds/Library/_workitems/edit/106850

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/CombinedFile.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/CombinedFile.cs
@@ -172,8 +172,8 @@ public class CombinedFile(WorkingFile? fileInDeposit, WorkingFile? fileInMets, s
             cachedDepositFileFormatMetadata = FileInDeposit?.GetFileFormatMetadata();
 
 
-            if (cachedDepositFileFormatMetadata != null && string.IsNullOrWhiteSpace(cachedDepositFileFormatMetadata?.FormatName))
-                cachedDepositFileFormatMetadata!.FormatName = "Blah";
+            if (cachedDepositFileFormatMetadata != null && string.IsNullOrWhiteSpace(cachedDepositFileFormatMetadata?.ContentType))
+                cachedDepositFileFormatMetadata!.ContentType = !string.IsNullOrWhiteSpace(FileInDeposit!.ContentType) ? FileInDeposit.ContentType : "All empty";
             if (cachedDepositFileFormatMetadata != null && string.IsNullOrWhiteSpace(cachedDepositFileFormatMetadata?.FormatName))
                 cachedDepositFileFormatMetadata!.FormatName = "[Not Identified]";
             if (cachedDepositFileFormatMetadata != null && string.IsNullOrWhiteSpace(cachedDepositFileFormatMetadata?.PronomKey))
--- a/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/CombinedFile.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/CombinedFile.cs
@@ -176,7 +176,7 @@ public class CombinedFile(WorkingFile? fileInDeposit, WorkingFile? fileInMets, s
                 cachedDepositFileFormatMetadata!.ContentType = !string.IsNullOrWhiteSpace(FileInDeposit!.ContentType) ? FileInDeposit.ContentType : "All empty";
             if (cachedDepositFileFormatMetadata != null && string.IsNullOrWhiteSpace(cachedDepositFileFormatMetadata?.FormatName))
                 cachedDepositFileFormatMetadata!.FormatName = "[Not Identified]";
-            if (cachedDepositFileFormatMetadata != null && string.IsNullOrWhiteSpace(cachedDepositFileFormatMetadata?.PronomKey))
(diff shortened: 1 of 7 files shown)


frank-kerrigan (approved):

src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/CombinedFile.cs
```diff
@@ -83,7 +83,7 @@ private List<FileMisMatch> GenerateMisMatches()
                     DepositFileFormatMetadata.PronomKey, MetsFileFormatMetadata.PronomKey));
             }
 
-            if (DepositFileFormatMetadata!.ContentType != FileInMets.ContentType)
+            if ((DepositFileFormatMetadata!.ContentType ?? FileInDeposit.ContentType) != FileInMets.ContentType)
```
tomcrane: Where would FileInDeposit.ContentType have come from if DepositFileFormatMetadata!.ContentType is null?

src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/CombinedFile.cs
```diff
@@ -170,6 +170,15 @@ public FileFormatMetadata? DepositFileFormatMetadata
                 return cachedDepositFileFormatMetadata;
             }
             cachedDepositFileFormatMetadata = FileInDeposit?.GetFileFormatMetadata();
+
+
+            if (cachedDepositFileFormatMetadata != null && string.IsNullOrWhiteSpace(cachedDepositFileFormatMetadata?.ContentType))
+                cachedDepositFileFormatMetadata!.ContentType = !string.IsNullOrWhiteSpace(FileInDeposit!.ContentType) ? FileInDeposit.ContentType : "All empty";
+            if (cachedDepositFileFormatMetadata != null && string.IsNullOrWhiteSpace(cachedDepositFileFormatMetadata?.FormatName))
```
tomcrane: I don't think this should set FormatName or PronomKey because meaningful values for them can only come from Siegfried or similar. I know that MetsParser sets these when _reading_ if it can't find anything else, but I don't think these should be set on the way in.

And if they aren't being set and it's just the ContentType, then `FileInDeposit?.GetFileFormatMetadata()` on line 175, if we use the same syntheticmetadata code branch when no FileFormatMetadata is present, should be the place that the  MimeTypes library is invoked. But now I am having second thoughts about that too (see below).

src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/WorkingFile.cs
```diff
@@ -53,7 +53,7 @@ public WorkingFile ToRootLayout()
             .OfType<FileFormatMetadata>()
             .ToList();
 
-        if (fileFormatMetadata.Count == 0 && FolderNames.IsMetadata(LocalPath))
+        if (fileFormatMetadata.Count == 0)
```
tomcrane: Having second thoughts about this and realising maybe why this was only run on metadata/* files. 

It's OK to use `MimeTypes.TryGetMimeType(..` in Deposit.cshtml.cs, because it's being used as a sort of placeholder content type until a pipeline can be run. And it's only used if the client-side detection didn't produce anything.

But doing it here is pretending to be a real FileFormatMetadata - but it's not as trustworthy as the Siegfried-produced FileFormatMatadata. So maybe we shouldn't do it all.

src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/WorkingFile.cs
```diff
@@ -70,8 +70,14 @@ public WorkingFile ToRootLayout()
                 syntheticMetadata.ContentType = foundMimeType;
             }
 
+            if (FolderNames.IsMetadata(LocalPath))
```
tomcrane: do we need this?

tomcrane (commented):
Going round in circles writing notes on this. I'm starting to feel like the changes in `CombinedFile` and `WorkingFile` aren't right and shouldn't be there. The addition to Deposit.cshtml.cs (using `MimeTypes` on the server if the browser didn't suggest a value) is fine, because it's no more authoritative than the client-side suggestion.

Does that on its own fix the original problem, without making any changes to `CombinedFile` and `WorkingFile`?

It's OK for us to return "synthetic" FileFormatMetadata for files under metadata/* because they can't get it any other way. But generating synthetic metadata for objects/* files is giving a false picture of what we know.

src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/WorkingFile.cs
brianmac1967: Displays dlip/unknown otherwise - only reason I put that in.

brianmac1967:
> Going round in circles writing notes on this. I'm starting to feel like the changes in `CombinedFile` and `WorkingFile` aren't right and shouldn't be there. The addition to Deposit.cshtml.cs (using `MimeTypes` on the server if the browser didn't suggest a value) is fine, because it's no more authoritative than the client-side suggestion.
> 
> Does that on its own fix the original problem, without making any changes to `CombinedFile` and `WorkingFile`?
> 
> It's OK for us to return "synthetic" FileFormatMetadata for files under metadata/* because they can't get it any other way. But generating synthetic metadata for objects/* files is giving a false picture of what we know.

changing Deposit.cshtml.cs alone does not fix the issues.
The code patches in Combined File and Working file I added as I went. I was not sure of the level/where the fix should be applied.

PR #110, opened 2025-12-03
Output Exif file to a metadata folder file during pipeline run

brianmac1967:
Ticket: https://digirati.atlassian.net/browse/LPII-1
The code uploads local process folder metadata folder exif output file to s3:
image

Tested this using the local pipeline API swagger instance at `https://localhost:7005/swagger/index.html` to trgigger the pipeline run:
image

Add the following setting to the Pipeline API `appsettings.Development.json` `BrunnhildeOptions` section- will get Frank to add this to the config settings terraform (maybe) - might just put in `appsettings.json`:

` "ExifToolLocation": "C:\\Exif\\exiftool.exe"`

image

Do not deploy as need to get Exif tool installed to the pipeline Docker container - need to chat with Frank about this.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.sln
+++ b/src/DigitalPreservation/DigitalPreservation.sln
@@ -159,9 +159,6 @@ Global
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
 	EndGlobalSection
-	GlobalSection(SolutionProperties) = preSolution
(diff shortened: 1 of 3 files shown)


tomcrane (approved):
I think we should rename BrunnhildeOptions to something else as most of those values are not specifically to do with brunnhilde. Will need to coordinate with tf change though as well as appsettings.

PR #121, opened 2025-12-17
Feature/lpii 5/parse exif metadata from mets

brianmac1967:
Ticket:
https://digirati.atlassian.net/browse/LPII-5
https://digirati.atlassian.net/browse/LPII-12

Have added a list comparer for the Deposit and METS Exif Metadata lists.
Added a test to work on the technique.
Handling duplicates in the Exif output.

To reproduce mismatches - just edit/add items in the `MetsParser.cs` collection when running locally after running a pipeline job.

To add an item to `exifMetadataList` collection:
`exifMetadataList.Add(new ExifTag { TagName = "OnlyInMets", TagValue = "METS Only" });`

Here in screenshot `exifMetadataList` collection

image

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/Extensions/Metadata/ExifMetadata.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/Extensions/Metadata/ExifMetadata.cs
@@ -1,10 +1,11 @@
 ﻿using System.Text.Json.Serialization;
+using DigitalPreservation.Common.Model.DepositHelpers;
 
 namespace DigitalPreservation.Common.Model.Transit.Extensions.Metadata;
 
 public class ExifMetadata : Metadata
 {
-    public Dictionary<string, string>? RawToolOutput { get; set; }
+    public List<ExifTag>? RawToolOutput { get; set; } = [];
     public int Width { get; set; }
     public int Height { get; set; }
     public double BitRate { get; set; }
--- a/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/Extensions/Metadata/ExifMetadata.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/Extensions/Metadata/ExifMetadata.cs
@@ -5,9 +5,5 @@ namespace DigitalPreservation.Common.Model.Transit.Extensions.Metadata;
 
 public class ExifMetadata : Metadata
 {
-    public List<ExifTag>? RawToolOutput { get; set; } = [];
-    public int Width { get; set; }
-    public int Height { get; set; }
-    public double BitRate { get; set; }
-    public double Duration { get; set; }
+    public List<ExifTag>? Tags { get; set; } = [];
 }
--- a/src/DigitalPreservation/DigitalPreservation.Core.Tests/Lists/ListsComparer.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Core.Tests/Lists/ListsComparer.cs
@@ -6,19 +6,7 @@ namespace DigitalPreservation.Core.Tests.Lists;
 public class ListsComparer
 {
     [Fact]
-    public void CompareTwoPersonLists()
-    {
-        var source = new List<Person>() { new("Ken", "Nakamura"), new("Nozomi", "Nakamura") };
-        var compare = new List<Person>() { new("Ken", "Nakamura"), new("Keiko", "Nakamura") };
-        //var result = source.Intersect(compare);
-
-        bool isEqual = source.SequenceEqual(compare, new PersonComparer());
-        var t = source.Except(compare, new PersonComparer());
-        var t1 = compare.Except(source, new PersonComparer());
-    }
-
-    [Fact]
-    public void CompareTwoTagValueLists()
+    public void CompareTwoTagValueListsWithMismatches()
     {
         var depositExifMetadata = new List<ExifTag>();
         depositExifMetadata.Add(new ExifTag {TagName = "ExifToolVersionNumber", TagValue = "13:42"});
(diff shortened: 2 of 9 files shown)


src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/Extensions/Metadata/ExifMetadata.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using System.Text.Json.Serialization;
+using DigitalPreservation.Common.Model.DepositHelpers;
 
 namespace DigitalPreservation.Common.Model.Transit.Extensions.Metadata;
 
 public class ExifMetadata : Metadata
 {
-    public Dictionary<string, string>? RawToolOutput { get; set; }
+    public List<ExifTag>? RawToolOutput { get; set; } = [];
```
tomcrane: I think this should be renamed as it is no longer the raw tool output - how about `Tags`?

tomcrane: Also, my IDE reports that all of these are not used - if we're getting this another way we can remove these fields:

```
    public int Width { get; set; }
    public int Height { get; set; }
    public double BitRate { get; set; }
    public double Duration { get; set; }
```

src/DigitalPreservation/DigitalPreservation.Core.Tests/Lists/ListsComparer.cs
```diff
@@ -0,0 +1,142 @@
+﻿using DigitalPreservation.Common.Model.DepositHelpers;
+using DigitalPreservation.Common.Model.Transit;
+using DigitalPreservation.Common.Model.Transit.Extensions.Metadata;
+
+namespace DigitalPreservation.Core.Tests.Lists;
+public class ListsComparer
```
tomcrane, line 6: I know this class is here to help develop `ExifTagComparer` but it could now be updated to make some assertions - have one test where we expect it to assert that there are no mismatches, and another test where it expects them.

src/DigitalPreservation/DigitalPreservation.Core.Tests/Lists/ListsComparer.cs
```diff
@@ -0,0 +1,142 @@
...
+        var resultDeposit = depositExifMetadata.Where(p => metsExifMetadata.All(p2 => p2.TagName != p.TagName));
+        misMatches.AddRange(resultDeposit.Select(depositField => new CombinedFile.FileMisMatch(nameof(ExifMetadata), depositField.TagName ?? string.Empty, depositField.TagValue, "field does not exist in METS")));
+        
+        var resultMets = metsExifMetadata.Where(p => depositExifMetadata.All(p2 => p2.TagName != p.TagName));
+        misMatches.AddRange(resultMets.Select(metsField => new CombinedFile.FileMisMatch(nameof(ExifMetadata), metsField.TagName ?? string.Empty, metsField.TagValue, "field does not exist in deposit")));
+    }
+
+}
+
+public class Person
```
tomcrane: And can probably get rid of this now

src/DigitalPreservation/DigitalPreservation.Core.Tests/Lists/ListsComparer.cs
```diff
@@ -0,0 +1,142 @@
...
+        //Get hash code for the name field if it is not null
+        int firstNameHashCode = !string.IsNullOrEmpty(person.FirstName) ? 0 : person?.FirstName?.GetHashCode() ?? 0;
+        int lastNameHashCode = !string.IsNullOrEmpty(person?.LastName) ? 0 : person?.LastName?.GetHashCode() ?? 0;
+        // Get hash code for marks also if its not 0
+
+        return firstNameHashCode ^ lastNameHashCode;
+    }
+}
+
+public class ExifTagComparer : IEqualityComparer<ExifTag>
```
tomcrane: This class should be removed from here and the test should test `DigitalPreservation.Common.Model.Transit.ExifTagComparer`

src/DigitalPreservation/DigitalPreservation.Workspace/MetadataReader.cs
```diff
@@ -513,15 +514,15 @@ public static List<ExifModel> ConvertExifResultStringToJson(string exifResultStr
 
             var exifModel = new ExifModel { Filepath = string.Empty, ExifMetadata = [] };
             var i = 0;
-            var exifMetadataForFile = new Dictionary<string, string>();
+            var exifMetadataForFile = new List<ExifTag>();
```
tomcrane, line 517: The enclosing method (ConvertExifResultStringToJson) should be called `ParseExifToolOutput`

tomcrane: Also (I can't put this comment on the line as it isn't a change in the current PR) - the variable `brunnhildeAvCommonPrefix` (line 139) should be renamed `brunnhildeAvCommonPrefix` as it is re-used for non-virus things

tomcrane: (and similarly the `source` passed on line 176:

`AddExifMetadata(brunnhildeAvCommonPrefix, "Exif", timestamp);`

I think that should be "ExifTool" - i.e., the tool name that generated the metadata (as with ClamAv earlier).

src/DigitalPreservation/Storage.Repository.Common/Mets/MetsParser.cs
```diff
@@ -630,6 +632,26 @@ private void ProcessChildStructDivs(MetsFileWrapper mets, XDocument xMets, XElem
...
+                    var exifMetadataList = new List<ExifTag>();
+                    foreach (var element in exifMetadataNode.Descendants())
+                    {
+                        exifMetadataList.Add(new ExifTag{TagName = element.Name.LocalName , TagValue = element.Value });
+                    }
+
+                    exifMetadata = new ExifMetadata
+                    {
+                        Source = "METS",
+                        Timestamp = DateTime.UtcNow,
```
tomcrane: Set a variable to DateTime.UtcNow outside of the loop that goes through all the files - and then assign that variable to `Timestamp` here, so that they all have the same exact timestamp - otherwise they will differ by a few microseconds. It's cleaner.

src/DigitalPreservation/Storage.Repository.Common/Mets/PremisManager.cs
```diff
@@ -95,7 +95,7 @@ public static PremisComplexType Create(FileFormatMetadata premisFile, ExifMetada
         {
             var document = new XmlDocument();
             var parentCollection = new Collection<XmlElement>();
-            var parentElement = GetXmlElement(new KeyValuePair<string, string>("ExifMetadata", string.Empty), document);
+            var parentElement = GetXmlElement(new ExifTag{ TagName = "ExifMetadata" , TagValue = string.Empty}, document);
```
tomcrane, line 97: I have a warning about `parentCollection` - updated, but never used - it's parentExtension that does the work.

src/DigitalPreservation/Storage.Repository.Common/Mets/PremisManager.cs
```diff
@@ -225,56 +222,53 @@ public static void Patch(PremisComplexType premis, FileFormatMetadata premisFile
         {
             var document = new XmlDocument();
             var parentCollection = new Collection<XmlElement>();
-            var parentElement = GetXmlElement(new KeyValuePair<string, string>("ExifMetadata", string.Empty), document);
-
-            if (parentElement != null)
+            var parentElement = GetXmlElement(new ExifTag { TagName = "ExifMetadata", TagValue = string.Empty }, document);
```
tomcrane, line 222: Same warning - parentCollection is updated but never used

tomcrane (changes requested):
Just a few changes, want to make sure warnings are eliminated and a few variables need clearer names.

tomcrane (approved):
LGTM

PR #122, opened 2025-12-19
Metadata Report viewing

frank-kerrigan:
When for new tab approach see https://digirati.atlassian.net/browse/LPII-7 for details.

Changes

1. Base64 encode (html encode failed) 
2. Button for open new tab
3. JS to decode base64 page and open

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.UI/Pages/Shared/_Layout.cshtml
+++ b/src/DigitalPreservation/DigitalPreservation.UI/Pages/Shared/_Layout.cshtml
@@ -4,7 +4,6 @@
 <html lang="en" data-bs-theme="auto">
 <head>
     <script src="~/js/color-modes.js" asp-append-version="true"></script>
-    <script src="https://cdn.jsdelivr.net/npm/js-base64@3.7.8/base64.min.js"></script>
     <meta charset="utf-8">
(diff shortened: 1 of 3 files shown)


src/DigitalPreservation/DigitalPreservation.UI/Pages/Shared/_Layout.cshtml
```diff
@@ -4,6 +4,7 @@
 <html lang="en" data-bs-theme="auto">
 <head>
     <script src="~/js/color-modes.js" asp-append-version="true"></script>
+    <script src="https://cdn.jsdelivr.net/npm/js-base64@3.7.8/base64.min.js"></script>
```
tomcrane: Do we need an external library? Most browsers (and all those we can expect Leeds to be using) can do this natively:

https://developer.mozilla.org/en-US/docs/Web/API/Window/atob

frank-kerrigan: Yes point taken,  

replaced with 
`
 const raw = decodeURIComponent(atob(htmlString).split('').map(function (c) {
     return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2);
 }).join(''));
`

brianmac1967 (approved):

PR #123, opened 2025-12-19
extra tests as issue when only difference is a property only in Mets or deposit

brianmac1967:
Ticket:
https://digirati.atlassian.net/browse/LPII-5

Added in extra tests and handle duplicates when there values change.
Faked by editing the Deposit `exif_output.txt` file by downloading, editing and uploading to S3 after deleting the original Exif file produced by the pipeline.

Ticket for Unit tests as the ListComparer has been removed as not testing Combined file as required.
EDIT: Unit Tests for ticket below added in.
https://digirati.atlassian.net/browse/LPII-22


src/DigitalPreservation/DigitalPreservation.Core.Tests/Lists/ListsComparer.cs
```diff
@@ -56,6 +60,120 @@ public void CompareTwoTagValueListsWithMismatches()
...
+        depositExifMetadata.Add(new ExifTag { TagName = "ExifToolVersionNumber", TagValue = "13:45" });
+        depositExifMetadata.Add(new ExifTag { TagName = "ColorVersion", TagValue = "Blah" });
+        depositExifMetadata.Add(new ExifTag { TagName = "Duration", TagValue = "Blah" });
+
+        var metsExifMetadata = new List<ExifTag>();
+        metsExifMetadata.Add(new ExifTag { TagName = "ExifToolVersionNumber", TagValue = "13:43" });
+        metsExifMetadata.Add(new ExifTag { TagName = "ExifToolVersionNumber", TagValue = "13:45" });
+        metsExifMetadata.Add(new ExifTag { TagName = "ColorVersion", TagValue = "Blah" });
+
+        var isEqual = depositExifMetadata.SequenceEqual(metsExifMetadata, new ExifTagComparer());
```
tomcrane: These code blocks here and in the next test should be testing CombinedFile rather than a duplicate of its logic, ie it should create a CombinedFile and then populate its metadata and then check its mismatches.

tomcrane (changes requested):
Just a tidy up of the tests so that they don't replicate any business logic but test what CombinedFile is doing with its own business logic

src/DigitalPreservation/DigitalPreservation.Core.Tests/Lists/ListsComparer.cs
brianmac1967: @tomcrane  On reflection, if you agree,  I will park these tests in this PR for now as I was using them for development and dev tests rather than actual unit tests - more for testing the `ExifTagComparer`
I will create a Jira ticket for the Combined file mismatches unit tests and will remove the tests for now in order for the other actual implementation to be reviewed.

I am not sure how to construct the CombinedFile for testing.

brianmac1967: Added a new class for running CombinedFile mismatch tests for Exif Metadata:
https://github.com/digirati-co-uk/digital-preservation/blob/2186c240e13c2047c35421237c239b98080af4cb/src/DigitalPreservation/DigitalPreservation.Core.Tests/Lists/ExifMetadataComparer.cs#L7C22-L7C23

src/DigitalPreservation/DigitalPreservation.Core.Tests/DigitalPreservation.Core.Tests.csproj
```diff
@@ -25,6 +25,7 @@
 
     <ItemGroup>
       <ProjectReference Include="..\DigitalPreservation.Core\DigitalPreservation.Core.csproj" />
+      <ProjectReference Include="..\Preservation.API.Tests\Preservation.API.Tests.csproj" />
```
tomcrane: This might be needed but seems a bit fishy, that one test project references another?

tomcrane (approved):
Pending the comment about project references (which might be OK)

src/DigitalPreservation/DigitalPreservation.Core.Tests/DigitalPreservation.Core.Tests.csproj
brianmac1967: Yeah I might have to refactor as was using the test structure methods for getting METS and Deposit structure from the Pres API tests project. 
Will put this test structure somewhere common.

PR #128, opened 2026-01-13
Introduce ResourceBase base class

tomcrane:
To lay the ground for future resources in a Deposit that need access conditions, links to record identifiers etc, make a common base class `ResourceBase` from which WorkingBase inherits.

Also move some constants to a new `Constants` class to avoid MetsParser having a reference to `IMetsManager`.


brianmac1967 (approved):
LGTM

PR #129, opened 2026-01-15
Namespace tidy-up - CombinedFile

tomcrane:
Moved Combined* classes and Whereabouts to own namespace rather than mixed in Transit
Added JsonProperty attributes to ExifMetadata
Moved ExifTag to be with ExifMetadata


brianmac1967 (approved):
lgtm

PR #131, opened 2026-01-16
Refactor MetsParser; isolate storage implementations.

tomcrane:
This separates out the storage code into S3 and Filesystem implementations.
It builds upon by the previous refactor, which moved storage out of MetsManager, to then move the I/O into separate impls.

Rather than a single implementation of IMetsStorage with the `switch` statements, we have `S3MetsStorage` and `FileSystemMetsStorage`. Tests use the latter, our DI in Program.cs uses the AWS S3 version.

I then started to try to use `IMetsStorage` in `MetsParser`, but it has very different needs. It also would introduce knowledge of the XmlGen classes to the storage interface used by `MetsParser`, which is unnecessary. So MetsParser now takes `IMetsLoader` and there are two implementations of this class, `S3MetsLoader` and `FileSystemMetsLoader`.

None of these introduce any functional changes, it's all refactoring.

It now means the MetsParser logic has no write capability, or dependency on any specific storage backend (filesystem, S3 or something else).

We now have:

IMetsStorage - used by MetsManager to load and save Mets files using the XmlGen classes
IMetsLoader - used by MetsParser to examine the "file system", find the Mets file, and load it.


src/DigitalPreservation/Storage.Repository.Common/Mets/StorageImpl/FileSystemMetsLoader.cs
```diff
@@ -0,0 +1,84 @@
...
+                firstXmlFile = childDirs[0].EnumerateFiles().FirstOrDefault(
+                    f => MetsUtils.IsMetsFile(f.Name.GetSlug(), true));
+                if (firstXmlFile == null)
+                {
+                    firstXmlFile = childDirs[0].EnumerateFiles().FirstOrDefault(
+                        f => MetsUtils.IsMetsFile(f.Name.GetSlug(), false));
+                }
+            }
+        }
+        if (firstXmlFile != null)
```
frank-kerrigan, line 42: This will never be null

src/DigitalPreservation/Storage.Repository.Common/Mets/StorageImpl/FileSystemMetsStorage.cs
```diff
@@ -0,0 +1,79 @@
...
+        DigitalPreservation.XmlGen.Mets.Mets? mets;
+        string? returnedETag;
+        var fileLocResult = await metsParser.GetRootAndFile(metsLocation);
+        var (_, file) = fileLocResult.Value;
+        if (file is null)
+        {
+            return Result.FailNotNull<FullMets>(ErrorCodes.NotFound, "No METS file in " + metsLocation);
+        }
+
+        var fi = new FileInfo(file.LocalPath);
```
frank-kerrigan: put fi in try scope if not used anywhere else ?

src/DigitalPreservation/Storage.Repository.Common/Mets/StorageImpl/S3MetsLoader.cs
```diff
@@ -0,0 +1,112 @@
...
+            resp = await s3Client.ListObjectsV2Async(listObjectsReq);
+            files = resp.S3Objects.Where(s => !s.Key.EndsWith('/')).ToList();
+            firstXmlKey = files.FirstOrDefault(s => MetsUtils.IsMetsFile(s.Key.GetSlug(), true));
+            if (firstXmlKey == null)
+            {
+                firstXmlKey = files.FirstOrDefault(s => MetsUtils.IsMetsFile(s.Key.GetSlug(), false));
+            }
+        }
+
+        if (firstXmlKey != null)
```
frank-kerrigan, line 56: firstXmlKey will never be null ?

frank-kerrigan (changes requested):
couple of comments

src/DigitalPreservation/Storage.Repository.Common/Mets/StorageImpl/FileSystemMetsLoader.cs
tomcrane: Is VS telling you that this can never be null? Because it can, and Rider will warn me if I remove that check.

frank-kerrigan: just leave it

frank-kerrigan (approved):

PR #135, opened 2026-01-20
Feat/lpii 29/refactor edit mets

brianmac1967:
This EditMets method could be refactored further but thought to draw a line and get a Draft PR review at this point.
I have broken the method down by extracting the code in EditMets into subroutines.
Metadata operations have been moved out to `IMetadataManager`

The requirement that needs a discussion is
"At the moment the code just goes through all the things that might need to be edited - premis file format data, virus metadata, exif metadata … and possibly more in future - and then edits them with the help of PremisManager and PremisEventManager. This is hard to follow and will just get bigger and bigger. How do we refactor this to make it easier to follow and maintain and extend in future?"

Ticket:
https://digirati.atlassian.net/browse/LPII-29

All BE Unit tests still passing for the branch:

image


src/DigitalPreservation/Storage.Repository.Common/Mets/MetsManager.cs
```diff
@@ -219,32 +205,8 @@ private Result EditMets(WorkingBase? workingBase, string? deletePath, FullMets f
...
-            }
-            testPath += element;
-            // This is navigating using our ID convention for directories
-            // If we don't want to do this, we can use the premis:originalName for the directory
-            var childDiv = div.Div.SingleOrDefault(d => d.Id == $"{PhysIdPrefix}{testPath}");
-            if (childDiv is null)
-            {
-                break;
-            }
+        var (counter, parent, div, elements, operationPath, testPath) = GetMetsElements(workingBase, deletePath, fullMets);
```
tomcrane, line 208: I don't know how you feel about tuples like this - sometimes a POCO to hold the information is clearer as a return value.

Also the name of this method - not sure what it should be. It's walking the Mets file to find the site of the operation (add, delete or update) that we need to carry out - getting parent and div.

`Find...` something

src/DigitalPreservation/Storage.Repository.Common/Mets/MetsManager.cs
```diff
@@ -253,269 +215,114 @@ private Result EditMets(WorkingBase? workingBase, string? deletePath, FullMets f
         {
             if (deletePath is not null)
             {
```
tomcrane, line 217: Not a request for a change - just to let you know that I expanded this out again - you had made it into a `xxx ? yyy : zzz` ternary operator but I'm only a fan of those if they are very *very* short. Often Rider will suggest I can turn an `if` block into a ternary operator. I let it do the refactor and then undo it immediately because I think it makes the code less clear.

tomcrane (approved):

PR #142, opened 2026-01-23
Refactor/mets parser lookups

tomcrane:
This PR should have no functional changes, no new behaviour.

This matches the way the Python METS Parser works - to avoid repeated Xml document tree traversal it builds maps of the main elements per file.

It also moves the static helper class that generates the root working directory to the `WorkingDirectory` class itself, avoiding a dependency on Storage.cs in MetsParser.


brianmac1967 (approved):
LGTM

PR #143, opened 2026-01-23
put in a 2 second delay before completion status

brianmac1967:
Ticket:
https://digirati.atlassian.net/browse/LPII-13


frank-kerrigan (changes requested):
waits are generally a very bad pattern.   Would be better to test for something is complete or exits before proceeding.

brianmac1967:
> waits are generally a very bad pattern. Would be better to test for something is complete or exits before proceeding.

As discussed put in a loop to attempt release lock multiple times within a configured time.

PR #158, opened 2026-02-04
Change to how siegfried.csv URI is retrieved and allowing non-standard metadata folders to be deleted when not in METS

brianmac1967:
Change to how siegfried.csv URI is retrieved
We could configure these folder values so we can calculate which non-standard/manual metadata folders we can ignore the METS delete result for.
Ticket(s):
https://digirati.atlassian.net/browse/LPII-50

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Workspace/Requests/DeleteItems.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Workspace/Requests/DeleteItems.cs
@@ -34,7 +34,6 @@ public class DeleteItemsHandler(
     IAmazonS3 s3Client,
     IMetsManager metsManager) : IRequestHandler<DeleteItems, Result<ItemsAffected>>
 {
-    private string[] pipelineMetadataFolders = ["metadata/brunnhilde", "metadata/exif", "metadata/virus-definition"]; 
     public async Task<Result<ItemsAffected>> Handle(DeleteItems request, CancellationToken cancellationToken)
     { 
         var goodResult = new ItemsAffected();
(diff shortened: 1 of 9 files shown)


src/DigitalPreservation/DigitalPreservation.Workspace/Requests/DeleteItems.cs
```diff
@@ -34,6 +34,7 @@ public class DeleteItemsHandler(
     IAmazonS3 s3Client,
     IMetsManager metsManager) : IRequestHandler<DeleteItems, Result<ItemsAffected>>
 {
+    private string[] pipelineMetadataFolders = ["metadata/brunnhilde", "metadata/exif", "metadata/virus-definition"];
```
tomcrane: Could knowledge of these folders be confined to the pipeline code rather than the DeleteItems handler? Doesn't seem right that this handler should know about them, its job is just to delete what it's told.

Maybe the DeleteSelection object has a new property - `ContinueIfFail`, defaults to [] - that takes an array of string prefixes? Then line 221 below checks against that?

tomcrane (changes requested):

tomcrane (approved):
Just one change and it's not directly in the changed files.

The class `BrunnhildeOptions` and the appSettings property "BrunnhildeOptions" should be renamed because it's really settings for Brunnhilde, ClamAV, Exif and maybe more tools in future, so the name is potentially confusing.

We already have a `PipelineOptions` class, which has the SQS/SNS settings, so maybe `BrunnhildeOptions` should be renamed `PipelineToolOptions`.

Not sure if there would need to be a terraform change, if any of these options are set by env vars in terraform.

PR #161, opened 2026-02-12
Allow non-new deposit.active flag to be toggled

frank-kerrigan:
Same format as Lock Deposit

### Unit Tests

Added some

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.UI/Pages/Deposits/Deposit.cshtml
+++ b/src/DigitalPreservation/DigitalPreservation.UI/Pages/Deposits/Deposit.cshtml
@@ -66,6 +66,29 @@
                                 </button>
                             </li>
                         }
+                         
+                        
+                        @if (Model.Deposit?.Status is not "new")
+                        {
+
+                            @if (Model.Deposit?.Active is true)
+                            {
+                                <li>
+                                    <button disabled="@jobRunning" class="dropdown-item @(jobRunning ? "text-secondary" : "text-primary")" type="submit" asp-page-handler="DeactivateDeposit">
+                                        <svg class="bi"><use xlink:href="#door-closed"/></svg><span class="ms-2">Deactivate Deposit</span>
+                                    </button>
+                                </li>
+                            }
+                            else
+                            {
+                                <li>
+                                    <button disabled="@jobRunning" class="dropdown-item @(jobRunning ? "text-secondary" : "text-primary")" type="submit" asp-page-handler="ActivateDeposit">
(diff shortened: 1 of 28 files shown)


src/DigitalPreservation/Preservation.API.Tests/PreservationApiClient/DepositsControlerTests.cs
```diff
@@ -0,0 +1,95 @@
+﻿using DigitalPreservation.Common.Model;
```
tomcrane, line 1: typo in class file name (doesn't match class name); should be `DepositsControllerTests.cs`

src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/ActivateDeposit.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System.Security.Claims;
```
tomcrane, line 1: Classes should be `ActivateDeposit` and `ActivateDepositHandler` and filename should be `ActivateDeposit.cs`

src/DigitalPreservation/DigitalPreservation.UI/Pages/Deposits/Deposit.cshtml
```diff
@@ -66,6 +66,29 @@
                                 </button>
                             </li>
                         }
+                         
+                        
+                        @if (Model.Deposit?.Status is not "new")
```
tomcrane, line 71: I think these checks should also be enforced for direct API callers (in the eventual API handler).

We need the checks as here to know what UI to show
But we also need to cater for direct API callers

tomcrane (changes requested):

src/DigitalPreservation/DigitalPreservation.UI/Pages/Deposits/Deposit.cshtml
frank-kerrigan: Added Guard for API Handler before it goes to db entity, added a couple of tests as well for this handler.

src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/ActivateDeposit.cs
frank-kerrigan: done

src/DigitalPreservation/Preservation.API.Tests/PreservationApiClient/DepositsControlerTests.cs
frank-kerrigan: done

tomcrane (approved):

PR #162, opened 2026-02-13
add Exif UI tests

brianmac1967:
changed scope of some of the MetadataReader class methods to allow testing.
Extended MetsParser FileFormatMetadata object to include size and ContentType. Added exif tests to check the ToolOutput is created after parsing the sample exif_output.txt file

Ticket: https://digirati.atlassian.net/browse/LPII-60

### DIFF
--- a/src/DigitalPreservation/XmlGen.Tests/ExifTests.cs
+++ b/src/DigitalPreservation/XmlGen.Tests/ExifTests.cs
@@ -14,7 +14,7 @@ using Xunit.Sdk;
 
 namespace XmlGen.Tests;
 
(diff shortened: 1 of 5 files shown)


src/DigitalPreservation/XmlGen.Tests/ExifTests.cs
```diff
@@ -0,0 +1,159 @@
+﻿using DigitalPreservation.Common.Model.Transit.Extensions.Metadata;
```
tomcrane, line 1: Minor (some unused imports)

tomcrane (approved):

PR #164, opened 2026-02-20
Encode spaces and XML characters in adm id identifiers

brianmac1967:
Encode spaces and XML characters in adm id identifiers.
Unit tests updated.

196 records of files with spaces in their name in Fedora database - see ticket comment:
https://digirati.atlassian.net/browse/LPII-33?focusedCommentId=149047

Ticket:
https://digirati.atlassian.net/browse/LPII-33


src/DigitalPreservation/Storage.Repository.Common/Mets/MetsManager.cs
```diff
@@ -568,9 +581,26 @@ private void SetFileAndFileGroup(DivType div, FullMets fullMets)
                 testPath += "/";
             }
             testPath += element;
+
+            bool objectFile;
```
tomcrane, line 585: What does objectFile do? test whether the current path in the loop is under /objects and is a file rather than a directory?

I think our IDs need to apply equally to files and folders, and equally to any file in the METS not just those under objects.

src/DigitalPreservation/Storage.Repository.Common/Mets/MetsManager.cs
```diff
@@ -568,9 +581,26 @@ private void SetFileAndFileGroup(DivType div, FullMets fullMets)
...
+            }
+            else
+            {
+                objectFile = !string.IsNullOrEmpty(workingBase?.LocalPath) && workingBase is WorkingFile &&
+                             !FolderNames.IsMetadata(workingBase?.LocalPath!);
+            }
+
+
+            var identifiers = FilenameHelpers.GetIdSafeOperationPath(testPath);
+            var physId = objectFile ? identifiers.PhysId : $"{Constants.PhysIdPrefix}{testPath}";
```
tomcrane, line 600: What happens when MetsManager is processing a METS file that was made before this new escaping behaviour? It won't find the div whose `ID` is `physId`. See general note on PR

src/DigitalPreservation/Storage.Repository.Common/Mets/MetsParser.cs
```diff
@@ -387,14 +387,37 @@ private void ProcessChildStructDivs(MetsFileWrapper mets, XElement parent,
...
+                        if (admIdList != null)
+                            foreach (var admIdString in admIdList)
+                            {
+                                lookupMaps.AmdSecMap.TryGetValue(admIdString, out techMd);
+                                if (techMd != null)
+                                    break;
+                            }
+                    }
+
+                    if(techMd == null)
```
tomcrane, line 418: Before, we didn't check this and the next line would throw an exception. Why are we now continuing - if `techMd` is null that seems like an exception condition that should throw here

src/DigitalPreservation/Storage.Repository.Common/Mets/MetsParser.cs
```diff
@@ -387,14 +387,37 @@ private void ProcessChildStructDivs(MetsFileWrapper mets, XElement parent,
...
 
+                    if (techMd == null)
+                    {
+                        //then try the first item in split array if adm id has spaces
+                        //if (admid has spaces)
+                        //then try each part
+                        if (admIdList != null)
+                            foreach (var admIdString in admIdList)
+                            {
+                                lookupMaps.AmdSecMap.TryGetValue(admIdString, out techMd);
```
tomcrane, line 412: This isn't the only place that we call lookupMaps.AmdSecMap.TryGetValue(..

Do we need to deal with spaces in those places too?

src/DigitalPreservation/Storage.Repository.Common/FilenameHelpers.cs
```diff
@@ -0,0 +1,32 @@
...
+        return
+            target
+                .Replace("&", "&amp;")
+                .Replace("<", "&lt;")
+                .Replace(">", "&gt;")
+                .Replace("\"", "&quot;")
+                .Replace("'", "&apos;")
+                .Replace("&amp;amp;", "&amp;")
+                .Replace(" ", "+");
+    }
```
tomcrane, line 31: The last two here can lead to collisions? Two different filenames encoding to the same id string?

Can we have some unit tests for `EscapeXmlCharactersAndSpaces`

```
EscapeXmlCharactersAndSpaces("test 1") => "test+1"
EscapeXmlCharactersAndSpaces("test+1") => "test+1"
```

These yield the same string

And I think 

```
EscapeXmlCharactersAndSpaces("M & S") => "M &amp; S"
EscapeXmlCharactersAndSpaces("M &amp; S") => "M &amp; S"
```

tomcrane (changes requested):
I'm still not sure about this approach.

We need unit and integration tests for:

 - METS file created before the change in ID generation, that is then edited: adding, removing and editing files in a METS that already has ID elements that would now be escaped but are not
 - METS file created after the change

As mentioned in a detail comment I think we need to record some information in the METS about the ID minting strategy. MetsManager can then see this value and act accordingly. But for METS that is either side of the 

The fact that Leeds have some existing production files with spaces means we need to do this.

MetsParser should _not_ look for such a flag, because it needs to always try with and without spaces - but there is more than one place where that happens.


Also - although it wasn't introduced in this PR - I must have missed it in a previous one, these two members of MetsManager are dangerous:

```
    private FileType? File { get; set; }
    private MetsTypeFileSecFileGrp? FileGroup { get; set; }
```

MetsManager is a singleton in application scope and even if that scope changes, I think having these class-level members is dangerous.

tomcrane:
If we do introduce a flag into the METS to say how its IDs were encoded, it should go into the root MODS section, but I'm not sure under what element.

tomcrane:
Superseded by #188

PR #173, opened 2026-04-02
Handle digest mismatches for manually uploaded bagit generated files

brianmac1967:
Ticket: https://digirati.atlassian.net/browse/LPII-69

If the metadata files timestamp changes like after a pipeline run the…n any manually uploaded files content could have changed like siegfried.csv filepaths. This would result in SHA256 digest mismatches as the bagit generated digest values will be no longer valid and thus unable to do an import job. This change checks when the mainfest-sha256.txt was generated and comapres with the file in deposit LastModified date.

Claude suggestions:
[project_pr173_review.md](https://github.com/user-attachments/files/26790670/project_pr173_review.md)

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/Combined/CombinedFile.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Common.Model/Transit/Combined/CombinedFile.cs
@@ -442,6 +442,15 @@ public class CombinedFile(WorkingFile? fileInDeposit, WorkingFile? fileInMets, s
 
     public List<string> GetDistinctDigests()
     {
+        if (FileInDeposit != null && DepositFileFormatMetadata != null && FolderNames.IsMetadata(FileInDeposit.LocalPath))
+        {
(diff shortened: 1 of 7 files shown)


tomcrane:
Changes in last two commits:

  CombinedFile.GetDistinctDigests — the condition now looks up the BagIt DigestMetadata entry directly instead of routing through GetDigestMetadata():

  var bagItEntry = FileInDeposit?.Metadata
      .OfType<DigestMetadata>()
      .FirstOrDefault(m => m.Source == "BagIt");

  if (bagItEntry != null && FileInDeposit!.Modified <= bagItEntry.Timestamp)
      distinctDigests.Add(bagItEntry.Digest ?? string.Empty);

  This fixes three things simultaneously:
  1. Correct timestamp — compares against the BagIt manifest's S3 LastModified, not Max(all sources)
  2. No exception — never calls GetDigestMetadata(), so mismatched BagIt/Siegfried digests no longer throw
  3. Better coverage without pipeline — the BagIt digest is now included for fresh files even when Siegfried hasn't run, so a genuine BagIt/METS
  mismatch is caught regardless of whether the pipeline has run (the second new test documents this as an improvement over the old behaviour)

PR #191, opened 2026-05-15
Active flag switch for Error status only

frank-kerrigan:
Changes

1. Only show on deposit.status == "error"
2. Change guard 
3. Change tests 

Other 

Fix tests failing in windows due to line endings


brianmac1967 (approved):
LGTM

PR #196, opened 2026-06-17
create metadata and ad-hoc folder for deposits exported or not

brianmac1967:
Ticket: https://digirati.atlassian.net/browse/LPII-133
create metadata and ad-hoc folder for deposits exported or not which will alow the deposit to be preserved

### DIFF
--- a/src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/CreateDepositBase.cs
+++ b/src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/CreateDepositBase.cs
@@ -170,16 +170,6 @@ public class CreateDepositBase(
                 createdDeposit.ArchivalGroupExists = true;
             }
 
-            if (!request.Export)
-            {
-                // refresh the file system
-                // The async export will do this at the end of its run, if we exported.
-                // But here we didn't, so just do a quick update (it won't take long)
-                await workspaceManagerFactory.CreateAsync(createdDeposit, true);
-            }
-
-            if (!request.Export) return Result.Ok(createdDeposit);
-
             var wrapperResult = await metsParser.GetMetsFileWrapper(createdDeposit.Files!);
 
             var metadataFolder = wrapperResult.Value?.PhysicalStructure!.FindDirectory(FolderNames.Metadata);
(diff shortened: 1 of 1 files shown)


src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/CreateDepositBase.cs
```diff
@@ -170,16 +170,6 @@ public class CreateDepositBase(
...
-            {
-                // refresh the file system
-                // The async export will do this at the end of its run, if we exported.
-                // But here we didn't, so just do a quick update (it won't take long)
-                await workspaceManagerFactory.CreateAsync(createdDeposit, true);
-            }
-
-            if (!request.Export) return Result.Ok(createdDeposit);
-
             var wrapperResult = await metsParser.GetMetsFileWrapper(createdDeposit.Files!);
```
tomcrane, line 173: Minor (non-blocking): for `template == None` or third-party METS deposits there's no editable METS to write to, so both `CreateFolderInMets` calls below will fail internally — `HandleCreateFolder` → `GetFullMets` returns `NotFound` (no METS) or `BadRequest` (not our `MetsCreatorAgent`). The failed `Result` is discarded, so it's not a crash, but it means two doomed S3 reads + spurious error logs on every such deposit creation (a path that previously never ran this code).

Suggest guarding on `MetsFileWrapper.Editable` (which is exactly `Agent == MetsCreatorAgent`, i.e. the same condition `GetFullMets` requires):

```csharp
var wrapperResult = await metsParser.GetMetsFileWrapper(createdDeposit.Files!);

// Only ensure the metadata/ad-hoc folders when there is a METS file we created and can edit.
// For template=None or third-party METS there is nothing to write to, and HandleCreateFolder
// would fail internally (GetFullMets returns NotFound / BadRequest).
if (wrapperResult.Value is { Editable: true } metsWrapper)
{
    if (metsWrapper.PhysicalStructure!.FindDirectory(FolderNames.Metadata) == null)
        await CreateFolderInMets(FolderNames.Metadata, FolderNames.Metadata, createdDeposit);

    if (metsWrapper.PhysicalStructure!.FindDirectory(FolderNames.MetadataAdHoc) == null)
        await CreateFolderInMets(FolderNames.MetadataAdHoc, FolderNames.AdHoc, createdDeposit);
}
```

The `is { Editable: true }` pattern also tidily handles `wrapperResult.Value` being null on a failed parse, which the current `?.` chain silently tolerated. Behaviour for real managed-METS deposits (the ones this PR fixes) is unchanged.

tomcrane (changes requested):

tomcrane (approved):

PR #176, opened 2026-04-13
add ad-hoc folder to the metadata folder and make non-deletable

brianmac1967:
Ticket:
https://digirati.atlassian.net/browse/LPII-9
add ad-hoc folder to the metadata folder and make non-deletable

PR #197, opened 2026-06-30
Fix/lpii 133/still issue with import job

brianmac1967:
Ticket: https://digirati.atlassian.net/browse/LPII-133
If the METS in the Archival Group deposit is non-editable then do not add the metadata and/or ad-hoc folder as it will cause an issue with the Import Job. I have noticed the issue with import job occurs if you add a file in objects in S3 and refresh storage on a deposit created from an archival Group with a non-editable METS.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
@@ -113,7 +113,7 @@ public class MetsParser(
             }
             catch (Exception e)
             {
-                logger.LogError(e, "Unable to Load Mets File");
+                logger.LogError(e, $"Unable to Load Mets File (file URI: {file.AbsoluteUri}) at location {metsLocation} due to error {e.Message}");
                 return Result.FailNotNull<MetsFileWrapper>(ErrorCodes.UnknownError, e.Message);
             }
         }
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
@@ -146,6 +146,7 @@ public class MetsParser(
             }
             mets.Editable = mets.Agent == Constants.MetsCreatorAgent;
         }
+        logger.LogInformation("Mets {mets}", mets);
         return Result.OkNotNull(mets);
     }
 
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
@@ -113,7 +113,7 @@ public class MetsParser(
             }
             catch (Exception e)
             {
-                logger.LogError(e, $"Unable to Load Mets File (file URI: {file.AbsoluteUri}) at location {metsLocation} due to error {e.Message}");
+                logger.LogError(e, "Unable to Load Mets File (file URI: {FileUri}) at location {MetsLocation}", file.AbsoluteUri, metsLocation);
                 return Result.FailNotNull<MetsFileWrapper>(ErrorCodes.UnknownError, e.Message);
             }
         }
@@ -146,7 +146,6 @@ public class MetsParser(
             }
             mets.Editable = mets.Agent == Constants.MetsCreatorAgent;
         }
-        logger.LogInformation("Mets {mets}", mets);
         return Result.OkNotNull(mets);
     }
 
--- a/src/DigitalPreservation/DigitalPreservation.Workspace/Requests/CreateFolder.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Workspace/Requests/CreateFolder.cs
@@ -80,7 +80,15 @@ public class CreateFolderHandler(
             var dirForMets = request.IsBagItLayout ? dir.ToRootLayout() : dir;
             if (newRootResult.Success)
             {
-                await metsManager.HandleCreateFolder(request.RootUri, dirForMets, request.MetsETag);
+                // Previously this result was discarded and the method always returned Ok(dir) here,
+                // regardless of whether the METS write actually succeeded - e.g. an ETag mismatch
+                // (PreconditionFailed) would silently leave the S3 folder marker created but the METS
+                // entry missing, with the caller told the call succeeded. See LPII-133.
+                var metsResult = await metsManager.HandleCreateFolder(request.RootUri, dirForMets, request.MetsETag);
+                if (metsResult.Failure)
+                {
+                    return Result.Generify<WorkingDirectory?>(metsResult);
+                }
                 return Result.Ok(dir);
             }
             return Result.Generify<WorkingDirectory?>(newRootResult);
--- a/src/DigitalPreservation/Preservation.API.Tests/Workspace/CreateFolderHandlerTests.cs
+++ b/src/DigitalPreservation/Preservation.API.Tests/Workspace/CreateFolderHandlerTests.cs
@@ -1,0 +1,78 @@
+using Amazon.S3;
+using Amazon.S3.Model;
+using DigitalPreservation.Common.Model;
+using DigitalPreservation.Common.Model.Results;
+using DigitalPreservation.Common.Model.Transit;
+using DigitalPreservation.Mets;
+using DigitalPreservation.Workspace.Requests;
+using FakeItEasy;
+using Storage.Repository.Common;
+
+namespace Preservation.API.Tests.Workspace;
+
+public class CreateFolderHandlerTests
+{
+    private readonly IAmazonS3 s3Client = A.Fake<IAmazonS3>();
+    private readonly IStorage storage = A.Fake<IStorage>();
+    private readonly IMetsManager metsManager = A.Fake<IMetsManager>();
+    private readonly CreateFolderHandler handler;
+
+    public CreateFolderHandlerTests()
+    {
+        handler = new CreateFolderHandler(s3Client, storage, metsManager);
+
+        A.CallTo(() => s3Client.PutObjectAsync(A<PutObjectRequest>.Ignored, A<CancellationToken>.Ignored))
+            .Returns(Task.FromResult(new PutObjectResponse { HttpStatusCode = System.Net.HttpStatusCode.Created }));
+
+        A.CallTo(() => s3Client.GetObjectMetadataAsync(A<GetObjectMetadataRequest>.Ignored, A<CancellationToken>.Ignored))
+            .Returns(Task.FromResult(new GetObjectMetadataResponse { LastModified = DateTime.UtcNow }));
+
+        A.CallTo(() => storage.AddToDepositFileSystem(A<Uri>.Ignored, A<WorkingDirectory>.Ignored, A<CancellationToken>.Ignored))
+            .Returns(Task.FromResult(Result.OkNotNull(new WorkingDirectory { LocalPath = "", Name = "root", Modified = DateTime.UtcNow })));
+    }
+
+    // LPII-133: CreateFolderHandler used to discard HandleCreateFolder's result and always return
+    // Result.Ok(dir) regardless, so an ETag mismatch (PreconditionFailed) on the METS write left the S3
+    // folder marker created but the METS entry silently missing - the caller was told the call succeeded.
+    [Fact]
+    public async Task Handle_ReturnsFailure_WhenMetsWriteFails()
+    {
+        A.CallTo(() => metsManager.HandleCreateFolder(A<Uri>.Ignored, A<WorkingDirectory>.Ignored, A<string>.Ignored))
+            .Returns(Task.FromResult(Result.Fail(ErrorCodes.PreconditionFailed, "Supplied ETag did not match METS")));
+
+        var request = new CreateFolder(
+            isBagItLayout: false,
+            rootUri: new Uri("s3://bucket/deposits/dep123/"),
+            name: "metadata",
+            newFolderSlug: "metadata",
+            parent: null,
+            metsETag: "stale-etag");
+
+        var result = await handler.Handle(request, CancellationToken.None);
+
+        result.Failure.Should().BeTrue();
+        result.Success.Should().BeFalse();
+        result.ErrorCode.Should().Be(ErrorCodes.PreconditionFailed);
+    }
+
+    [Fact]
+    public async Task Handle_ReturnsSuccess_WhenMetsWriteSucceeds()
+    {
+        A.CallTo(() => metsManager.HandleCreateFolder(A<Uri>.Ignored, A<WorkingDirectory>.Ignored, A<string>.Ignored))
+            .Returns(Task.FromResult(Result.Ok()));
+
+        var request = new CreateFolder(
+            isBagItLayout: false,
+            rootUri: new Uri("s3://bucket/deposits/dep123/"),
+            name: "metadata",
+            newFolderSlug: "metadata",
+            parent: null,
+            metsETag: "current-etag");
+
+        var result = await handler.Handle(request, CancellationToken.None);
+
+        result.Success.Should().BeTrue();
+        result.Value.Should().NotBeNull();
+        result.Value!.Name.Should().Be("metadata");
+    }
+}
--- a/src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/CreateDepositBase.cs
+++ b/src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/CreateDepositBase.cs
@@ -180,7 +180,9 @@ public class CreateDepositBase(
             var wrapperResult = await metsParser.GetMetsFileWrapper(createdDeposit.Files!);
 
             var editable = wrapperResult?.Value?.Editable;
+            var agent = wrapperResult?.Value?.Agent;
             logger.LogInformation("METS is editable: {editable}", editable);
+            logger.LogInformation("METS agent name: {agentName}", agent);
 
             // Only ensure the metadata/ad-hoc folders when there is a METS file we created and can edit.
             // For template=None or third-party METS there is nothing to write to, and HandleCreateFolder
--- a/src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/CreateDepositBase.cs
+++ b/src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/CreateDepositBase.cs
@@ -209,6 +209,7 @@ public class CreateDepositBase(
                     var result = await CreateFolderInMets(FolderNames.Metadata, FolderNames.Metadata, createdDeposit);
 
                     logger.LogInformation("Creating Metadata folder in mets result success: {resultSuccess}", result.Success);
+                    logger.LogInformation("Creating Metadata folder in mets result message: {resultSuccess}", result.CodeAndMessage());
                 }
 
                 var metadataAdhocFolderExists = metsWrapper?.PhysicalStructure!.FindDirectory(FolderNames.MetadataAdHoc) != null;
@@ -221,6 +222,7 @@ public class CreateDepositBase(
                     var result = await CreateFolderInMets(FolderNames.MetadataAdHoc, FolderNames.AdHoc, createdDeposit);
 
                     logger.LogInformation("Creating Metadata ad-hoc folder in mets result success: {resultSuccess}", result.Success);
+                    logger.LogInformation("Creating Metadata ad-hoc folder in mets result message: {resultSuccess}", result.CodeAndMessage());
                 }
             }
 
--- a/src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/CreateDepositBase.cs
+++ b/src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/CreateDepositBase.cs
@@ -7,7 +7,6 @@ using DigitalPreservation.Common.Model.Transit;
 using DigitalPreservation.Core.Auth;
 using DigitalPreservation.Mets;
 using DigitalPreservation.Workspace;
-using DigitalPreservation.XmlGen.Mets;
 using LeedsDlipServices.Identity;
 using Preservation.API.Data;
 using Preservation.API.Mutation;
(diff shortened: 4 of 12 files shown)


tomcrane:
Thanks for digging into this one — the investigation here helped me understand what's actually going on. Having traced it through, I think the underlying problem is slightly different from what this PR addresses, and there's a simpler fix available. Here's my analysis.

## The underlying problem

This is a **timing/ordering bug introduced by the ad-hoc folder feature (#176)**, with two legs:

1. **S3 scaffolding is unconditional.** `Storage.GetWorkingFilesLocation` creates `metadata/` and `metadata/ad-hoc/` folder markers in S3 for every RootLevel/BagIt deposit, at the very start of `CreateDepositBase.HandleBase`, before anything is known about the METS.

2. **The METS-side half runs too early.** The matching step — "if the METS is editable and lacks those folders, add them to the METS" — runs at the end of `HandleBase`, against the deposit's METS *at creation time*. But when a deposit is created from an existing archival group with `Export=true` (the UI's clone flow), the export is queued and runs asynchronously — the METS hasn't arrived in the deposit yet. `GetMetsFileWrapper` returns an empty wrapper with `Editable=false`, so the folder-creation block is silently skipped. The AG's METS lands later, and nothing ever reconciles it.

So for any AG whose stored METS doesn't already contain `metadata/ad-hoc` — anything preserved before #176 ("historic"), or anything with a third-party METS — the cloned deposit ends up with the folders in S3 but not in the METS, and `GetDiffImportJob` hard-fails with `Could not find folder metadata/ad-hoc in METS file`. This is also why manually doing "add to METS" on dev appeared to fix it — that's exactly the reconciliation the code never performs.

Note the system *already has* the right reconciliation point: `GetDepositBase.GetDeposit` is where an `Exporting` deposit is detected as finished (`wasExportingAndNowFinished`) and the workspace refreshed.

## Concerns with the approach in this PR

- **The `using DigitalPreservation.XmlGen.Mets;` in `CreateDepositBase` is dead code** — no type from that namespace is referenced anywhere in the file (`GetMetsFileWrapperFromXDocument` already existed on `IMetsParser`). It can just be deleted, and I'd really rather not see XmlGen appear as a conceptual dependency of deposit creation. But the deeper issues:

- **It only fixes half the bug.** `ShouldCreateMetadataFolders` downloads and parses the AG's METS (an extra storage-map fetch plus a full METS stream + parse on every deposit-from-AG creation — METS files can be large) purely to check `Editable`. If the AG METS is *non-editable*, skipping the S3 scaffolding fixes the third-party-METS case. But if the AG METS is **editable and merely historic** (our own AGs preserved before #176 — the case the ticket is named after), it still scaffolds the S3 folders and then calls `CreateFolderInMets` against a deposit whose METS *still hasn't been exported* — `HandleCreateFolder` fails, the failure is logged and ignored, and the diff import job breaks exactly as before.

- **New NRE risk.** The original `wrapperResult.Value is { Editable: true }` pattern was null-safe. `wrapperResult.Value.Editable || createMetadataFolders` throws if the parse failed (`FailNotNull` returns a null `Value`).

- **The original guard's intent is undone**: with `|| createMetadataFolders`, `CreateFolderInMets` is now attempted even when there's no editable METS to write to — the exact situation the deleted comment warned about — with failures swallowed.

- **Leftover debug noise**: the many `LogInformation` lines narrating booleans, `logger.LogInformation("Mets {mets}", mets)` inside `MetsParser` on *every* parse system-wide, a malformed template (`"Directory for METS {deposit.MetsETag!}"`), and an interpolated `LogError` that duplicates `e.Message`. And no tests.

## Suggested alternative

The simple, complete shape of the fix, using signals already in hand (no METS fetching at creation time):

1. **Don't scaffold `metadata/` folders in S3 when the deposit is for an existing AG** — `archivalGroupExists is true` is already known in `HandleBase` when `GetWorkingFilesLocation` is called. The AG's exported content should define the deposit's content. (The non-export clone path copies the METS synchronously, so the existing end-of-`HandleBase` editable check still works there.)

2. **Reconcile when the export actually completes** — in `GetDepositBase`, inside the existing `wasExportingAndNowFinished` branch: parse the now-present METS, and if it's editable and lacks `metadata`/`metadata/ad-hoc`, run the same ensure-folders logic (S3 + METS together). Editability is then read from the deposit's own METS file that's actually there — no cross-service AG-METS fetch, no XmlGen, no guessing.

A worthwhile complementary hardening: `GetDiffImportJob` arguably shouldn't hard-fail on an **empty** deposit directory that's absent from the METS — an empty folder has no preservable payload, so skipping it (with a warning) would make the diff robust against this whole class of drift. That alone would stop the 422 in every variant, though it treats the symptom rather than the ordering problem, so I'd pair it with (1)+(2) rather than use it alone.

One thing to verify whichever route we take: whether `metsManager.HandleCreateFolder` also writes the S3 folder marker or only the METS entry — if METS-only, step 2 should create both so the workspace view stays consistent.

*(Analysis assisted by Claude Code)*

brianmac1967:
> Thanks for digging into this one — the investigation here helped me understand what's actually going on. Having traced it through, I think the underlying problem is slightly different from what this PR addresses, and there's a simpler fix available. Here's my analysis.
> 
> ## The underlying problem
> This is a **timing/ordering bug introduced by the ad-hoc folder feature (#176)**, with two legs:
> 
> 1. **S3 scaffolding is unconditional.** `Storage.GetWorkingFilesLocation` creates `metadata/` and `metadata/ad-hoc/` folder markers in S3 for every RootLevel/BagIt deposit, at the very start of `CreateDepositBase.HandleBase`, before anything is known about the METS.
> 2. **The METS-side half runs too early.** The matching step — "if the METS is editable and lacks those folders, add them to the METS" — runs at the end of `HandleBase`, against the deposit's METS _at creation time_. But when a deposit is created from an existing archival group with `Export=true` (the UI's clone flow), the export is queued and runs asynchronously — the METS hasn't arrived in the deposit yet. `GetMetsFileWrapper` returns an empty wrapper with `Editable=false`, so the folder-creation block is silently skipped. The AG's METS lands later, and nothing ever reconciles it.
> 
> So for any AG whose stored METS doesn't already contain `metadata/ad-hoc` — anything preserved before #176 ("historic"), or anything with a third-party METS — the cloned deposit ends up with the folders in S3 but not in the METS, and `GetDiffImportJob` hard-fails with `Could not find folder metadata/ad-hoc in METS file`. This is also why manually doing "add to METS" on dev appeared to fix it — that's exactly the reconciliation the code never performs.
> 
> Note the system _already has_ the right reconciliation point: `GetDepositBase.GetDeposit` is where an `Exporting` deposit is detected as finished (`wasExportingAndNowFinished`) and the workspace refreshed.
> 
> ## Concerns with the approach in this PR
> * **The `using DigitalPreservation.XmlGen.Mets;` in `CreateDepositBase` is dead code** — no type from that namespace is referenced anywhere in the file (`GetMetsFileWrapperFromXDocument` already existed on `IMetsParser`). It can just be deleted, and I'd really rather not see XmlGen appear as a conceptual dependency of deposit creation. But the deeper issues:
> * **It only fixes half the bug.** `ShouldCreateMetadataFolders` downloads and parses the AG's METS (an extra storage-map fetch plus a full METS stream + parse on every deposit-from-AG creation — METS files can be large) purely to check `Editable`. If the AG METS is _non-editable_, skipping the S3 scaffolding fixes the third-party-METS case. But if the AG METS is **editable and merely historic** (our own AGs preserved before [add ad-hoc folder to the metadata folder and make non-deletable #176](https://github.com/digirati-co-uk/digital-preservation/pull/176) — the case the ticket is named after), it still scaffolds the S3 folders and then calls `CreateFolderInMets` against a deposit whose METS _still hasn't been exported_ — `HandleCreateFolder` fails, the failure is logged and ignored, and the diff import job breaks exactly as before.
> * **New NRE risk.** The original `wrapperResult.Value is { Editable: true }` pattern was null-safe. `wrapperResult.Value.Editable || createMetadataFolders` throws if the parse failed (`FailNotNull` returns a null `Value`).
> * **The original guard's intent is undone**: with `|| createMetadataFolders`, `CreateFolderInMets` is now attempted even when there's no editable METS to write to — the exact situation the deleted comment warned about — with failures swallowed.
> * **Leftover debug noise**: the many `LogInformation` lines narrating booleans, `logger.LogInformation("Mets {mets}", mets)` inside `MetsParser` on _every_ parse system-wide, a malformed template (`"Directory for METS {deposit.MetsETag!}"`), and an interpolated `LogError` that duplicates `e.Message`. And no tests.
> 
> ## Suggested alternative
> The simple, complete shape of the fix, using signals already in hand (no METS fetching at creation time):
> 
> 1. **Don't scaffold `metadata/` folders in S3 when the deposit is for an existing AG** — `archivalGroupExists is true` is already known in `HandleBase` when `GetWorkingFilesLocation` is called. The AG's exported content should define the deposit's content. (The non-export clone path copies the METS synchronously, so the existing end-of-`HandleBase` editable check still works there.)
> 2. **Reconcile when the export actually completes** — in `GetDepositBase`, inside the existing `wasExportingAndNowFinished` branch: parse the now-present METS, and if it's editable and lacks `metadata`/`metadata/ad-hoc`, run the same ensure-folders logic (S3 + METS together). Editability is then read from the deposit's own METS file that's actually there — no cross-service AG-METS fetch, no XmlGen, no guessing.
> 
> A worthwhile complementary hardening: `GetDiffImportJob` arguably shouldn't hard-fail on an **empty** deposit directory that's absent from the METS — an empty folder has no preservable payload, so skipping it (with a warning) would make the diff robust against this whole class of drift. That alone would stop the 422 in every variant, though it treats the symptom rather than the ordering problem, so I'd pair it with (1)+(2) rather than use it alone.
> 
> One thing to verify whichever route we take: whether `metsManager.HandleCreateFolder` also writes the S3 folder marker or only the METS entry — if METS-only, step 2 should create both so the workspace view stays consistent.
> 
> _(Analysis assisted by Claude Code)_

What changed:

1. CreateDepositBase.cs — replaced ShouldCreateMetadataFolders() (which fetched and parsed the AG's METS via the Storage API just to check editability — his "only fixes half the bug" + "extra storage-map fetch plus a full METS stream + parse" concern) with a one-line check: archivalGroupExists is not true. This is his point 1 — the signal was already available for free. Also restored the original null-safe wrapperResult.Value is { Editable: true } guard (removing the NRE risk from wrapperResult.Value.Editable || createMetadataFolders), removed the dead using DigitalPreservation.XmlGen.Mets;, and stripped all the debug-noise LogInformation calls.
2. GetDepositBase.cs — added the reconciliation step at his point 2: inside the existing wasExportingAndNowFinished branch, once the real METS has arrived, it checks whether that METS is editable and lacks metadata/metadata/ad-hoc, and if so creates them via WorkspaceManager.CreateFolder(...) — the same path the interactive "create folder" feature uses, which writes the S3 marker, the deposit file-system cache, and the METS entry together. I confirmed this matters: HandleCreateFolder alone (what CreateFolderInMets calls) only touches METS, so using it in isolation here would've just inverted the original bug (METS has the folder, S3 doesn't).
3. MetsParser.cs — fixed the two logging complaints: the logger.LogInformation("Mets {mets}", mets) that ran on every METS parse system-wide is gone; the interpolated LogError duplicating e.Message is now structured ({FileUri}, {MetsLocation}) without the duplication.
4. Tests — added two tests in StorageTests.cs directly verifying GetWorkingFilesLocation's createMetadataFolders parameter actually skips/creates the S3 folder markers as intended. I looked into testing the GetDepositBase reconciliation too, but WorkspaceManager/WorkspaceManagerFactory are concrete, non-virtual classes with no interface — not unit-testable in isolation without a larger refactor, so that part still lacks direct test coverage (worth flagging back to your colleague).

 _(Code changes assisted by Claude Code)_

tomcrane:
Re-reviewed the updated branch — **this is now the right shape**, thanks for the rework. It addresses everything from the first review: `createMetadataFolders` is derived from `archivalGroupExists` (no AG-METS fetching), the reconciliation happens in `GetDepositBase`'s `wasExportingAndNowFinished` branch via `WorkspaceManager.CreateFolder` (the correct choice — and good spot that `HandleCreateFolder` alone is METS-only), the null-safe `is { Editable: true }` guard is restored, the dead `XmlGen.Mets` using is gone, the debug logging is stripped, the `MetsParser` log lines are structured properly, and there are unit tests for the `Storage` change.

I also verified the concern I'd left open earlier: METS-only folders (which the non-export clone path now produces, since we no longer scaffold S3 folders for existing AGs) are safe — `CombinedBuilder.Build` takes the *union* of deposit and METS directories, and `GetDiffImportJob`'s container check only requires `DirectoryInMets`, so they simply become `ContainersToAdd`.

## One real bug remaining: the "both folders missing" case recreates LPII-133

When the exported METS lacks *both* `metadata` and `metadata/ad-hoc`, `EnsureMetadataFoldersAfterExport` makes two sequential `CreateFolder` calls, and the second one breaks:

1. `deposit.MetsETag` is set once, from the freshly fetched wrapper. `WorkspaceManager` holds the same `Deposit` reference and passes `Deposit.MetsETag` into each mediator `CreateFolder` request — and nothing ever refreshes it after a write.
2. The first call (`metadata`) succeeds and rewrites the METS → its S3 ETag changes.
3. The second call (`ad-hoc`) PUTs the S3 folder marker *first* (`CreateFolderHandler`), then calls `metsManager.HandleCreateFolder` with the now-stale ETag. `S3MetsStorage.GetFullMets` sends that ETag as `EtagToMatch` on the S3 read → `PreconditionFailed` → the METS update fails.
4. Worse, the failure is invisible: `CreateFolderHandler` discards `HandleCreateFolder`'s result and returns `Result.Ok` regardless, so even the `LogWarning` in `EnsureMetadataFoldersAfterExport` never fires.

Net result: `metadata/ad-hoc/` marker in S3, absent from METS — exactly the original bug state, reported as success, and never retried since the reconciliation runs only once on the `Exporting → New` transition. The common historic case (AG has `metadata/` from pipeline output, only `ad-hoc` missing) is a single call and works fine — which is probably why testing didn't catch this variant.

Two fixes, ideally both:

- In `EnsureMetadataFoldersAfterExport`: after the first successful `CreateFolder`, re-fetch the wrapper (`GetMetsFileWrapper(deposit.Files!, parse: false)`) and update `deposit.MetsETag` before the ad-hoc call.
- In `CreateFolderHandler`: stop swallowing `HandleCreateFolder`'s result. This is pre-existing (harmless until now because `CreateDepositBase` passes a null ETag, which disables the guard), but the new reconciliation is the first ETag-guarded double-call path, so it's now load-bearing — and it's the same silent-failure class we flagged in #175.

## Minor notes

- The reconciliation passes `callerIdentity: null` to `CreateFolder`, which bypasses the lock check entirely (`GetOtherLockOwner` returns null for a blank caller). Probably intended for a system action — fine, but worth a code comment saying so.
- No test coverage for the reconciliation itself — understood re: `WorkspaceManager`/factory not being mockable without a refactor, and the E2E test covers the main scenario. But the both-missing variant above is precisely the kind of thing a unit test would have caught, so worth revisiting when those classes get interfaces.
- SonarQube reports 8 new issues on the PR — worth a glance before merge.

*(Analysis assisted by Claude Code)*

brianmac1967:
> Re-reviewed the updated branch — **this is now the right shape**, thanks for the rework. It addresses everything from the first review: `createMetadataFolders` is derived from `archivalGroupExists` (no AG-METS fetching), the reconciliation happens in `GetDepositBase`'s `wasExportingAndNowFinished` branch via `WorkspaceManager.CreateFolder` (the correct choice — and good spot that `HandleCreateFolder` alone is METS-only), the null-safe `is { Editable: true }` guard is restored, the dead `XmlGen.Mets` using is gone, the debug logging is stripped, the `MetsParser` log lines are structured properly, and there are unit tests for the `Storage` change.
> 
> I also verified the concern I'd left open earlier: METS-only folders (which the non-export clone path now produces, since we no longer scaffold S3 folders for existing AGs) are safe — `CombinedBuilder.Build` takes the _union_ of deposit and METS directories, and `GetDiffImportJob`'s container check only requires `DirectoryInMets`, so they simply become `ContainersToAdd`.
> 
> ## One real bug remaining: the "both folders missing" case recreates LPII-133
> When the exported METS lacks _both_ `metadata` and `metadata/ad-hoc`, `EnsureMetadataFoldersAfterExport` makes two sequential `CreateFolder` calls, and the second one breaks:
> 
> 1. `deposit.MetsETag` is set once, from the freshly fetched wrapper. `WorkspaceManager` holds the same `Deposit` reference and passes `Deposit.MetsETag` into each mediator `CreateFolder` request — and nothing ever refreshes it after a write.
> 2. The first call (`metadata`) succeeds and rewrites the METS → its S3 ETag changes.
> 3. The second call (`ad-hoc`) PUTs the S3 folder marker _first_ (`CreateFolderHandler`), then calls `metsManager.HandleCreateFolder` with the now-stale ETag. `S3MetsStorage.GetFullMets` sends that ETag as `EtagToMatch` on the S3 read → `PreconditionFailed` → the METS update fails.
> 4. Worse, the failure is invisible: `CreateFolderHandler` discards `HandleCreateFolder`'s result and returns `Result.Ok` regardless, so even the `LogWarning` in `EnsureMetadataFoldersAfterExport` never fires.
> 
> Net result: `metadata/ad-hoc/` marker in S3, absent from METS — exactly the original bug state, reported as success, and never retried since the reconciliation runs only once on the `Exporting → New` transition. The common historic case (AG has `metadata/` from pipeline output, only `ad-hoc` missing) is a single call and works fine — which is probably why testing didn't catch this variant.
> 
> Two fixes, ideally both:
> 
> * In `EnsureMetadataFoldersAfterExport`: after the first successful `CreateFolder`, re-fetch the wrapper (`GetMetsFileWrapper(deposit.Files!, parse: false)`) and update `deposit.MetsETag` before the ad-hoc call.
> * In `CreateFolderHandler`: stop swallowing `HandleCreateFolder`'s result. This is pre-existing (harmless until now because `CreateDepositBase` passes a null ETag, which disables the guard), but the new reconciliation is the first ETag-guarded double-call path, so it's now load-bearing — and it's the same silent-failure class we flagged in [Run bagit in pipeline and upload bagit files to deposit in s3. #175](https://github.com/digirati-co-uk/digital-preservation/pull/175).
> 
> ## Minor notes
> * The reconciliation passes `callerIdentity: null` to `CreateFolder`, which bypasses the lock check entirely (`GetOtherLockOwner` returns null for a blank caller). Probably intended for a system action — fine, but worth a code comment saying so.
> * No test coverage for the reconciliation itself — understood re: `WorkspaceManager`/factory not being mockable without a refactor, and the E2E test covers the main scenario. But the both-missing variant above is precisely the kind of thing a unit test would have caught, so worth revisiting when those classes get interfaces.
> * SonarQube reports 8 new issues on the PR — worth a glance before merge.
> 
> _(Analysis assisted by Claude Code)_

1. CreateFolder.cs — CreateFolderHandler no longer discards HandleCreateFolder's result; a failed METS write (e.g. stale-ETag PreconditionFailed) now correctly propagates as a failure instead of the caller being told the call succeeded while the S3 folder marker exists but METS doesn't know about it. Added regression tests: CreateFolderHandlerTests.cs (2 tests, both passing) — confirms failure propagates and success still works.
2. GetDepositBase.cs — when both metadata and metadata/ad-hoc need reconciling after export, the ETag is re-fetched between the two CreateFolder calls so the second call doesn't fail its precondition check against a now-stale ETag from before the first write.
3. Added the callerIdentity: null explanatory comment Tom requested.

tomcrane:
Third pass over the two new commits — **all three findings from the last review are properly fixed, and I couldn't break the new code. This looks ready to me.**

## Verified fixes

1. **`CreateFolderHandler` now propagates the METS write failure**, with two focused unit tests (fail on `PreconditionFailed`, succeed on Ok). I swept all three call sites for regressions from this behaviour change:
   - **UI** (`OnPostCreateFolder`): the new-folder trigger in `_RenderCombinedDirectoryAsTableRows` is gated on `Editable: true`, so users can only reach `CreateFolder` on an editable METS — any error now surfaced to them is a genuine one that *should* be shown.
   - **Pipeline** (`CreateMetadataSubFolderOnS3`): the result list is only logged; the job fails only when the list is *empty*, so a propagated failure changes logs, not behaviour.
   - **Reconciliation** (`EnsureMetadataFoldersAfterExport`): failures land in the `LogWarning`s as intended — which now actually fire, unlike before.

2. **ETag refresh between the two reconciliation calls** — correctly guarded by `needsMetadata && needsAdHoc` (only re-fetch when the first write happened), `parse: false` is right, and the updated `deposit.MetsETag` flows into `WorkspaceManager` via the shared `Deposit` reference. If the re-fetch itself fails, the ad-hoc call fails its precondition and is logged — and any orphaned S3 marker it leaves is now benign because of fix 3. Nice defence in depth.

3. **Diff-time hardening** — empty deposit-only folders are skipped with a log line instead of failing the whole diff; folders with real content still hard-fail (correctly — and their files would already have failed the earlier `FileInMets` check anyway). `HasAnyDescendantFile` recursion is right.

CI: build and tests green; the latest SonarCloud analysis for this PR is **Quality Gate passed, 0 new issues** (the earlier 8 are cleared). The one red "SonarCloud Code Analysis" check links to the *main branch* dashboard rather than this PR, so it looks like a stale branch-level check — worth a re-run if it blocks merge.

## Two observations — neither blocking, just conscious-decision items

- **The skipped empty folder still ends up in `ContainersToAdd`.** The `continue` in the container loop leaves the container in `sourceContainers`, and `PopulateDiffTasks` sweeps any source container not already in the AG into `ContainersToAdd` — so an empty, METS-unlisted folder does get preserved into the Archival Group. Round-tripping is stable (next export brings it back empty → skipped again), so this is safe; it's just worth deciding deliberately whether unlisted empty folders *should* be preserved. If the answer is no, the precise change is to exclude skipped paths from `ContainersToAdd` only — do **not** remove them from `sourceContainers`, because `ContainersToDelete` is computed from absence in that list and removal would wrongly delete existing empty containers on re-preservation.
- **`CreateFolderHandler` still PUTs the S3 marker before the METS write**, so a METS failure leaves an orphaned empty marker. Now benign at diff time and reported to the caller, so I don't think reordering or rollback is worth the complexity — just noting the residual asymmetry.

*(Analysis assisted by Claude Code)*

tomcrane (approved):

PR #206, opened 2026-07-28
Sonar cleanup: repo-wide mechanical fixes + stronger tests

tomcrane:
## What

Follow-up to #203: the same mechanical-only treatment applied to the rest of main. Local SonarAnalyzer warnings drop **620 → 241** (unique issues 570 → 191; the remainder are accepted buckets and deliberately-skipped rules, below).

**402 warnings targeted, 378 fixed** across 101 files (+544/−642):

- **Logging hygiene** (~230): PascalCase placeholders, interpolation/concat → message templates, exceptions passed to loggers, correct `ILogger<T>` categories, two broken placeholders (`{}`, `{di.Name}`) fixed — note the `QueueImportJob` one previously swallowed its argument, so that log line's rendered output actually improves
- **Dead weight**: commented-out code, unused private members/fields/locals/params (reflection- and serializer-checked first)
- **Mechanical style**: private-only `static`, `readonly` fields, redundant booleans, `First()`→`[0]`, char-overload `StartsWith` (ASCII literals only), interface-conformance defaults and parameter names (better name wins — declaration side renamed where it had the worse name)

## Tests strengthened

- 14 previously assertion-free PREMIS/MODS tests now assert real round-trip values (digests, PRONOM keys, patched sizes)
- New pinning tests: `AwsChecksum.FromBase64ToHex`, `StringUtils.BuildPath`
- Full suite: **532 passed, 0 failed** (up from 527)

## Deliberately NOT fixed

- Behavior-changing rules: S112 (exception types), S1104 (public field→property changes JSON shape), S6580 (culture-sensitive parsing), plus 4 logic rewrites in untested Pipeline.API paths and 3 `!` removals the C# compiler disagrees with
- ~12 S2325 'make static' on public members of DI-instantiated managers (`PremisManager`/`PremisManagerExif`/`PremisEventManager`, `WorkspaceManager.ValidateImportJob`) and JSON-serialized `Type` properties on queue messages — **suggest bulk-accepting these in SonarCloud**
- Accepted buckets from #203 (S1075/S5332/S1135), cognitive complexity (by design — not a mechanical concern)

## Notes for reviewers

- `Dlcs` lost an unread `ILogger` ctor param (public signature; only caller is Registrant, updated)
- `Deposit.cshtml.cs` `OnGet` still declares `readFromStorage`/`writeToStorage` query params but `BindDeposit` never read them — they were already inert; consider removing from the route contract separately
- A latent NRE was found (not fixed, out of scope): `PremisManager.Read` on a PREMIS object with PronomKey-only format (null `FormatDesignation`) — documented in a test comment in `Premis_Test.cs`
- Structured-log property names change casing (`{status}`→`{Status}`): dashboards/alerts querying properties case-sensitively need updating


src/DigitalPreservation/Storage.API/Features/Import/SqsImportJobQueue.cs
```diff
@@ -65,7 +65,7 @@ public async ValueTask<string> DequeueRequest(CancellationToken cancellationToke
                 foreach (var message in response.Messages!)
                 {
                     if (cancellationToken.IsCancellationRequested) return string.Empty;
-                    logger.LogDebug("Received SQS message {messageBody}", message.Body);
+                    logger.LogDebug("Received SQS message {MessageBody}", message.Body);
```
brianmac1967: **SqsImportJobQueue dequeue has no timeout guard**

`DequeueRequest`'s `GetQueueUrlAsync`/`ReceiveMessageAsync` calls have no bounded timeout, unlike the sibling `SqsPipelineQueue.DequeueRequest` which this same PR fixed for the identical documented hang bug.

`SqsPipelineQueue.cs` (also touched by this PR) adds a 30s `CancellationTokenSource` around the same two calls, citing LPII-135: if the underlying SQS connection dies without erroring, these calls can hang indefinitely with no exception, wedging the single-threaded consumer forever. `SqsImportJobQueue.cs` was edited in this same PR (the log-template casing on this line) but never got the same guard, so the Storage.API import-job consumer can still wedge forever on the same failure class.

src/DigitalPreservation/Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs
```diff
@@ -116,16 +116,16 @@ private async Task<Result<LogPipelineStatusResult>> UpdateAnyJobStatus(
 
         if (updateResult is { Success: true, Value: not null } && errors.IsNullOrWhiteSpace())
         {
-            logger.LogInformation("Job {jobIdentifier} status updated: {status}", jobId, status);
+            logger.LogInformation("Job {JobIdentifier} status updated: {Status}", jobId, status);
```
brianmac1967, line 119: **Log field casing changed from `status` to `Status`**

Structured-log property casing changed from `{status}` to `{Status}` (and similar) on the job-status-update log line.

Any downstream CloudWatch Logs Insights query, metric filter, or alert rule that pattern-matches the lowercase `status` field name (case-sensitive in most log query engines) silently stops matching new log entries after deploy, potentially going dark on job-failure alerting. This is disclosed by the author as intentional in the PR description, but worth confirming dashboards/alerts were actually updated before merge.

src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/LockDeposit.cs
```diff
@@ -26,15 +26,12 @@ public async Task<Result> Handle(LockDeposit request, CancellationToken cancella
...
-                return Result.Fail(ErrorCodes.Conflict, "Deposit is locked by " + entity.LockedBy);
-            }
+            return Result.Fail(ErrorCodes.Conflict, "Deposit is locked by " + entity.LockedBy);
         }
         var callerIdentity = request.User.GetCallerIdentity();
-        logger.LogInformation("Locking deposit {id} for user {user}", request.Id, callerIdentity);
+        logger.LogInformation("Locking deposit {Id} for user {User}", request.Id, callerIdentity);
         entity.LockedBy = callerIdentity;
         entity.LockDate = DateTime.UtcNow;
         await dbContext.SaveChangesAsync(cancellationToken);
```
brianmac1967, line 37: **LockDeposit lacks the try/catch its unlock twin just got**

`LockDepositHandler.Handle` calls `dbContext.SaveChangesAsync` with no try/catch, while its twin `DeleteDepositLockHandler.Handle` in the same PR got a new try/catch that converts a save failure into `Result.Fail`.

If `SaveChangesAsync` throws (e.g. a `DbUpdateException` from concurrent modification or a transient DB error) while locking a deposit, the exception propagates unhandled instead of returning a clean `Result.Fail` like the unlock path now does — an inconsistency introduced by this PR's own change to the unlock handler.

src/DigitalPreservation/Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs
```diff
@@ -543,7 +543,7 @@ private async Task<ProcessPipelineResult> ExecuteBrunnhilde(ExecutePipelineJob r
             var metsResult = await AddObjectsToMets(request, depositPath);
 
             if (metsResult.Failure)
-                logger.LogInformation("Issue adding objects to METS in pipeline run: {error}", metsResult.ErrorMessage);
+                logger.LogInformation("Issue adding objects to METS in pipeline run: {Error}", metsResult.ErrorMessage);
 
             var start = DateTime.Now;
             var lockReleased = false;
```
brianmac1967, line 549: **Lock-release loop uses external mutable flag**

Lock-release retry loop tracks success via an external mutable `lockReleased` flag set inside one of two break paths and read ~40 lines later, instead of the loop being extracted into a method that returns the result directly. This pattern predates this PR, but this PR does touch the surrounding logging, so flagging while here.

A future edit that adds a third break/exit path to the `while(true)` loop can easily forget to set `lockReleased`, silently making the post-loop check wrong (treating a released lock as not-released, or vice versa). Extracting to a private `TryReleaseLockWithRetries(...)` returning `bool` removes the flag and the risk entirely.

src/DigitalPreservation/Preservation.API/Features/ImportJobs/Requests/GetDiffImportJob.cs
```diff
@@ -246,12 +243,12 @@ public async Task<Result<ImportJob>> Handle(GetDiffImportJob request, Cancellati
         var binariesToCheck = new List<Binary>();
         binariesToCheck.AddRange(importJob.BinariesToAdd);
         binariesToCheck.AddRange(importJob.BinariesToPatch);
-        foreach (var binary in binariesToCheck)
+        foreach (var binaryId in binariesToCheck.Select(binary => binary.Id))
```
brianmac1967, line 246: **`foreach->Select(Id)` indirection, no real benefit**

`foreach (var binaryId in binariesToCheck.Select(binary => binary.Id))` replaces a plain `foreach (var binary in binariesToCheck)` here and similarly in `Storage.API/Features/Import/Requests/ExecuteImportJob.cs:310`, adding indirection with no benefit since both loop bodies are multi-statement and still need a null-forgiving `binaryId!`.

Same line count, one extra closure/projection, and the same null-forgiving operator still needed at the use sites — a reader has to mentally re-substitute 'binaryId is binary.Id', and a future edit needing another property of `binary` forces a second lookup. Reverting to `foreach (var binary in ...)` matches the surrounding loops in the same files.

brianmac1967 (commented):
Automated review findings (correctness + reuse/simplification/efficiency). Note: 3 additional findings from the initial pass turned out to be pre-existing code already on `main` rather than changes introduced by this PR, so they're omitted here.

src/DigitalPreservation/Storage.API/Features/Import/SqsImportJobQueue.cs
tomcrane: Agreed - this was the strongest finding of the review. Applied in 9c87615: same 30s bounded-timeout pattern as SqsPipelineQueue, covering both `GetQueueUrlAsync` (in `EnsureOptions`, which previously swallowed resolution failures and left `queueUrl` null) and `ReceiveMessageAsync`. The enclosing try/catch matters here too: `ImportJobExecutorService` has no catch of its own, so before this change any receive error would have stopped the BackgroundService outright. One attribution note: the SqsPipelineQueue guard actually landed via PR #209 and shows here through the merge of main - but the failure class applies to this consumer identically.

src/DigitalPreservation/Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs
tomcrane: Fair check - it is disclosed in the PR description, and the remaining action is operational, not code: before this deploys, CloudWatch Logs Insights queries, metric filters and alarms that reference lowercase property names (`status`, `jobId`, etc.) need reviewing. Leaving this thread open as the pre-merge checklist item for that confirmation.

src/DigitalPreservation/Preservation.API/Features/ImportJobs/Requests/GetDiffImportJob.cs
tomcrane: Declining the revert: the `foreach (var x in y.Select(...))` shape exists to satisfy Sonar S3267 (loop uses only a projection of the element), which is the point of this PR. The costs you list are real but trivial, and reverting reintroduces two findings. If the team prefers the plain `foreach`, the coherent move is to accept S3267 in SonarCloud (alongside the other bulk-accepts proposed in the PR description) and revert both sites together - not to carry the finding.

src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/LockDeposit.cs
tomcrane: True but declining the try/catch: the unlock handler's guard came from PR #209 (via the merge), and exists specifically because ExecutePipelineJob's retry loop needs a `Result.Fail` it can retry on. LockDeposit is only reached from controller paths, where an unexpected `DbUpdateException` surfacing as a 500 via middleware is a reasonable outcome - wrapping it in a broad catch returning UnknownError doesn't buy much. This comment did surface a real miss though: the catch block #209 added logs with lowercase `{id}` (DeleteDepositLock.cs:40), which escaped this PR's PascalCase sweep because it merged in afterwards - fixed in 9c87615, and a sweep found no other merge-introduced instances.

src/DigitalPreservation/Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs
tomcrane: Reasonable refactor, deferring: as noted the pattern predates this PR, and this lock-release region was just carefully rewritten for two real bugs (Seconds vs TotalSeconds, the swallowed release failure). Restructuring untested Pipeline.API lock logic again purely for style, on a PR that's otherwise review-clean, is risk without payoff. Good candidate for the next functional change to this file - `TryReleaseLockWithRetries(...)` returning bool is the right shape when that happens.

src/DigitalPreservation/Preservation.API/Features/ImportJobs/Requests/GetDiffImportJob.cs
tomcrane: Decision (tomcrane): keeping the LINQ form; S3267 stays active. Resolving.

src/DigitalPreservation/Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs
tomcrane: Confirmed (tomcrane): nothing in CloudWatch currently queries structured log properties by name - message-text matching is unaffected by the casing change, so there is no dashboard/alert exposure. (Structured-property-based alerting is a future ambition, not a present dependency; when that work happens it will be built against the new PascalCase names.) Resolving.

brianmac1967 (approved):
LGTM

PR #209, opened 2026-08-06
Fix/lpii 135/failing clam av concurrent threads

brianmac1967:
Ticket: https://digirati.atlassian.net/browse/LPII-135

## Summary

Fixes a cluster of related Pipeline API reliability issues tracked under LPII-135. Individually these looked like separate incidents (ClamAV connection errors, tasks going "unhealthy" and getting replaced mid-run, jobs silently leaving deposits locked), but investigating them turned up several distinct, compounding root causes in how Pipeline API manages threads, memory, external processes, and S3/METS calls during a pipeline run. All were reproduced and fixed against a real large deposit (thousands of files) on dev.

## Changes

**ClamAV / concurrency**
- Reconfigured `clamd` (`MaxConnectionQueueLength`, `MaxThreads`) to handle concurrent scan connections without dropping requests under load.
- `GetVirusDefinition()` made properly async — the previous sync call blocked a ThreadPool thread waiting on `clamd`, starving the rest of the app (including health checks) whenever `clamd` was mid-scan.

**Process reliability**
- `clamd` and `freshclam` are now supervised and auto-restart if they crash. Previously a dead `clamd` didn't fail the scan — `clamscan-shim.sh` just logged "Connection refused" and reported a false clean result. Required a follow-up fix: both processes self-daemonize by default, which the first supervision pass mistook for a crash-loop; now run with `--foreground`.
- `PYTHONUTF8=1` — stops Brunnhilde hanging on non-ASCII filenames, which wedged every subsequent SQS job behind it.
- `SqsPipelineQueue`'s dequeue now has a client-side timeout instead of hanging indefinitely.
- ExifTool output is streamed straight to disk instead of buffered fully in memory — was contributing to OOM kills on large deposits.

**Memory leak**
- A `System.Timers.Timer` (`processTimer`) was never stopped on job completion. It kept the entire job's handler instance alive indefinitely and re-fired a background HTTP call every 10s per completed job — root cause of "every consecutive pipeline run provisions a new ECS task." Now stopped unconditionally in the handler's `finally` block.
- Tuned GC (`DOTNET_GCConserveMemory`, switched to Workstation GC) to release memory back to the OS between jobs instead of retaining it as a high-water mark. An initial max-aggressive setting (`GCConserveMemory=9`) traded the memory creep for a worse failure mode — multi-minute blocking GC pauses that froze the health endpoint — dialed back to a lighter setting that avoids both.

**METS correctness**
- `ParseExifToolOutput` hardened against malformed lines and false file-boundary matches.
- `significantProperties` conflict exceptions now include the file path, making stale-METS conflicts diagnosable instead of an opaque property-name-only error.

**Auth token refresh**
- Pipeline API's Bearer token to Preservation API was attached once, at `HttpClient` construction time, and never refreshed for the life of that client — so any job running long enough to cross the token's actual expiry had every subsequent API call fail with 401 partway through, cascading into failed lock releases and status updates. Replaced with a `DelegatingHandler` (`MachineAuthTokenInjector`) that fetches a current token per request, mirroring the pattern already used correctly for user-delegated auth.

**Deposit lock release**
- A transient DB error releasing a deposit's lock went unhandled and surfaced as a generic failure; the retry loop around it was a tight busy-loop with a broken timeout check (compared `TimeSpan.Seconds`, the 0-59 *component*, instead of `TotalSeconds`); and exhausting all retries still reported the job `Completed` — a deposit could be left locked indefinitely with nothing in the job's own status to show it. Now retries with backoff, times out correctly, and reports `CompletedWithErrors` with an explicit message on genuine failure.

**Redundant S3 API calls**
- Every metadata folder/file created after Brunnhilde ran triggered a fula-time S3 scan of the *entire* deposit (`GenerateDepositFileSystem`) torefresh a `WorkspaceManager` that was then thrown away — ~40s per call on a large deposit, repeated 20-25 times per job just for uploading a handful of report files. Fixed by
fetching the deposit metadata fresh (cheap, needed for METS ETag correctne-triggering the expensive full file-system walk, which`Storage.AddToDepositFileSystem` already keeps incrementally accurate. Confirmed on a real run: 25 scans → 3, upload-phase duration ~16 min → ~4 min.

## Test plan
- [x] `dotnet build` — clean
- [x] Full non-integration/non-manual test suite — 521/521 passing
- [x] Validated against a real large deposit on dev: two consecutive runssh-loop, no memory-driven task replacement, no thread-pool-starvation warnings
- [x] Confirmed the auth-token and lock-release fixes correctly surface as `CompletedWithErrors` instead of silently swallowing it
- [x] Confirmed the S3-refactor doesn't break METS write correctness (initial version did — see fixup commit `22c26e7` — caught via a real run before merge)
- [ ] Recommend monitoring `MemoryUtilized` and ECS task-replacement freqost-merge before promoting to test/prod

**PR description assisted by Claude**

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Workspace/MetadataReader.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Workspace/MetadataReader.cs
@@ -531,7 +531,12 @@ public class MetadataReader : IMetadataReader
 
             foreach (var str in exifResultList)
             {
-                if (str.Contains("========") || str.Contains("directories scanned"))
+                // ExifTool always emits the file-boundary marker as the first characters of the
+                // line ("======== <path>") - matching with StartsWith (rather than Contains)
+                // avoids treating an embedded metadata value that merely contains "========"
+                // (e.g. inside a raw ICC profile dump) as a spurious new-file boundary, which
+                // would otherwise split or merge real files' fields into the wrong file's block.
+                if (str.StartsWith("========") || str.Contains("directories scanned"))
                 {
                     if (i > 0)
                     {
@@ -561,7 +566,12 @@ public class MetadataReader : IMetadataReader
                     sbRawOutput.Append(str);
                     sbRawOutput.Append("\\n");
 
+                    // Not every line is a "Field : Value" pair - stray warnings or crashed-tool
+                    // text can end up mixed into the combined stdout for the whole batch. Skip
+                    // anything without a colon rather than throwing: one malformed line used to
+                    // abort parsing for every file in the deposit via the catch-all below.
                     var metadataPair = str.Split(":", 2);
+                    if (metadataPair.Length < 2) continue;
 
                     var rgx = new Regex("[^a-zA-Z0-9]");
                     var key = rgx.Replace(metadataPair[0].Trim(), "");
--- a/src/DigitalPreservation/DigitalPreservation.Workspace/MetadataReader.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Workspace/MetadataReader.cs
@@ -531,57 +531,20 @@ public class MetadataReader : IMetadataReader
 
             foreach (var str in exifResultList)
             {
-                // ExifTool always emits the file-boundary marker as the first characters of the
-                // line ("======== <path>") - matching with StartsWith (rather than Contains)
-                // avoids treating an embedded metadata value that merely contains "========"
-                // (e.g. inside a raw ICC profile dump) as a spurious new-file boundary, which
-                // would otherwise split or merge real files' fields into the wrong file's block.
-                if (str.StartsWith("========", StringComparison.Ordinal) || str.Contains("directories scanned"))
+                if (!IsFileBoundaryLine(str))
                 {
-                    if (i > 0)
-                    {
-                        exifModel.ExifMetadata = new List<ExifTag>(exifMetadataForFile);
-                        exifModel.ExifMetadata.Add(new ExifTag
-                        {
-                            TagName = "RawOutput",
-                            TagValue = sbRawOutput.ToString()
-                        });
-
-                        exifMetadataForFile.Clear();
-                        sbRawOutput.Clear();
-                        result.Add(exifModel);
-
-                        if (str.Contains("directories scanned"))
-                            break;
-
-                    }
-
-                    var fileName = str.Replace("========", string.Empty).Trim();
-                    exifModel = new ExifModel { Filepath = fileName };
-
-                    i++;
+                    AppendExifFieldLine(str, sbRawOutput, exifMetadataForFile);
+                    continue;
                 }
-                else
+
+                if (i > 0)
                 {
-                    sbRawOutput.Append(str);
-                    sbRawOutput.Append("\\n");
-
-                    // Not every line is a "Field : Value" pair - stray warnings or crashed-tool
-                    // text can end up mixed into the combined stdout for the whole batch. Skip
-                    // anything without a colon rather than throwing: one malformed line used to
-                    // abort parsing for every file in the deposit via the catch-all below.
-                    var metadataPair = str.Split(":", 2);
-                    if (metadataPair.Length < 2) continue;
-
-                    var rgx = new Regex("[^a-zA-Z0-9]");
-                    var key = rgx.Replace(metadataPair[0].Trim(), "");
-
-                    exifMetadataForFile.Add(new ExifTag
-                    {
-                        TagName = key,
-                        TagValue = metadataPair[1].Trim()
-                    });
+                    FinalizeCurrentFile(exifModel, exifMetadataForFile, sbRawOutput, result);
+                    if (str.Contains("directories scanned")) break;
                 }
+
+                exifModel = new ExifModel { Filepath = str.Replace("========", string.Empty).Trim() };
+                i++;
             }
 
             return result;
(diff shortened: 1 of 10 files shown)


src/DigitalPreservation/Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs
```diff
@@ -291,9 +318,16 @@ private async Task<ProcessPipelineResult> ExecuteBrunnhilde(ExecutePipelineJob r
             metadataPathForProcessFilesAndDirectories);
         logger.LogInformation("depositName {depositId}", request.DepositId);
 
+        // BagIt deposits get a SHA256 for every object file independently from bagit.py
+        // (BagitScript includes --sha256), so asking Siegfried to also hash every byte of every
+        // file via --hash is pure duplicated I/O with no gain in digest coverage - it only costs
+        // Brunnhilde's own duplicate-file report. For non-BagIt (RootLevel) deposits, Siegfried's
+        // hash is the only digest source recorded, so it can't be dropped there.
+        var hashArg = workspaceManager.IsBagItLayout ? string.Empty : "--hash sha256 ";
```
tomcrane: **Blocking.** Omitting `--hash` doesn't disable hashing — Brunnhilde defaults to md5 (`_determine_hash_type` in brunnhilde.py returns `"md5"` when the flag is absent, and `"none"` isn't an accepted value either). So Siegfried still reads every byte (no I/O saved) and writes an `md5` column to siegfried.csv. `SiegfriedCsvRowMap` maps the header name `sha256`, so CsvHelper throws `HeaderValidationException` — I confirmed this with a probe test against `SiegfriedOutput.FromCsvString`. `ParseSiegfriedOutput` swallows the exception and returns null, so BagIt deposits silently lose all Siegfried format metadata (PRONOM IDs, format names, MIME types) in METS while the job still reports success.

Suggest reverting this to unconditionally pass `--hash sha256` — the duplicated-digest cost is real but unavoidable via Brunnhilde's CLI.

src/DigitalPreservation/Pipeline.API/Features/Pipeline/SqsPipelineQueue.cs
```diff
@@ -45,15 +45,25 @@ public async ValueTask QueueRequest(string jobIdentifier, string depositName, st
...
 
             logger.LogInformation($"About to check queue {queue} for messages");
-            var queueUrlResponse = await sqsClient.GetQueueUrlAsync(queue, cancellationToken);
+
+            // WaitTimeSeconds below is an SQS-side long-poll duration, not a client-side timeout - if
+            // the underlying connection to SQS goes dead without erroring, these calls can hang
+            // indefinitely with no exception, silently wedging this single-threaded consumer loop
+            // forever (seen in practice - see LPII-135). Bound the whole dequeue attempt so a hang
+            // degrades to a retried failure on the next loop iteration instead of a permanent stall.
+            using var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
```
tomcrane, line 54: Minor: `linkedToken` also governs the `DeleteMessage` call below. If the 30s budget expires after a successful `ReceiveMessage` but before the delete completes, the catch-all logs the cancellation and still returns the populated `messageModel` — the job runs, but the undeleted message comes back after the visibility timeout and the job runs again. The window is narrow (GetQueueUrl + a ≤10s long poll leaves plenty of headroom), but it's easy to close: scope the timeout to `GetQueueUrlAsync`/`ReceiveMessageAsync` and pass the plain `cancellationToken` to `DeleteMessage`.

src/DigitalPreservation/DigitalPreservation.Workspace/MetadataReader.cs
```diff
@@ -531,7 +531,12 @@ public static List<ExifModel> ParseExifToolOutput(string exifResultStr)
 
             foreach (var str in exifResultList)
             {
-                if (str.Contains("========") || str.Contains("directories scanned"))
+                // ExifTool always emits the file-boundary marker as the first characters of the
+                // line ("======== <path>") - matching with StartsWith (rather than Contains)
+                // avoids treating an embedded metadata value that merely contains "========"
+                // (e.g. inside a raw ICC profile dump) as a spurious new-file boundary, which
+                // would otherwise split or merge real files' fields into the wrong file's block.
+                if (str.StartsWith("========") || str.Contains("directories scanned"))
```
tomcrane: Nit: the Contains→StartsWith change is right, but `string.StartsWith(string)` is culture-sensitive by default — prefer `str.StartsWith("========", StringComparison.Ordinal)` (Sonar will flag S4058 here eventually too).

src/DigitalPreservation/Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs
```diff
@@ -194,6 +195,32 @@ public async Task<Result> Handle(ExecutePipelineJob request, CancellationToken c
...
+            // handling API where forcing collections mid-stream would be harmful. That idle period
+            // turns out not to reliably trigger cleanup on its own - GC isn't time-based, it's
+            // allocation-pressure-based, and routine SQS polling/health checks generate far too
+            // little garbage to cross the threshold for a full collection. So the large object graphs
+            // built while processing this job (the deposit's CombinedDirectory tree, rebuilt many
+            // times over per LPII-135, plus Brunnhilde/exif output buffers) can sit uncollected
+            // indefinitely, and MemoryUtilized just reflects whatever's currently committed rather
+            // than what's actually still live. Force a full, compacting pass now instead of waiting
+            // for enough allocation pressure from a future job to trigger one naturally.
+            GCSettings.LargeObjectHeapCompactionMode = GCLargeObjectHeapCompactionMode.CompactOnce;
```
tomcrane: Nit: `GCCollectionMode.Aggressive` already implies a blocking, compacting collection including the LOH, so the `LargeObjectHeapCompactionMode.CompactOnce` line is redundant — harmless, but it reads as if it's doing something the next line doesn't already do.

tomcrane (changes requested):

tomcrane (approved):

PR #210, opened 2026-08-07
Fix NRE in PremisManager.Read for PRONOM-key-only formats

tomcrane:
## What

`PremisManager.Read` threw a `NullReferenceException` when reading PREMIS whose format element has a PRONOM registry key but no `FormatDesignation`.

## Why

`Patch` with a `PronomKey` but no `FormatName` (a legitimate sequence — see `Premis_Digest_and_Size_then_Edit_More`) creates a `FormatComplexType` containing a `FormatRegistry` only. `Read` then dereferenced `pronomFormat.FormatDesignation.FormatName.Value` unconditionally.

This was found during the MetsManager work in March and documented in a test comment rather than fixed (out of scope at the time); this PR is the proper fix.

## How

- `Read` now uses null-conditional access: `FormatDesignation?.FormatName?.Value`, leaving `FormatName` null when there is no designation.
- New regression test `Premis_Read_After_PronomKey_Only_Patch` — verified it fails with the NRE against the previous code and passes with the fix.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Mets/PremisManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/PremisManager.cs
@@ -39,7 +39,8 @@ public class PremisManager
 
         if (pronomFormat != null)
         {
-            premisFile.FormatName = pronomFormat.FormatDesignation.FormatName.Value;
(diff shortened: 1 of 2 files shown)


brianmac1967 (approved):
LGTM

Issue #188, opened 2026-04-29
Use of '/' in METS XML IDs


tomcrane:
**When we generate IDs for the various METS XML elements in MetsManager, we base them on file paths. This means they have `/` characters in them.** 

> Note - the discussion applies to the changes introduced in the branch `experimental/complex-mets-partial` which at time of writing is not merged to `main`. The changes in this branch make the problem a lot easier to see and (hopefully) solve.

EDIT - now merged to `main`.

XML `ID` attributes must conform to `xs:ID`, which inherits the `NCName` production rule from the XML Namespaces spec. `NCName` allows letters, digits, hyphens, underscores, and periods, but not `/`. METS 1.x uses `xs:ID` / `xs:IDREF` throughout, so IDs like `PHYS_metadata/ad-hoc` and cross-references that point to them are schema-invalid.

**Why it hasn't caused problems:** Most XML parsers (including .NET's XmlReader) don't validate ID uniqueness or NCName conformance by default - only a validating parser with a DTD or XSD does. Since the XmlGen serializer just writes strings and `MetsParser` just reads them back, both sides are equally lenient and everything works internally.

But:

  - Schema validation: Run the METS through a METS XSD validator and it will fail on every path-derived ID.
  - Interoperability: Any external system that consumes the METS with a validating parser (e.g., a digital preservation system, a repository ingest tool) will reject it.
  - XSLT `id()` function: If anyone ever processes the METS with XSLT using the id() lookup function, it won't find elements by these IDs.

Potential fix:

Replace `/` with `_` when constructing IDs (but not labels or OriginalName). So `PHYS_metadata/ad-hoc` becomes `PHYS_metadata_ad-hoc`. Since both the ID and its cross-references (ADMID, FILEID, etc.) are generated by the same code, we change the ID-building logic once and the METS stays internally consistent. The hyphen in ad-hoc is fine - `NCName` allows hyphens after the first character.

## Implications

The IDs are not opaque. Look at `LocateMetsDivByLocalPath`:

```c#
  var elements = localPath!.Split('/', StringSplitOptions.RemoveEmptyEntries);
  // ...
  testPath += element;   // rebuilt with / between segments
  // This is navigating using our ID convention for directories
  // If we don't want to do this, we can use the premis:originalName for the directory
  var childDiv = div.Div.SingleOrDefault(d => d.Id == $"{Constants.PhysIdPrefix}{testPath}");
```

[...]

PR #211, opened 2026-08-11
Decouple METS navigation from ID format via path cache (#188 step 1)

tomcrane:
## My (Tom) explanation:

This PR builds a new cache when working with METS to allow finding the mets XML element for the physical file, given its relative path on disk.

e.g., for the file `objects/images/hello.jpg`, where is its `mets:div`? And from there you can find `mets:file` in the `fileSec`, and traverse the XML via `ADMID` attributes to the metadata for that file, and so on..

We used to answer this question by basing the file XML element's `ID` on the file path, e.g., `objects/images/hello.jpg` => `PHYS_objects/images/hello.jpg`

But this introduces two problems - 
1. `/` in an XML element ID is not valid according to the strict schema (all XML parsers don't mind it but a correct METS parser does)
2. What about the file `objects/docs/my nice word document.doc`? It has spaces in it. 

So we end up with `<mets:div ID="PHYS_objects/docs/my nice word document.doc" ADMID="ADM_objects/docs/my nice word document.doc" ... >`

... but :collision:, because this _actually_ means:

"I point to FOUR separate ADM xml elements and their IDs are `ADM_objects/docs/my`, `nice`, `word`, and `document.doc`"

For files, we use the `FLocat@xlink:href` on the `mets:file` element (div → fptr → FILE → FLocat).

For folders, we will now use `premis:originalName` in ADM to record the actual relative path on disk of the file, and traverse back to the file mets:div element that points at it; this builds a map of original path on disk to file divs, which is what we need.

This puts some constraints on what a valid _editable_ METS looks like, for the future work. It is not needed just to _read_ Goobi METS.

## Claude's explanation

Step 1 of the fix for #188 (METS XML IDs like `PHYS_objects/my file.pdf` are not valid `xs:ID`/NCName values). This PR does **not** change any IDs — it removes the last dependency on their format, so that step 2 (minting NCName-safe IDs via `XmlConvert.EncodeLocalName`) can ship without breaking navigation or backward compatibility. Full plan: [issue-188-plan.md](https://github.com/digirati-co-uk/digital-preservation/blob/main/docs/issues/188/issue-188-plan.md) (v2), [issue-188-reassessment.md](https://github.com/digirati-co-uk/digital-preservation/blob/main/docs/issues/188/issue-188-reassessment.md) and the [IDREFS companion plan](https://github.com/digirati-co-uk/digital-preservation/blob/main/docs/issues/188/issue-188-idrefs-plan.md) (in `docs/issues/188/` on main, via PR #212).

**Zero behavior change to produced XML.** No existing test assertion was altered; the full XmlGen.Tests suite (312 tests) passes unchanged.

## How it works

- **`FullMets.PhysicalDivsByPath`** — a cache mapping each deposit-relative path to its div in the PHYSICAL structMap, built by the new **`MetsCache`** from the div's own metadata: `premis:originalName` for directories, the referenced FILE's `FLocat` href for files. Never from the div's ID.
- Populated when either `IMetsStorage` implementation loads a METS, lazily on first navigation for a directly-constructed `FullMets`, and maintained by `AddNewFile` / `AddNewDirectory` / `DeleteDiv`. A `[Conditional("DEBUG")]` assertion verifies the maintained cache equals a fresh rebuild on every navigation, so every existing MetsManager test exercises cache consistency.
- **`LocateMetsDivByLocalPath` navigates in three tiers**: cache lookup → resolving the current div's children by their own metadata (defeats duplicate-path collisions) → the legacy `PHYS_ + path` ID convention. Navigation is therefore a strict superset of what main does today; the legacy tier is the step-2 backward-compat mechanism and is removable after a step-3 bulk migration.
- **Diagnosability**: `MetsCache.Populate` returns (and keeps on `FullMets.PathDiagnostics`) one entry per div whose path could not be resolved or that collided — appended to navigation error messages, and the seed of a future "is this METS editable?" conformance check (relevant to editing third-party METS).
- Path normalisation (`./`, BagIt `data/` prefix, trailing `/`) is applied identically to cache keys, incoming edit paths, minted IDs and `FLocat` comparisons.

## Also in this PR

- **ClamAV digiprovMD lookup fix** (`MetsParser`): the fallback was a case-insensitive *substring* match, which cross-talks between files whose IDs are prefixes of each other — under step 2's encoded IDs that becomes a real hazard (`ADM_a` ⊂ `ADM_a_x002F_b`). Now an exact case-insensitive match, with regression tests. Note for the docs repo: `02c-METS-Parsing.md` says digiprov IDs are matched "by containment" — needs updating when this merges.
- **Hardening from a high-effort review pass** (second commit): tolerate zero/multiple PHYSICAL structMaps at load (diagnostic, not exception); adds attach nothing to the METS until the fallible metadata step succeeds, and refuse to add a div whose ID already exists (no duplicate-ID corruption); `Set*ByPath` methods no longer write metadata to an ancestor div when a path only partially resolves.

## Test coverage

14 new cache/navigation tests + 2 ClamAV regressions, all deliberately **ID-agnostic** (they assert on paths, labels and reference identity, never on ID attribute values) so they survive step 2 unchanged. Coverage includes: population on load, add/update/delete maintenance verified against fresh rebuilds, lazy population, normalisation variants (`data/`, `./`, `./data/`, trailing `/`), duplicate-path impostor resolution, unresolvable-div diagnostics, failed-add atomicity, and editing the frozen legacy spaces fixture (`Samples/path-fixture-spaces.xml`) by path.

## Sequencing

The [IDREFS companion](https://github.com/digirati-co-uk/digital-preservation/blob/main/docs/issues/188/issue-188-idrefs-plan.md) is already implemented on `feat/188-idrefs-resolution`, stacked on this branch; its PR opens once this merges. Step 2 (`feat/188-encoded-mets-ids`) then branches from that and must not merge first. This PR deploys standalone with no coordination requirements — it produces byte-identical METS.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
@@ -100,7 +100,9 @@ public class MetsManager(
 
     private Result EditMets(WorkingBase? workingBase, string? deletePath, FullMets fullMets)
     {
-        var localPath = FolderNames.RemovePathPrefix(workingBase?.LocalPath ?? deletePath)!;
+        // Normalise once so ID minting, FLocat writing and cache keys all use the same
+        // canonical deposit-relative form of the path.
+        var localPath = MetsCache.NormalisePathKey(workingBase?.LocalPath ?? deletePath) ?? string.Empty;
         var (contextDiv, parent, foundDepth, totalDepth) = LocateMetsDivByLocalPath(fullMets, localPath);
 
         if (foundDepth == totalDepth)
@@ -127,8 +129,12 @@ public class MetsManager(
             return Result.Fail(ErrorCodes.BadRequest, "No working directory or working file supplied to add.");
         }
 
-        return Result.Fail(ErrorCodes.BadRequest,
-            $"Could not edit METS because not all parts of the path '{localPath}' have been added to METS.");
+        var message = $"Could not edit METS because not all parts of the path '{localPath}' have been added to METS.";
+        if (fullMets.PathDiagnostics.Count > 0)
+        {
+            message += " METS path diagnostics: " + string.Join("; ", fullMets.PathDiagnostics);
+        }
+        return Result.Fail(ErrorCodes.BadRequest, message);
     }
 
     private Result UpdateExistingFile(DivType contextDiv, FullMets fullMets, WorkingFile workingFile, string localPath)
@@ -137,7 +143,7 @@ public class MetsManager(
             return Result.Fail(ErrorCodes.BadRequest, "WorkingFile path does not end on a file");
 
         var (file, _) = SetFileAndFileGroup(contextDiv, fullMets);
-        if (file.FLocat[0].Href != localPath)
+        if (MetsCache.NormalisePathKey(file.FLocat[0].Href) != localPath)
             return Result.Fail(ErrorCodes.BadRequest, "WorkingFile path doesn't match METS flocat");
 
         PopulateDmdFromResource(fullMets, workingFile, contextDiv);
@@ -161,6 +167,15 @@ public class MetsManager(
         var physId = Constants.PhysIdPrefix + localPath;
         var fileId = Constants.FileIdPrefix + localPath;
 
+        // Reaching an add for a path whose div already exists means that div could not be
+        // resolved by path OR by the legacy ID fallback - adding again would write duplicate
+        // ID attributes into the METS. Refuse rather than corrupt.
+        if (parentDiv.Div.Any(d => d.Id == physId))
+        {
+            return Result.Fail(ErrorCodes.BadRequest,
+                $"METS already contains a div '{physId}' that could not be resolved to path '{localPath}'.");
+        }
+
         var childItemDiv = new DivType
         {
             Type = Constants.ItemType,
@@ -168,14 +183,15 @@ public class MetsManager(
             Id = physId,
             Fptr = { new DivTypeFptr { Fileid = fileId } }
         };
-        parentDiv.Div.Add(childItemDiv);
 
-        PopulateDmdFromResource(fullMets, workingFile, childItemDiv);
+        // Nothing is attached to the METS until the fallible metadata step has succeeded,
+        // so a failed add leaves the document (and the cache) exactly as it was.
         var metadataResult = metadataManager.ProcessAllFileMetadata(fullMets, childItemDiv, workingFile, localPath, true);
         if (metadataResult.Failure)
             return metadataResult;
 
-        // After ProcessAllFileMetadata, so the FILE/FLocat entry the cache mirrors exists
+        PopulateDmdFromResource(fullMets, workingFile, childItemDiv);
+        parentDiv.Div.Add(childItemDiv);
         fullMets.PhysicalDivsByPath[localPath] = childItemDiv;
 
         SortChildDivs(parentDiv);
@@ -187,6 +203,12 @@ public class MetsManager(
         var physId = Constants.PhysIdPrefix + localPath;
         var admId = Constants.AdmIdPrefix + localPath;
         var techId = Constants.TechIdPrefix + localPath;
+
+        if (parentDiv.Div.Any(d => d.Id == physId))
+        {
+            return Result.Fail(ErrorCodes.BadRequest,
+                $"METS already contains a div '{physId}' that could not be resolved to path '{localPath}'.");
+        }
 
         var childDirectoryDiv = new DivType
         {
@@ -328,7 +350,7 @@ public class MetsManager(
         {
             var (file, fileGroup) = SetFileAndFileGroup(div, fullMets);
 
-            if (file.FLocat[0].Href != operationPath)
+            if (MetsCache.NormalisePathKey(file.FLocat[0].Href) != operationPath)
             {
                 return Result.Fail(ErrorCodes.BadRequest, "Delete path doesn't match METS flocat");
             }
@@ -372,9 +394,10 @@ public class MetsManager(
         EnsureCache(fullMets);
         AssertCacheConsistent(fullMets);
 
+        localPath = MetsCache.NormalisePathKey(localPath) ?? string.Empty;
         var elements = localPath.Split('/', StringSplitOptions.RemoveEmptyEntries);
 
-        var div = fullMets.Mets.StructMap.Single(sm => sm.Type == Constants.Physical).Div!;
+        var div = fullMets.Mets.StructMap.First(sm => sm.Type == Constants.Physical).Div!;
         DivType? parent = null;
         var testPath = string.Empty;
         var counter = 0;
@@ -388,17 +411,17 @@ public class MetsManager(
             testPath += element;
 
             // Navigate by path (premis:originalName / FLocat href, via the cache), not by
-            // reconstructing div IDs from the path - IDs are opaque (issue #188).
-            if (!fullMets.PhysicalDivsByPath.TryGetValue(testPath, out var childDiv))
+            // reconstructing div IDs from the path - IDs are opaque (issue #188). The cached
+            // div is only accepted if it really is a child of the current div; a malformed
+            // source can have two unrelated divs resolving to the same path, and the cache
+            // holds whichever was walked first.
+            if (!fullMets.PhysicalDivsByPath.TryGetValue(testPath, out var childDiv) ||
+                !div.Div.Contains(childDiv))
             {
-                break;
+                childDiv = FindChildDivByPath(div, testPath, fullMets.Mets);
             }
 
-            // Guard against a malformed source in which two unrelated divs resolve to the same
-            // path: only accept the cached div if it really is a child of the current div. The
-            // walk then breaks with foundDepth < totalDepth, producing the same "not all parts
-            // of the path have been added" error as a plain miss.
-            if (!div.Div.Contains(childDiv))
+            if (childDiv is null)
             {
                 break;
             }
@@ -409,6 +432,19 @@ public class MetsManager(
         }
 
         return (div, parent, counter, elements.Length);
+    }
+
+    /// <summary>
+    /// Fallback resolution among the current div's children when the cache has no usable entry
+    /// for a path: first by each child's own path metadata (correct even when an unrelated div
+    /// elsewhere claims the same path), then by the legacy ID convention, which keeps divs with
+    /// broken path metadata navigable exactly as they were before the cache existed. The legacy
+    /// tier can be removed after a bulk ID migration (issue #188 step 3).
+    /// </summary>
+    private static DivType? FindChildDivByPath(DivType parent, string testPath, DigitalPreservation.XmlGen.Mets.Mets mets)
+    {
+        var byMetadata = parent.Div.FirstOrDefault(d => MetsCache.TryResolvePath(d, mets) == testPath);
+        return byMetadata ?? parent.Div.FirstOrDefault(d => d.Id == $"{Constants.PhysIdPrefix}{testPath}");
     }
 
     /// <summary>
@@ -522,7 +558,9 @@ public class MetsManager(
 
     public void SetRecordInfoByPath(FullMets mets, string localPath, RecordInfo recordInfo)
     {
-        var (div, _, _, _) = LocateMetsDivByLocalPath(mets, localPath);
+        var (div, _, foundDepth, totalDepth) = LocateMetsDivByLocalPath(mets, localPath);
+        // A partial walk means div is an ANCESTOR of the requested path - never write to it
+        if (foundDepth != totalDepth) return;
         SetRecordInfoForDiv(mets, div, recordInfo);
     }
 
@@ -543,7 +581,8 @@ public class MetsManager(
 
     public void SetRightsStatementByPath(FullMets mets, string localPath, Uri? rightsStatement)
     {
-        var (div, _, _, _) = LocateMetsDivByLocalPath(mets, localPath);
+        var (div, _, foundDepth, totalDepth) = LocateMetsDivByLocalPath(mets, localPath);
+        if (foundDepth != totalDepth) return;
         SetRightsStatementForDiv(mets, div, rightsStatement);
     }
 
@@ -559,7 +598,8 @@ public class MetsManager(
     // which removes the element and allows parent rights to flow through.
     public void SuppressRightsInheritanceByPath(FullMets mets, string localPath)
     {
-        var (div, _, _, _) = LocateMetsDivByLocalPath(mets, localPath);
+        var (div, _, foundDepth, totalDepth) = LocateMetsDivByLocalPath(mets, localPath);
+        if (foundDepth != totalDepth) return;
         SuppressRightsInheritanceForDiv(mets, div);
     }
 
@@ -594,7 +634,8 @@ public class MetsManager(
 
     public void SetAccessRestrictionsByPath(FullMets mets, string localPath, List<string> accessRestrictions)
     {
-        var (div, _, _, _) = LocateMetsDivByLocalPath(mets, localPath);
+        var (div, _, foundDepth, totalDepth) = LocateMetsDivByLocalPath(mets, localPath);
+        if (foundDepth != totalDepth) return;
         SetAccessRestrictionsForDiv(mets, div, accessRestrictions);
     }
 
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
@@ -391,23 +391,31 @@ public class MetsManager(
         }
 
         parent!.Div.Remove(div);
-
-        // Only evict the cache entry if it is actually THIS div's - when the deleted div was
-        // reached via a fallback tier (its own path metadata being broken), the key may map to
-        // a different div that legitimately owns the path.
-        if (fullMets.PhysicalDivsByPath.TryGetValue(operationPath!, out var cachedDiv) &&
-            ReferenceEquals(cachedDiv, div))
-        {
-            fullMets.PhysicalDivsByPath.Remove(operationPath!);
-            if (fullMets.PathDiagnostics.Count > 0)
-            {
-                // A malformed doc may contain another div that claimed the same path (see the
-                // duplicate-path diagnostics) - rebuild so the cache reflects post-delete reality.
-                MetsCache.Populate(fullMets);
-            }
-        }
+        EvictFromPathCache(fullMets, div, operationPath!);
 
         return Result.Ok();
+    }
+
+    /// <summary>
+    /// Only evict the cache entry if it is actually the deleted div's - when the deleted div
+    /// was reached via a fallback tier (its own path metadata being broken), the key may map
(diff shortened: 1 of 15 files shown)


src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -362,9 +391,13 @@ private static (FileType file, MetsTypeFileSecFileGrp fileGroup) SetFileAndFileG
...
     private static (DivType contextDiv, DivType? parent, int foundDepth, int totalDepth) LocateMetsDivByLocalPath(FullMets fullMets, string localPath)
     {
+        EnsureCache(fullMets);
+        AssertCacheConsistent(fullMets);
+
+        localPath = MetsCache.NormalisePathKey(localPath) ?? string.Empty;
         var elements = localPath.Split('/', StringSplitOptions.RemoveEmptyEntries);
 
-        var div = fullMets.Mets.StructMap.Single(sm => sm.Type == Constants.Physical).Div!;
+        var div = fullMets.Mets.StructMap.First(sm => sm.Type == Constants.Physical).Div!;
```
brianmac1967: **Unhandled crash when METS has zero PHYSICAL structMaps**

`LocateMetsDivByLocalPath` still throws an unhandled `InvalidOperationException` when the METS has zero PHYSICAL structMaps, contradicting the PR's stated goal of tolerating this case as a diagnostic rather than an exception.

`MetsCache.Populate`/`Build` already tolerates 0 PHYSICAL structMaps (returns a diagnostic, leaves the cache empty, no throw). But any subsequent edit (`AddToMets`/`DeleteFromMets`/`SetModsInformation`) on such a `FullMets` calls `LocateMetsDivByLocalPath`, which does `fullMets.Mets.StructMap.First(sm => sm.Type == Constants.Physical).Div!` unconditionally — `.First()` throws 'Sequence contains no matching element' for a malformed/legacy METS with no PHYSICAL structMap, crashing the request instead of returning the intended graceful `Result.Fail` with diagnostics.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -466,7 +558,9 @@ private static void PopulateDmdFromResource(FullMets mets, ResourceBase resource
 
     public void SetRecordInfoByPath(FullMets mets, string localPath, RecordInfo recordInfo)
     {
-        var (div, _, _, _) = LocateMetsDivByLocalPath(mets, localPath);
+        var (div, _, foundDepth, totalDepth) = LocateMetsDivByLocalPath(mets, localPath);
```
brianmac1967, line 672: **Silent no-op reports success when path resolution fails**

`SetRecordInfoByPath`/`SetRightsStatementByPath`/`SuppressRightsInheritanceByPath`/`SetAccessRestrictionsByPath` are void methods that now silently no-op when `foundDepth != totalDepth`, but callers (e.g. `SetModsInformationHandler`) have no way to detect this and still return `Result.Ok()`.

A client calls `SetModsInformation` for a localPath that doesn't fully resolve in METS (e.g. stale path, race with a pending add, or normalisation mismatch). Each `Set*ByPath` call silently returns without writing anything, but `SetModsInformationHandler` (`DigitalPreservation.Workspace/Requests/SetModsInformation.cs:39-58`) never inspects any failure signal from these void calls, writes the (unmodified) METS, and returns `Result.Ok()` — the API reports success while none of the requested access restrictions/rights/record info were actually applied.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -392,6 +434,56 @@ private static (DivType contextDiv, DivType? parent, int foundDepth, int totalDe
...
     }
 
+    /// <summary>
+    /// Fallback resolution among the current div's children when the cache has no usable entry
+    /// for a path: first by each child's own path metadata (correct even when an unrelated div
+    /// elsewhere claims the same path), then by the legacy ID convention, which keeps divs with
+    /// broken path metadata navigable exactly as they were before the cache existed. The legacy
+    /// tier can be removed after a bulk ID migration (issue #188 step 3).
+    /// </summary>
+    private static DivType? FindChildDivByPath(DivType parent, string testPath, DigitalPreservation.XmlGen.Mets.Mets mets)
```
brianmac1967, line 515: **FirstOrDefault silently masks duplicate-ID METS corruption**

`FindChildDivByPath`'s legacy-ID fallback tier uses `FirstOrDefault` instead of the `SingleOrDefault` it replaced, silently picking one div when two children share the same conventional 'PHYS_+path' ID instead of surfacing the corruption.

Two direct-child divs under the same parent both have broken/missing path metadata (so cache and per-child metadata resolution both fail) but happen to share the reconstructed ID `PHYS_{testPath}` (duplicate/corrupted METS). Previously `div.Div.SingleOrDefault(d => d.Id == ...)` would throw, surfacing the corruption loudly; now `parent.Div.FirstOrDefault(d => d.Id == ...)` silently returns whichever div is first in the list, and an edit can land on the wrong div without any error or diagnostic.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -184,6 +204,12 @@ private Result AddNewDirectory(DivType parentDiv, FullMets fullMets, WorkingDire
         var admId = Constants.AdmIdPrefix + localPath;
         var techId = Constants.TechIdPrefix + localPath;
 
+        if (parentDiv.Div.Any(d => d.Id == physId))
+        {
+            return Result.Fail(ErrorCodes.BadRequest,
+                $"METS already contains a div '{physId}' that could not be resolved to path '{localPath}'.");
+        }
+
         var childDirectoryDiv = new DivType
```
brianmac1967, line 227: **AddNewDirectory can desync tree/cache on mid-method exception**

`AddNewDirectory` attaches the new div to the METS tree and adds its AmdSec before the fallible metadata/mods steps run, unlike `AddNewFile` which was explicitly hardened to attach nothing until the fallible step succeeds.

If `metadataManager.GetAmdSecType(...)` or `PopulateDmdFromResource(...)` throws partway through `AddNewDirectory` (e.g. an unexpected exception during premis/mods construction), `childDirectoryDiv` has already been added to `parentDiv.Div` and its AmdSec already added to `fullMets.Mets.AmdSec`, but `fullMets.PhysicalDivsByPath[localPath] = childDirectoryDiv;` (the last line) never runs — leaving the METS tree and the path cache out of sync. In DEBUG this trips `AssertCacheConsistent`'s `InvalidOperationException` on the next navigation; in Release it silently degrades to slower per-child metadata/ID fallback resolution for that directory.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -348,6 +376,7 @@ private static Result DeleteDiv(DivType div, FullMets fullMets, DivType? parent,
         }
 
         parent!.Div.Remove(div);
+        fullMets.PhysicalDivsByPath.Remove(operationPath!);
```
brianmac1967: **DeleteDiv can evict an unrelated div's cache entry**

`DeleteDiv` removes the cache entry keyed by `operationPath` without verifying (for directories) that `operationPath` actually is the deleted div's own resolved path, unlike the equivalent FLocat check that exists for files.

In a malformed METS where two divs resolve to the same path (the 'impostor' scenario this PR's own tests construct for Add), deleting a directory div reached via the legacy-ID or per-child-metadata fallback (not the cache) still does `fullMets.PhysicalDivsByPath.Remove(operationPath!)` unconditionally — if that key currently maps to a different, unrelated div (the winner of the earlier path collision), the delete removes that unrelated div's valid cache entry, forcing it back onto the slower fallback path with no diagnostic raised.

brianmac1967 (commented):
Automated review findings (checked against docs/issues/188/issue-188-plan.md as well as the diff). 5 inline findings below.

One additional finding didn't fit as an inline comment since it's about a file the PR doesn't touch: `Storage.Repository.Common/Mets/MetsFromArchivalGroup.cs` (`AddResourceToMets`/`AddBinariesToMets`) builds the physical structMap directly and is never updated to populate or maintain `FullMets.PhysicalDivsByPath`, even though the Step 1 design doc (§1.2/§1.4/§1.7) explicitly lists this file as in-scope for cache population/maintenance. Not an active runtime bug today — the `FullMets` these methods build is written to disk immediately and reloaded fresh via `IMetsStorage.GetFullMets`, which does call `MetsCache.Populate` — but it means the debug-build cache-consistency assertion this PR relies on for coverage never exercises this construction path, so a future divergence here (e.g. a duplicate-path bug) would go undetected by the same safety net that covers every `MetsManager` mutation.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
tomcrane: Fixed in 498c900. `LocateMetsDivByLocalPath` now uses `FirstOrDefault` and signals a missing PHYSICAL structMap with a `foundDepth: -1` sentinel; `EditMets` catches it and returns a clean `Result.Fail` carrying the load-time path diagnostics ("METS has no PHYSICAL structMap"). New test: `Editing_A_Mets_With_No_Physical_StructMap_Fails_Cleanly`.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
tomcrane: Fixed in 498c900 — agreed this was the worst consequence of the earlier ancestor-write fix. All eight `Set*ByPath`/`Set*ByDivId` methods now return `Result` (NotFound, with the path diagnostics attached for the ByPath variants); `IMetsManager` is updated and `SetModsInformationHandler` propagates the first failure instead of writing the unmodified METS and reporting success. The ByDivId variants also lose their null-forgiving `!`, which would have NRE'd on an unknown divId. Test updated to assert the failure as well as the absence of ancestor writes.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
tomcrane: Fixed in 498c900. Both fallback tiers now require a UNIQUE match among the parent's children (`UniqueOrNull`); ambiguity returns null, so the walk breaks and the caller gets the standard incomplete-path error with the load-time diagnostics attached. Chose clean failure over restoring main's `SingleOrDefault` throw — same "surface, don't guess" outcome without an unhandled exception. New test: `Two_Children_Sharing_A_Conventional_Id_Fail_Cleanly_Instead_Of_Guessing`.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
tomcrane: Fixed in 498c900 — `AddNewDirectory` now builds the div and amdSec, runs `PopulateDmdFromResource`, and only then attaches div + amdSec + cache entry together, matching `AddNewFile`'s attach-nothing-until-the-fallible-steps-succeed principle.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
tomcrane: Fixed in 498c900. `DeleteDiv` only evicts the cache entry when it reference-equals the deleted div, so a colliding div's valid entry survives. One step further than the finding: when the deleted div WAS the cache owner and the document carries duplicate-path diagnostics, the cache is rebuilt — otherwise the previously-losing div would now be cacheable and the debug consistency assertion would (correctly) flag the maintained cache as stale.

tomcrane:
Re the review-body finding on `MetsFromArchivalGroup` (not updated for cache population, contra plan §1.2/§1.4/§1.7): adopted in spirit in 498c900. Live cache *maintenance* genuinely doesn't apply there — no navigation ever happens on the `FullMets` it builds (write-only; the next reader loads fresh through `IMetsStorage.GetFullMets`, which populates) — which is why the implementation deviated from the plan doc. But the coverage point stands: that construction path sat outside the debug safety net. `CreateStandardMets` now populates the cache on its write-only `FullMets` and asserts (debug builds/tests) that the structMap it built is fully navigable by path, so a future divergence there fails Storage.API tests instead of surfacing as diagnostics on some later load.

tomcrane:
Post-review hardening summary (for re-review), beyond the five inline-comment fixes in 498c900:

- **dffdfeb** — the zero-PHYSICAL-structMap crash the review caught in `LocateMetsDivByLocalPath` had an unfixed sibling in `LocateMetsDivByDivId`; fixed the same way. Also replaced a `Debug.Assert` with a `[Conditional("DEBUG")]` throw (a failed `Debug.Assert` can kill the whole test process in .NET test hosts).
- **4a7b9e1** — Sonar S3776 on `DeleteDiv` (16 → under 15) via `EvictFromPathCache` extraction.
- **991563d** — a second high-effort adversarial review over the fix commits themselves found 10 more defects, all fixed: the four `Set*ForDiv` methods now fail (instead of silently succeeding) when a div's dmdSec can't be materialised as MODS; `LinkFile`/`SetFileLinks` normalise paths before minting `FILE_` ids; navigation ambiguity is genuinely terminal (no legacy-ID fallthrough, and strict per-child resolution whenever load diagnostics exist); **`MetsFromArchivalGroup` no longer duplicates the template `metadata` div/amdSec for AGs with a preserved `metadata/` folder — a pre-existing latent bug surfaced by the new debug assert**; `DeleteDiv` tolerates empty/dangling ADMID/DMDID and evicts cache entries by value; `LocateMetsDivByDivId` searches all PHYSICAL structMaps and skips empty ones; error codes harmonised (unresolvable path = BadRequest everywhere, unknown divId = NotFound). Seven new regression tests.

Suite: 323 XmlGen tests green, no pre-existing assertions changed.

brianmac1967 (commented):
Re-review after the latest 6 commits. All 6 findings from the previous review pass are confirmed fixed — verified against the actual code (not just commit messages) and cross-referenced with the new PhysicalDivsCacheTests covering each scenario (e.g. Two_Siblings_Resolving_The_Same_Path_Make_Edits_Of_That_Path_Fail, Deleting_A_Div_Cached_Under_A_Different_Path_Leaves_No_Dangling_Entry). Full suite passes (324/324, 0 failed).

Two new, minor findings below, both isolated to the newly-touched MetsFromArchivalGroup.cs.

src/DigitalPreservation/Storage.Repository.Common/Mets/MetsFromArchivalGroup.cs
```diff
@@ -26,17 +26,41 @@ public class MetsFromArchivalGroup(IMetsManager metsManager, IMetsParser metsPar
...
         AddResourceToMets(mets, archivalGroup.Id!, mets.StructMap[0].Div, archivalGroup);
-        
-        var writeResult = await metsManager.WriteMets(new FullMets{ Mets = mets, Uri = file });
+
+        var fullMets = new FullMets { Mets = mets, Uri = file };
+        // This FullMets is write-only (navigation happens on a fresh load, which populates the
+        // cache), but building the cache here puts this construction path under the same
+        // navigability check as MetsManager mutations: a structMap this class builds that
+        // cannot be resolved by path is a bug, caught in debug builds and tests.
+        var pathDiagnostics = MetsCache.Populate(fullMets);
```
brianmac1967: **Unconditional cache build is wasted work in Release**

`MetsCache.Populate(fullMets)` runs unconditionally in `CreateStandardMets`, but its only consumer (`AssertNavigable`) is `[Conditional("DEBUG")]` and compiled out in Release, so the cache walk is pure wasted work in production.

In a Release build, every `CreateStandardMets` call walks the entire structMap (`Walk`/`ResolvePath` in `MetsCache.cs`, including `XmlElement.GetElementsByTagName` lookups per directory div and FileGrp/File scans per item div) to build `fullMets.PhysicalDivsByPath`, which is never read again (`WriteMets` only serializes `mets.Mets`; `GetMetsFileWrapper` re-loads by URI). For an archival group with many files/folders this adds real, unbounded CPU/time to production METS creation for zero observable benefit — the population call itself should also be gated (e.g. wrapped in `#if DEBUG` or moved inside a Conditional helper) rather than only its consumer.

src/DigitalPreservation/Storage.Repository.Common/Mets/MetsFromArchivalGroup.cs
```diff
@@ -49,16 +73,16 @@ private void AddResourceToMets(DigitalPreservation.XmlGen.Mets.Mets mets, Uri ar
...
-                // The objects div should already exist from our template
-                childDirectoryDiv = mets.StructMap[0].Div.Div.Single(d => d.Id == Constants.ObjectsDivId);
-            }
+            var localPath = childContainer.Id!.LocalPath.RemoveStart(agLocalPath).RemoveStart("/");
+
+            // The template already contains divs (and amdSecs) for objects/, metadata/ and
+            // metadata/ad-hoc/ - reuse any div the parent already has for this path rather
+            // than adding a duplicate div and duplicate-ID amdSec (which previously happened
+            // for any Archival Group with a preserved metadata/ folder).
+            var childDirectoryDiv = div.Div.FirstOrDefault(d => d.Id == $"{Constants.PhysIdPrefix}{localPath}");
```
brianmac1967: **Div reuse rebuilds the ID-from-path pattern being retired**

The new duplicate-div fix reuses existing template divs by reconstructing the conventional ID from the path (`PHYS_<path>`), the same ID-from-path pattern issue #188 is eliminating elsewhere in this PR.

This is explicitly out of scope for step 1 per the PR description (this class only needs to produce readable METS today), but it means this new reuse-matching logic will need to be rewritten again in step 2 when IDs become NCName-encoded and no longer derivable from the path via simple string concatenation — a special case built on the very mechanism being retired, rather than using the new path-cache (premis:originalName / FLocat) resolution this PR introduces everywhere else.

src/DigitalPreservation/Storage.Repository.Common/Mets/MetsFromArchivalGroup.cs
tomcrane: Fixed in 20cd1ad. `MetsCache.Populate` has moved inside the `[Conditional("DEBUG")]` `AssertNavigable` method (which now takes the `FullMets`), so in Release the entire call — argument evaluation included — compiles out and no cache walk happens. Debug builds and tests keep the navigability check unchanged.

src/DigitalPreservation/Storage.Repository.Common/Mets/MetsFromArchivalGroup.cs
tomcrane: Fixed in 20cd1ad. The reuse match is now by resolved path — `div.Div.FirstOrDefault(d => MetsCache.TryResolvePath(d, mets) == localPath)` — using the same premis:originalName resolution the cache uses, instead of reconstructing the conventional ID. `TryResolvePath` is made public to support this. New divs this class *mints* still use the `PHYS_+path` convention, which is intentional for step 1: minting changes centrally in step 2, and this matching logic now won't need to change with it.

brianmac1967 (approved):

PR #213, opened 2026-08-12
feat: IDREFS-aware ID resolution on both METS stacks (#188 companion)

tomcrane:
Companion to #211 (issue #188). Implements [docs/issues/188/issue-188-idrefs-plan.md](https://github.com/digirati-co-uk/digital-preservation/blob/main/docs/issues/188/issue-188-idrefs-plan.md).

## The problem

`ADMID`/`DMDID` are schema-typed **IDREFS** — whitespace-separated lists of ID references. Our two METS-reading stacks fail in mirror-image ways:

- **XmlGen typed model**: XmlSerializer splits the attribute into tokens, so a legacy space-containing ID (`ADM_objects/my file.pdf`) arrives broken. The existing `string.Join(' ', …)` workaround reconstructs it — but that same join would *break* a genuine multi-reference (`ADMID="ADM_A ADM_B"` matches no element). Several sites also used `Single()`/`[0]`, which throw on no-match or empty collections.
- **XDocument (`MetsParser`)**: raw whole-string lookups mean legacy space IDs work naturally, but a genuine IDREFS list finds nothing — no split fallback existed.

## The fix

**`IdRefs` helper (new, XmlGen side)** — tiered resolution: try the joined form first (a legacy space-containing single ID; unambiguous, because no schema-valid ID can contain a space), then each token individually. Never throws; null means unresolvable. Applied at all six join sites (`MetadataManager`, `ModsManager` ×2, `MetsManager.DeleteDiv` ×2, `MetsCache`).

**String overload for the parser** — whole-value lookup first, then split into tokens. Applied at the five `MetsParser` IDREFS lookup sites.

**Derived keys come from the resolved element, not the reference** — `ctx.FileAdmId` and the parser's ClamAV `digiprovMD_` key are now built from the resolved amdSec/techMD's actual `ID`, so they always embed a real ID even in the multi-IDREFS case.

**Crash fixes along the way** — `DeleteDiv` no longer throws `IndexOutOfRange` on an empty `Admid` collection or `InvalidOperationException` on a dangling reference; deletion is deliberately *tolerant* of unresolvable references (a broken reference is exactly what a delete may be cleaning up, and DMDIDs dangle by design due to lazy dmdSec creation).

## Step 2 interplay

After step 2 (`XmlConvert.EncodeLocalName` minting), no new ID can contain a space, so every new reference tokenises to exactly one member and the joined tier degenerates to `tokens[0]`. The tiering is purely a bridge for legacy and third-party content — removable from the XmlGen side only after a step 3 migration, never removable from the parser (third-party METS can carry real IDREFS lists).

## Tests

19 new tests: `IdRefsTests` (13 unit tests over the tiering), `IdRefsMetsManagerTests` (4, including genuine two-token ADMID/DMDID fixtures and clean-failure DeleteDiv cases), `IdRefsParsingTests` (2, including ClamAV key derivation from the resolved ID). Legacy `path-fixture-spaces.xml` coverage stays green.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
@@ -45,6 +45,27 @@ public static class IdRefs
     }
 
     /// <summary>
+    /// Mirror-image resolution for the XDocument/raw-string side (MetsParser), where the whole
+    /// IDREFS attribute value arrives as one string: try the whole value first (a single ID,
+    /// or a legacy ID containing spaces), then each whitespace-separated token (schema-valid
+    /// IDREFS list - XML allows tab/newline/CR separators as well as spaces), first match
+    /// winning.
+    /// </summary>
+    public static T? ResolveSingle<T>(string attributeValue, Func<string, T?> lookupById)
+        where T : class
+    {
+        var whole = lookupById(attributeValue);
+        if (whole != null || attributeValue.IndexOfAny(XmlWhitespace) < 0)
+        {
+            return whole;
+        }
+        return attributeValue
+            .Split(XmlWhitespace, StringSplitOptions.RemoveEmptyEntries)
+            .Select(lookupById)
+            .FirstOrDefault(match => match != null);
+    }
+
+    /// <summary>
     /// Resolve an IDREFS token collection to every referenced element. The same tiering as
     /// <see cref="ResolveSingle{T}(IReadOnlyList{string},Func{string,T})"/>: a joined-form
     /// match means the tokens are ONE legacy space-containing ID, so exactly that element is
@@ -97,26 +118,5 @@ public static class IdRefs
         {
             tokens.Remove(resolvedId);
         }
-    }
-
-    /// <summary>
-    /// Mirror-image resolution for the XDocument/raw-string side (MetsParser), where the whole
-    /// IDREFS attribute value arrives as one string: try the whole value first (a single ID,
-    /// or a legacy ID containing spaces), then each whitespace-separated token (schema-valid
-    /// IDREFS list - XML allows tab/newline/CR separators as well as spaces), first match
-    /// winning.
-    /// </summary>
-    public static T? ResolveSingle<T>(string attributeValue, Func<string, T?> lookupById)
-        where T : class
-    {
-        var whole = lookupById(attributeValue);
-        if (whole != null || attributeValue.IndexOfAny(XmlWhitespace) < 0)
-        {
-            return whole;
-        }
-        return attributeValue
-            .Split(XmlWhitespace, StringSplitOptions.RemoveEmptyEntries)
-            .Select(lookupById)
-            .FirstOrDefault(match => match != null);
     }
 }
--- a/src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
@@ -49,7 +49,11 @@ public static class IdRefs
     /// IDREFS attribute value arrives as one string: try the whole value first (a single ID,
     /// or a legacy ID containing spaces), then each whitespace-separated token (schema-valid
     /// IDREFS list - XML allows tab/newline/CR separators as well as spaces), first match
-    /// winning.
+    /// winning. Deliberately NOT implemented by delegating to the token-collection overload:
+    /// this side's first tier must match the attribute value VERBATIM (a legacy ID could
+    /// contain any whitespace a filename can), whereas the collection overload's joined tier
+    /// reconstructs with single spaces because that is all the XmlSerializer's split leaves
+    /// it - the two first tiers are subtly different and each is right for its input.
     /// </summary>
     public static T? ResolveSingle<T>(string attributeValue, Func<string, T?> lookupById)
         where T : class
@@ -94,14 +98,6 @@ public static class IdRefs
                     .ToList();
         }
     }
-
-    /// <summary>
-    /// True when the token collection references the element with this ID - as its only token,
-    /// as one token of a genuine IDREFS list, or as the joined form of a legacy
-    /// space-containing ID. Pure string comparison; no lookup involved.
-    /// </summary>
-    public static bool References(IReadOnlyList<string> tokens, string id) =>
-        tokens.Contains(id) || (tokens.Count > 1 && Joined(tokens) == id);
 
     /// <summary>
     /// Remove from an IDREFS token collection the reference to an element previously resolved
--- a/src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
@@ -17,6 +17,12 @@ namespace DigitalPreservation.Mets;
 /// </summary>
 public static class IdRefs
 {
+    /// <summary>
+    /// The characters XML treats as whitespace, and therefore as IDREFS separators. The
+    /// XmlSerializer splits on all of them; the raw-string side must do the same.
+    /// </summary>
+    private static readonly char[] XmlWhitespace = [' ', '\t', '\n', '\r'];
+
     /// <summary>The single intended ID that a legacy space-containing ID was split from.</summary>
     public static string Joined(IEnumerable<string> tokens) => string.Join(' ', tokens);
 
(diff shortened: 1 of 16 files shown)


brianmac1967 (commented):
Automated review: 4 correctness findings, all variations on the same root cause — multi-token IDREFS handling that only resolves/matches the first token, silently orphaning the rest.

One finding couldn't be anchored inline because it's in a method this PR doesn't touch, so noting it here instead:

**`MetsManager.cs` — `RemoveLogicalStructMapDmdSecs` (around line 886)**: this method was not updated to use the new `IdRefs` helper and still does an exact per-token dmdId match (`mets.Mets.DmdSec.FirstOrDefault(d => d.Id == dmdId)` per raw token in `foreach (var dmdId in div.Dmdid)`), so it misses the same legacy-space-ID case this whole PR exists to fix. A logical structMap div carrying a legacy DMDID that `XmlSerializer` split into multiple tokens (e.g. `"DMD_objects/my"` + `"file.pdf"`, reconstructing to `"DMD_objects/my file.pdf"`) will never match either token against the real space-containing dmdSec ID, so the dmdSec is silently left behind (orphaned) when the structMap is removed — even though the PR's stated goal was to apply IdRefs resolution at every ADMID/DMDID join site.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -364,58 +363,50 @@ private static Result DeleteDiv(DivType div, FullMets fullMets, DivType? parent,
...
+            admTokens = div.Admid;
         }
 
-        RemoveById(fullMets.Mets.AmdSec, admId, a => a.Id);
-        RemoveById(fullMets.Mets.DmdSec, JoinedIdOrNull(div.Dmdid), d => d.Id);
+        // ADMID/DMDID are IDREFS token collections (see IdRefs). Deletion is tolerant of a
+        // section that doesn't resolve - a dangling reference is exactly the kind of breakage
+        // a delete may be cleaning up, and DMDID references dangle by design until metadata
+        // is first set (lazy dmdSec creation).
+        var amdSec = IdRefs.ResolveSingle(admTokens, id => fullMets.Mets.AmdSec.FirstOrDefault(a => a.Id == id));
```
brianmac1967: **Correctness:** `DeleteDiv` resolves and removes only the first matching amdSec/dmdSec via `IdRefs.ResolveSingle`, leaking any other genuinely-referenced sections when ADMID/DMDID is a real multi-token IDREFS list.

**Failure scenario:** A div/file whose ADMID (or DMDID) genuinely lists two distinct real amdSec/dmdSec IDs (e.g. from ingested Archivematica/Goobi METS, as this PR's own MetsCache test proves reachable) is deleted via `DeleteFromMets`: `IdRefs.ResolveSingle` returns only the first-matching amdSec/dmdSec, `AmdSec.Remove`/`DmdSec.Remove` remove that one element, and the div/file entry is removed from the structMap — but the second referenced amdSec/dmdSec is never removed, permanently orphaned in the METS with no test covering this path.

src/DigitalPreservation/DigitalPreservation.Mets/ModsManager.cs
```diff
@@ -168,14 +168,17 @@ public static void SetRecordInfo(this ModsDefinition modsDefinition, RecordInfo
...
+        // lazy creation is the joined legacy form only when no dmdSec resolves (see IdRefs).
+        var dmd = IdRefs.ResolveSingle(div.Dmdid, id => mets.DmdSec.FirstOrDefault(x => x.Id == id));
+        return GetModsForDmdId(mets, dmd?.Id ?? IdRefs.Joined(div.Dmdid), createDmd);
     }
 
     
     public static void SetModsForDiv(DigitalPreservation.XmlGen.Mets.Mets mets, DivType div, ModsDefinition mods)
     {
-        var normalised = string.Join(' ', div.Dmdid);
+        var dmd = IdRefs.ResolveSingle(div.Dmdid, id => mets.DmdSec.FirstOrDefault(x => x.Id == id));
```
brianmac1967, line 190: **Correctness:** `SetModsForDiv` resolves only the first matching dmdSec, but `TryRemoveRedundantDmd`'s `div.Dmdid.Clear()` drops all DMDID token references unconditionally, orphaning any additional genuinely-referenced dmdSec.

**Failure scenario:** A div has a genuine two-token DMDID pointing at two real, MODS-only dmdSecs; clearing its MODS content calls `SetModsForDiv` → `TryRemoveRedundantDmd`, which resolves and removes only the first dmdSec (this line's `IdRefs.ResolveSingle` picks one), then clears the entire `div.Dmdid` collection further down. The second dmdSec is left in `mets.DmdSec` with no div referencing it any more — a silent, permanent orphan that also can't be reached to be cleaned up later.

src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// Mirror-image resolution for the XDocument/raw-string side (MetsParser), where the whole
+    /// IDREFS attribute value arrives as one string: try the whole value first (a single ID,
+    /// or a legacy ID containing spaces), then each whitespace-separated token (schema-valid
+    /// IDREFS list), first match winning.
+    /// </summary>
+    public static T? ResolveSingle<T>(string attributeValue, Func<string, T?> lookupById)
+        where T : class
+    {
+        var whole = lookupById(attributeValue);
+        if (whole != null || !attributeValue.Contains(' '))
```
brianmac1967: **Correctness:** This string-attribute overload of `ResolveSingle` only splits on a literal space character, not the full XML whitespace set (tab/newline/CR), so an IDREFS list separated by those characters silently fails to resolve.

**Failure scenario:** A third-party METS (or hand-edited legacy file) has `ADMID="ADM_missing\tADM_real"` (tab-separated, valid per XML's whitespace-collapse rule for IDREFS). `ResolveSingle` tries the whole string (fails), then checks `attributeValue.Contains(' ')` — false, since only a tab is present — and returns the failed whole-value lookup (null) without ever attempting to split and resolve `ADM_real`, even though a genuinely valid reference exists.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
tomcrane: Fixed in a06ee9e, with one refinement over "remove them all": blindly removing every resolved section would break the *shared*-section case — e.g. Archivematica commonly references one shared rightsMD from many files' ADMID lists, and deleting one file must not take the shared section with it. So `DeleteDiv` now uses a new tiered `IdRefs.ResolveAll` (joined legacy form → exactly that one section; otherwise every resolving token) and removes each resolved amdSec/dmdSec **unless `SectionReferencedElsewhere`** — a scan of every div (all structMaps) and file, using `IdRefs.References` (single token, one-of-list, or joined legacy form) with the deletee excluded.

Covered by `Deleting_A_File_With_Genuine_Multi_Admid_Removes_Its_Sections_But_Keeps_Shared_Ones`: a file with ADMID = [own techMD, own extra section, shared section] loses the first two on delete while the shared section survives (still referenced by the other file).

src/DigitalPreservation/DigitalPreservation.Mets/ModsManager.cs
tomcrane: Fixed in a06ee9e. `TryRemoveRedundantDmd` now takes the *resolved* dmdSec rather than re-looking-up by string, and after pruning it calls the new `IdRefs.RemoveReference(div.Dmdid, dmd.Id)` — which clears all tokens only when they jointly formed the legacy space-containing ID, and otherwise removes just the token equal to the pruned dmdSec's ID. A genuine multi-token DMDID keeps its other references and their dmdSecs. The full `Dmdid.Clear()` now happens only when *nothing* resolved at all (every token dangling — clearing is cleanup, and a no-op when DMDID was never set, preserving the original intent).

Covered by `Clearing_Mods_On_A_Div_With_Genuine_Multi_Dmdid_Keeps_The_Other_DmdSec`: after clearing, the div's own dmdSec is gone, the second dmdSec survives, and `div.Dmdid` still references it.

src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
tomcrane: Fixed in a06ee9e: the string overload now uses a shared `XmlWhitespace` set (space, tab, LF, CR) for both the "does this need splitting?" check (`IndexOfAny`) and the split itself — matching the separator set the XmlSerializer uses on the typed-model side. Covered by a `[Theory]` over tab, LF and CRLF separators (`Idrefs_attribute_value_splits_on_any_xml_whitespace`). The whole-value tier is unchanged: legacy platform-minted IDs only ever contained plain spaces (they come from file paths), so the joined/whole-string forms stay space-based.

tomcrane:
All four findings addressed in a06ee9e — thanks, the un-anchored one was the most interesting because `RemoveLogicalStructMapDmdSecs` was indeed missing from the site inventory in `issue-188-idrefs-plan.md`.

That method now resolves each logical div's DMDID through the new `IdRefs.ResolveAll` (so a legacy space-containing dmdSec ID split across tokens is found via the joined form, and a genuine multi-token list removes every resolving dmdSec), guarded by `SectionReferencedElsewhere` so a dmdSec still referenced from outside the structMap being removed survives. Covered by `Replacing_A_Logical_StructMap_Removes_A_Legacy_Spaced_DmdSec`.

The same guard applies in `DeleteDiv` — see the inline replies for why plain remove-all would have traded the orphan bug for a shared-section deletion bug. A post-fix sweep of every `.Admid`/`.Dmdid` use in `DigitalPreservation.Mets` confirms the remaining raw uses are token-count checks, minting, and the deliberate `IdRefs.Joined` fallbacks — no other exact-match lookup sites remain. The plan doc's status header and its "add `ResolveAll` only when a caller needs it" note are updated to record this round.

brianmac1967 (approved):
LGTM

PR #214, opened 2026-08-13
feat: mint schema-valid METS IDs (#188 step 2)

tomcrane:
Step 2 of #188. Every METS `xs:ID` the platform mints is now schema-valid.

`PHYS_objects/my file.pdf` — which is neither a valid NCName nor a single IDREFS token — becomes `PHYS_objects_x002F_my_x0020_file.pdf`, via a single `MetsIdEncoding.ToMetsId()` (`XmlConvert.EncodeLocalName`) that every minting site goes through. Paths themselves are untouched: FLocat `href` and `premis:originalName` keep their spaces and slashes, and remain what navigation actually uses (the step 1 path cache).

**Backwards compatibility is preserved and tested.** Existing raw IDs are never rewritten, legacy documents stay navigable and editable, and anything we write *into* one references the IDs that document actually holds.

## Beyond the plan's inventory

Four things needed more than the planned find-and-replace:

- **`BuildFptr`/`LinkFile`/`UnLinkFile`/`SetFileLinks` now RESOLVE the FILE element's real ID** from the fileSec (`ResolveFileId`) instead of minting one. Minting would have written dangling FILEID/smLink references into legacy documents, and made their existing raw-ID smLinks unremovable (removal matches on the same derivation). This is finding **F3** from the whole-feature review; the pin test added for it back then passes unchanged.
- **The ID-convention navigation tier had to learn both forms.** It reconstructed only `PHYS_` + raw path, so a div minted with an encoded ID that later lost its path metadata became unreachable — the exact failure that tier exists to prevent. Found by an existing test, not a new one.
- **The duplicate-div guard needed widening.** An encoded mint no longer collides with a legacy raw-ID div for the same path, so without checking both forms an add could quietly create a *second* div for one path.
- **dmdSec IDs now come from the div's path, not its ID.** Setting metadata on a legacy div therefore mints a valid ID rather than a second invalid one. The `ByDivId` entry points recover the path from the cache.

## New validation at the client boundary

`SetStructMap` is the only route by which a caller-supplied string becomes an `xs:ID`. It now validates the whole range tree and returns `BadRequest` **before writing anything** (so a bad replacement can't destroy the structMap already there). IDs are rejected, not encoded — the client round-trips them, so silently changing one would break its own references. The UI only ever mints `LOG_<epoch-ms>`, which passes.

`IMetsManager.SetStructMap` returns `Result` as a result; `SetLogicalStructMapHandler` propagates it. A range with no ID is now written *without* an ID attribute rather than as `ID=""` (also invalid) — this shape is real, it is what the parser reports for third-party logical divs, and replace/remove still match it.

## Tests

`MetsIdIntegrityTests` is the merge gate: every ID valid and unique, every ADMID/FILEID/smLink reference resolving, across empty, awkward-path, logical-structMap, Archival-Group and **mixed legacy** documents. Nothing in the suite checked this before — `XmlSerializer` does not enforce `xs:ID`, which is how #188 survived from the beginning.

It stands in for the planned XSD validation, which would need the METS + xlink schema set committed for offline CI and would currently fail on an unrelated pre-existing issue: the template writes DMDID onto the objects/metadata/ad-hoc divs while their dmdSecs are only created lazily, so those references dangle by design. **Dangling-DMDID and full XSD validation are both recorded as conformance-checker backlog**, not fixed here.

Expectations are built through `MetsIds`/`ToMetsId`, never as encoded literals, so a change to the encoding shows up as one deliberate edit rather than a wall of strings. The frozen legacy corpus in `Samples/` is untouched.

XmlGen 444 passed (Debug **and** Release), Storage.API 55, Preservation.API 106, Core 68, Common.Model 8.

## ⚠️ Release engineering

`DigitalPreservation.Mets` and `Storage.Repository.Common` are compiled into **Preservation.API, Pipeline.API, DigitalPreservation.UI, Storage.API, Storage.API.Importer and Deposit.Archiver**. A mixed-version window means one service minting raw IDs while another mints encoded ones into the same METS. **This must ship as a single release tagging all of them.**

Step 3 (bulk migration of existing documents) stays deferred. Until it happens the legacy tiers — the ID-convention fallback, the IdRefs rejoining, and the raw-form checks added here — all stay.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
@@ -512,15 +512,15 @@ public class MetsManager(
 
         // fptrs live on divs in every structMap - the physical div being deleted takes its own
         // with it, but a LOGICAL div painting this file keeps pointing at a file that is gone.
-        foreach (var div in AllDivs(mets))
-        {
-            var staleFptrs = div.Fptr
+        // Materialise before mutating - the divs cannot be walked while their fptrs are removed.
+        var stalePointers = AllDivs(mets)
+            .SelectMany(div => div.Fptr
                 .Where(fptr => fptr.Fileid == fileId || fptr.Area?.Fileid == fileId)
-                .ToList();
-            foreach (var fptr in staleFptrs)
-            {
-                div.Fptr.Remove(fptr);
-            }
+                .Select(fptr => (div, fptr)))
+            .ToList();
+        foreach (var (div, fptr) in stalePointers)
+        {
+            div.Fptr.Remove(fptr);
         }
     }
 
@@ -572,21 +572,11 @@ public class MetsManager(
     /// legal XML and does occur in malformed documents - dereferencing it unguarded is a real
     /// crash, so this is the single place that tolerance lives.
     /// </summary>
-    private static IEnumerable<DivType> AllDivs(DigitalPreservation.XmlGen.Mets.Mets mets)
-    {
-        foreach (var structMap in mets.StructMap)
-        {
-            var root = structMap.Div;
-            if (root is null)
-            {
-                continue;
-            }
-            foreach (var div in SelfAndDescendants(root))
-            {
-                yield return div;
-            }
-        }
-    }
+    private static IEnumerable<DivType> AllDivs(DigitalPreservation.XmlGen.Mets.Mets mets) =>
+        mets.StructMap
+            .Select(structMap => structMap.Div)
+            .Where(root => root is not null)
+            .SelectMany(SelfAndDescendants);
 
     private static IEnumerable<DivType> SelfAndDescendants(DivType div)
     {
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
@@ -118,7 +118,7 @@ public class MetsManager(
             if (workingBase is WorkingFile workingFile)
                 return UpdateExistingFile(contextDiv, fullMets, workingFile, localPath);
             if (workingBase is WorkingDirectory workingDirectory)
-                return UpdateExistingDirectory(fullMets, workingDirectory, contextDiv);
+                return UpdateExistingDirectory(fullMets, workingDirectory, contextDiv, localPath);
             return Result.Fail(ErrorCodes.BadRequest, "WorkingBase is unsupported type");
         }
 
@@ -157,12 +157,12 @@ public class MetsManager(
         if (MetsCache.NormalisePathKey(file.FLocat[0].Href) != localPath)
             return Result.Fail(ErrorCodes.BadRequest, "WorkingFile path doesn't match METS flocat");
 
-        var dmdResult = PopulateDmdFromResource(fullMets, workingFile, contextDiv);
+        var dmdResult = PopulateDmdFromResource(fullMets, workingFile, contextDiv, localPath);
         if (dmdResult.Failure) return dmdResult;
         return metadataManager.ProcessAllFileMetadata(fullMets, contextDiv, workingFile, localPath);
     }
 
-    private static Result UpdateExistingDirectory(FullMets fullMets, WorkingDirectory workingDirectory, DivType contextDiv)
+    private static Result UpdateExistingDirectory(FullMets fullMets, WorkingDirectory workingDirectory, DivType contextDiv, string localPath)
     {
         if (contextDiv.Type != Constants.DirectoryType)
             return Result.Fail(ErrorCodes.BadRequest, "WorkingDirectory path does not end on a directory");
@@ -170,21 +170,19 @@ public class MetsManager(
         if (workingDirectory.Name.HasText())
             contextDiv.Label = workingDirectory.Name;
 
-        return PopulateDmdFromResource(fullMets, workingDirectory, contextDiv);
+        return PopulateDmdFromResource(fullMets, workingDirectory, contextDiv, localPath);
     }
 
     private Result AddNewFile(DivType parentDiv, FullMets fullMets, WorkingFile workingFile, string localPath)
     {
-        var physId = Constants.PhysIdPrefix + localPath;
-        var fileId = Constants.FileIdPrefix + localPath;
+        var physId = Constants.PhysIdPrefix + localPath.ToMetsId();
+        var fileId = Constants.FileIdPrefix + localPath.ToMetsId();
 
-        // Reaching an add for a path whose div already exists means that div could not be
-        // resolved by path OR by the legacy ID fallback - adding again would write duplicate
-        // ID attributes into the METS. Refuse rather than corrupt.
-        if (parentDiv.Div.Any(d => d.Id == physId))
+        var conflict = FindConflictingChild(parentDiv, localPath, physId);
+        if (conflict != null)
         {
             return Result.Fail(ErrorCodes.BadRequest,
-                $"METS already contains a div '{physId}' that could not be resolved to path '{localPath}'.");
+                $"METS already contains a div '{conflict}' that could not be resolved to path '{localPath}'.");
         }
 
         var childItemDiv = new DivType
@@ -201,7 +199,7 @@ public class MetsManager(
         if (metadataResult.Failure)
             return metadataResult;
 
-        var dmdResult = PopulateDmdFromResource(fullMets, workingFile, childItemDiv);
+        var dmdResult = PopulateDmdFromResource(fullMets, workingFile, childItemDiv, localPath);
         if (dmdResult.Failure) return dmdResult;
 
         parentDiv.Div.Add(childItemDiv);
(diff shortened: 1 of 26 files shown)


tomcrane:
## Four extra commits: fixes from a review of the whole #188 change set

Before asking for review I ran an adversarial review over the **cumulative** identifier work — #211, #213 and this PR together, `2228267..HEAD` — rather than just the diff here. Six independent passes over separate dimensions: the path cache and navigation, `IdRefs`, encoding and mixed-format documents, deletion and metadata-section survival, the parser and downstream consumers, and an audit of the test suite itself. Most findings were reproduced with throwaway tests before being believed.

It found more than the earlier review rounds did. Ten items belonged to this work and are fixed in `d787070`..`ad9f628`; the rest are pre-existing and are **not** touched here.

### The performance one is the most important

**Navigation had become quadratic on the file-add path**, and worse, cache population was quadratic *on every load*: `MetsCache` resolved each div with a linear scan of the whole fileSec. There's now a single ID index per walk. Separately, the add path fell through to resolving every sibling, because the last element of an add always misses the cache — on a document with no diagnostics the cache is a complete map, so that tier is provably empty and is skipped.

Adding 1600 files, in memory, Release: **14,017 ms → 705 ms** (8.76 ms → 0.44 ms per file). Extrapolated to a 20,000-file deposit the old behaviour was roughly 15–20 minutes of navigation CPU for one bulk `AddItemsToMets` sync.

Worth flagging for review: my first attempt at this went further and treated a cache miss as fully authoritative, which also dropped the ID-convention fallback tier. An **existing** test (`Deleting_A_Div_Cached_Under_A_Different_Path_Leaves_No_Dangling_Entry`) failed immediately — that tier answers a different question, and dropping it would have narrowed backwards compatibility for a speed win. Only the expensive tier is provably skippable. The residual super-linearity is `SortChildDivs`, which predates #188 and is deliberately left alone: turning its re-sort into an insert-in-position would stop it sorting a previously unsorted legacy document.

### The rest

- **The ClamAV `digiprovMD` ID was still minted from the resolved amdSec ID**, which on a pre-#188 document is the raw form — so every pipeline run over a legacy deposit minted a *new* ID containing a slash and a space, from the code this PR was supposed to have made schema-valid. It now comes from the encoded path. For documents written after step 2 the value is unchanged.
- **`DeleteDiv` left `fptr/@FILEID` and `smLink` references to the file it removed** — both IDREFs. This PR made it unrecoverable: a file re-added at the same path is minted an encoded ID, so the stale reference can never be matched again and `UnLinkFile` cannot remove it. Delete now sweeps both.
- **The duplicate-div guard compared IDs only.** When two siblings already resolve to one path, navigation refuses to guess and the operation arrives as an *add*; divs carrying a third-party ID scheme slipped past and produced a third div for the same path. It now also rejects on the resolved path.
- **`SetStructMap` validated NCName-ness but not uniqueness**, and `xs:ID` means both. Two ranges sharing an ID, or one colliding with an existing div ID, wrote a document no schema-aware consumer can load. The UI mints `LOG_<epoch-ms>`, so two ranges in the same millisecond collided.
- **`ResolveFileId` flattened every fileGrp**, so a derivative sharing the master's href (THUMBS, ALTO) could win a link. `OBJECTS` is searched first, matching `SetFileAndFileGroup`.
- **Every logical div with no ID derived the dmdSec ID `DMD_`** — the bare prefix — so they shared one section and each save destroyed the previous div's metadata. dmdSec IDs are ours to choose, unlike div IDs, so an unused one is minted. Pre-existing, but this PR blessed the ID-less shape without noticing it.

### Three test defects, two of them mine

- Replacing the **entire** encoding scheme failed only 2 of 444 tests, and neither was in `MetsIdEncodingTests` — because expectations are computed through the production encoder. That catches a mint site that forgets to encode, but not a change to the encoding itself, which is a permanent on-disk format a future step 3 migration will be written against. Added literal anchors, and removed a tautology.
- I had weakened `MetsManagerWithPremis`'s "no duplicate amdSec" assertion to an ID-equality check while doing the encoding sweep. That cannot see a duplicate minted under a different ID scheme, which is the regression it exists to catch. It now matches on the techMD's `premis:originalName`.
- The merge gate's structLink check iterated a collection that is empty when nothing is written, so a no-op `LinkFile` passed it.

**Every new regression test was verified by reverting its fix and confirming the test fails** — all seven confirmed. Given that the audit's central finding was tests that could not fail, asserting the fixes without that check seemed unwise.

### Deliberately not in this PR

- **~20 pre-existing findings.** Several are severe (a virus scan overwriting an unrelated PREMIS event; IDREFS resolution taking the first token that resolves *anything*, which can drop a file's SHA256 on Archivematica-shaped input; `AddNewFile`'s atomicity comment being untrue). None is caused by #188, so they belong in their own tickets rather than growing this one further. Happy to raise them.
- **One CI question for the team**: the gating `test-dotnet` job runs `--configuration Release`, where `[Conditional("DEBUG")]` assertions compile out. `MetsFromArchivalGroup.AssertNavigable` turns out to be load-bearing — breaking navigability in the AG export fails 8 tests in Debug but only 1 in Release. That's a one-line workflow change, but it's a change to how the project is tested rather than to this feature, so I've left it.

### Unchanged

`XmlGen` 460 passing in Debug **and** Release, Storage.API 55, Preservation.API 106, Core 68, Common.Model 8. SonarCloud reports no issues on new code. The frozen legacy corpus in `Samples/` is untouched.

The release constraint from the description still stands: the METS libraries compile into six deployables, and a mixed-version window would have one service minting raw IDs while another mints encoded ones into the same document — this has to ship as a single release.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -718,32 +833,45 @@ private static void AssertCacheConsistent(FullMets fullMets)
...
         return Result.Ok();
     }
 
+    /// <summary>
+    /// The deposit-relative path a div stands for, taken from the path cache. Null for a logical
+    /// div (only physical divs are cached) and for a physical div whose path metadata could not
+    /// be resolved. Used by the ByDivId entry points, which are given an ID rather than a path,
+    /// so that a lazily created dmdSec is still identified by path (issue #188).
+    /// </summary>
+    private static string? PathForDiv(FullMets mets, DivType div)
```
brianmac1967, line 895: `PathForDiv` does a linear O(n) reverse scan of the path cache (`FirstOrDefault` with `ReferenceEquals`) to recover a div's path — reintroducing, in a new code path, the exact "resolve by scanning the whole document" pattern R1 just eliminated from the add path.

Not reachable today: `SetRecordInfoByDivId`/`SetRightsStatementByDivId`/`SuppressRightsInheritanceByDivId`/`SetAccessRestrictionsByDivId` all call this, but none is currently wired to a controller. The `IMetsManager` doc comment does describe the `ByDivId` shape as intended for callers to loop over many div IDs against one `FullMets`, so the moment such a bulk caller exists, editing N divs in one deposit becomes O(N × cache size) — same bug class as R1, just dormant. Worth a note for whoever wires up the first `ByDivId` caller, rather than a blocker now.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -592,27 +695,39 @@ private static (DivType contextDiv, DivType? parent, int foundDepth, int totalDe
...
+                .Where(d => MetsCache.TryResolvePath(d, mets, index) == testPath)
+                .Take(2)
+                .ToList();
+            if (byMetadata.Count > 0)
+            {
+                return byMetadata.Count == 1 ? byMetadata[0] : null;
+            }
         }
-        return UniqueOrNull(parent.Div.Where(d => d.Id == $"{Constants.PhysIdPrefix}{testPath}"));
+        var encodedId = $"{Constants.PhysIdPrefix}{testPath.ToMetsId()}";
```
brianmac1967: The "encoded-ID-or-legacy-raw-ID" pair is reconstructed independently here and in `FindConflictingChild` (lines 272–273) instead of through one shared helper.

Low risk today, but if the ID-minting scheme changes at the Step 3 migration the plan doc anticipates, a fix applied to one site and missed in the other would silently reintroduce the mismatch R4 was written to catch — the duplicate-div guard and navigation disagreeing about which ID forms are valid. Might be worth a small `internal static (string encoded, string legacy) CandidateIds(string prefix, string path)` helper next time either site is touched.

brianmac1967 (commented):
Follow-up review after the R1–R7/T1–T3 fixes: verified all nine are genuinely and correctly fixed (traced the logic, ran the full suite in Release — 464 tests pass including the 94 new ones pinning these fixes). Two small, non-blocking findings below; neither is a live bug today since nothing currently calls the affected paths in a way that triggers them.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
tomcrane: Agreed, and the shape is worth recording even though nothing triggers it yet — but I'd draw the line slightly differently from "same bug class as R1", because the fix here is not the same fix.

R1 was O(siblings × files) **per operation on the hottest path in the system**, with every file upload paying it. This is O(cache) per call on a path with no callers at all. The order is bad for a bulk caller, as you say; the magnitude and the urgency aren't comparable.

More importantly, the obvious optimisation is a trap, so I've documented it rather than left it for whoever gets there first to discover. The reverse scan isn't only recovering a path — **the miss is what identifies the div as logical**. Swapping it for `MetsCache.TryResolvePath` looks like the same answer for less work, but a logical div of `TYPE="Item"` carrying an `fptr` resolves the path of the file it paints, so it would be handed the `DMD_` id of the *physical* div for that path and the two would silently share a dmdSec. That is R7 again by another route.

There also isn't a cheaper per-call answer: resolving a Directory div scans the amdSecs and an Item div scans the fileSec, so every option is O(n) for one call. The real fix belongs to the bulk caller — hoist a reverse map or a physical-div set across the loop — which is where the `IMetsManager` doc comment points.

`d33ac91` adds that reasoning to `PathForDiv`, including the cost and where the index should be hoisted, so the constraint is visible at the point someone would change it. Happy to raise a ticket against the first `ByDivId` caller as well if you'd rather it were tracked outside the code.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
tomcrane: Fixed in `d33ac91` rather than deferred to "next time either site is touched" — the failure mode you describe is precisely that the next person touches *one* of them, so leaving two copies until then is leaving the bug armed.

There is now a single `PhysicalDivIdCandidates(localPath)` returning the `(Encoded, Legacy)` pair, used by both `FindChildDivByPath` and `FindConflictingChild`. I dropped the `prefix` parameter from your suggested signature: only `PHYS_` divs are ever looked up by ID convention, and a general helper would invite callers to reconstruct `FILE_`/`ADM_` ids by convention too, which is the habit #188 is trying to end. `FindConflictingChild` also no longer takes the minted `physId` — it gets both forms from the helper, so there is one expression rather than a parameter that has to be kept in step with it.

The xmldoc states the invariant explicitly: the two sites must agree on the set, because if they disagree a div one can find is invisible to the other — which is the duplicate-div bug the guard exists to catch.

brianmac1967 (approved):

PR #218, opened 2026-08-13
fix: the three defects #188 introduced (#215, #216, #217)

### LINKED ISSUE #215: IDREFS resolution picks the first token that resolves anything, not what the caller needs
Found by the cumulative adversarial review of the #188 work (steps 1 + IDREFS + 2). Introduced by `8ea3b9a` (PR #213), so this is ours rather than inherited. Nothing is in production yet.

## What happens

`IdRefs.ResolveSingle`/`ResolveAll` resolve an IDREFS attribute by taking **the first token that resolves to anything**, without checking that what it found is usable by the caller. IDREFS order is not significant, but every call site behaves as if it were.

The shape that breaks it is an ADMID listing a non-object section first — the shared-`rightsMD` pattern common in Archivematica METS, which our own comments in `MetsManager.DeleteDiv` cite as a case to support:

```xml
<mets:file ADMID="ADM_rights ADM_doc">   <!-- ADM_rights has only a rightsMD; ADM_doc holds the techMD -->
```

Three separate failures follow, all reproduced:

1. **`MetsParser.cs:429`** — `techMd` resolves to the rights amdSec and `haveUsedAdmIdAlready` is set, so the real techMD is never consulted. The parsed `WorkingFile` has `Digest = null`, `Size = 0` and no `FileFormatMetadata`. The binary then reaches `GetDiffImportJob` **with no SHA256**, which that contract requires.
2. **`MetadataManager.cs:204` → `:312`** — the resolved amdSec has no techMD, and `SetAmdSec`'s else-branch indexes `ctx.AmdSec.TechMd[0]`, throwing `ArgumentOutOfRangeException` out of `EditMets`/`HandleSingleFileUpload`. That is an unhandled 500 on a path where every other failure is a `Result.Fail`. Note `:212` two lines earlier already uses `TechMd.FirstOrDefault()?`, so the code knows the collection can be empty.
3. **`MetsCache.cs:116`** — a directory div with `ADMID="ADM_rights ADM_sub"` resolves the rights section, yields no `premis:originalName`, and so gets no path at all. Diagnostic: `Directory div 'PHYS_sub' has no premis:originalName via ADMID '…'`. Every edit beneath that div then fails with the incomplete-path error.

## Why it matters more than it looks

Before #213 a multi-token ADMID resolved **nothing** — the wrong answer, but a safe one. Since #213 it resolves the **wrong section**, silently. We replaced a visible failure with an invisible one.

## Suggested fix

Cheap, because `lookupById` is already a `Func`: each call site passes a predicate that rejects unusable candidates, and resolution walks on to the next token.

- amdSec-for-metadata: `a.Id == id && a.TechMd.Count > 0`
- path resolution: `… && ExtractPremisOriginalName(a) != null`
- parser: require a `premis:object`

`MetsParser.cs:429` already demonstrates the compound-per-candidate pattern.

Guard `MetadataManager.cs:312` regardless — an empty `TechMd` should be a `Result.Fail`, not an exception.

## Note for whoever picks this up

`IdRefsMetsManagerTests.Deleting_A_File_With_Genuine_Multi_Admid_…` currently passes only because the real amdSec happens to be at index 0. Reversing the order in that fixture breaks it, which makes it a good starting test.

### LINKED ISSUE #216: AddNewFile's atomicity guarantee is false; a failed add orphans a FILE and the retry mints a duplicate xs:ID
Found by the cumulative adversarial review of the #188 work. The atomicity claim was introduced by `d7fde2f` (PR #211), so this is ours. Nothing is in production yet.

## What happens

`MetsManager.AddNewFile` (`MetsManager.cs:196`) says:

> Nothing is attached to the METS until the fallible metadata step has succeeded, so a failed add leaves the document (and the cache) exactly as it was.

That is not true. `ProcessAllFileMetadata` **does** attach: it adds the `FileType` to `FileSec.FileGrp[0]` (`MetadataManager.cs:66`) and the `AmdSecType` to `Mets.AmdSec` before returning Ok. The step that runs *after* it — `PopulateDmdFromResource` — is fallible, returning `ModsUnavailable` when the div's dmdSec cannot be materialised as MODS.

Reproduced sequence:

1. The document already contains a dmdSec whose ID equals the one the add would mint (`DMD_objects_x002F_new.tif`) with a non-MODS `MdWrap`. `AddToMets` for a file carrying access restrictions fails with "could not be read or created as MODS" — but the fileSec now holds an orphaned `FILE_objects_x002F_new.tif` and a matching amdSec, with **no div anywhere**.
2. Retry after fixing the metadata: it succeeds, and the fileSec now holds **two FILE elements with the same `ID` and the same href** — an invalid document.
3. Any later update or delete of that path throws `InvalidOperationException: Sequence contains more than one matching element` from `SetFileAndFileGroup` (`MetsManager.cs:551`). The path is now unrepairable through the API.

## Suggested fix

Either make the claim true — have `ProcessAllFileMetadata` build the FILE and amdSec without attaching them, and let `AddNewFile` attach everything after the last fallible step — or drop the claim and make the add roll back what it attached. The first is closer to the intent of `d7fde2f`.

Whichever way, `AddNewFile` should not be able to leave a FILE or amdSec behind without a div, and a retry must not be able to mint a duplicate `xs:ID`.

## Test coverage

`Failed_File_Add_Leaves_Mets_And_Cache_Untouched_And_Retry_Succeeds` exists but only covers the *early* (digest-mismatch) failure, which returns before the attach. A test for the late failure is what this needs.

### LINKED ISSUE #217: The dmdSec sibling sweep drops live references to legacy space-containing sections (silent metadata loss)
Found by the cumulative adversarial review of the #188 work. Introduced by `62e9dfa` (PR #213), while addressing a round-2 review comment — so this is ours, and it reintroduced the exact bug class the IDREFS work existed to close. Nothing is in production yet.

## What happens

`ModsManager.TryRemoveRedundantDmd` ends by sweeping sibling DMDID tokens that resolve nothing (`ModsManager.cs:242`):

```csharp
if (mets.DmdSec.All(d => d.Id != token))
{
    div.Dmdid.RemoveAt(i);
}
```

The comparison is verbatim, token by token. But the whole premise of `IdRefs` is that a **legacy space-containing dmdSec ID arrives from the XmlSerializer as several tokens, none of which equals any `d.Id`**. So every token of a *live* legacy reference looks dangling, and all of them are removed.

Reproduced: a div whose DMDID holds its own `DMD_objects` plus the two tokens `DMD_objects/my` and `file.pdf` of the real dmdSec `DMD_objects/my file.pdf`. Clearing the div's MODS correctly prunes its own dmdSec, then sweeps both remaining tokens — `Dmdid` ends **empty**. The `DMD_objects/my file.pdf` section survives in the document but is now orphaned, and the div has silently lost its descriptive record.

## Why it matters

This is data loss, it is silent, and it bites specifically on the legacy documents #214's headline promise covers — that METS written before the fix stays fully readable and editable. It should not outlive the step 2 release.

## Suggested fix

Mirror `IdRefs.ResolveAll`: a token is dangling only if **neither it nor the joined form of the surviving collection** resolves to a dmdSec. The tiered logic already exists; the sweep just needs to use it instead of comparing raw tokens.

tomcrane:
Fixes #215, fixes #216, fixes #217 — the three defects the cumulative #188 review found that **this body of work introduced**, rather than inherited. One commit per issue, each with a regression test verified to fail without its fix.

These were held back from #214 deliberately: that PR was review-ready and #215 touches the exact `MetsCache` lines its performance fix rewrote. The gate for these is the **first release**, not the merge — nothing in #188 is in production yet, so the whole effort can ship correct as a unit.

### #217 — a live legacy DMDID reference was being stripped
`ModsManager`'s sweep compared each DMDID token verbatim against `dmdSec.Id`. A legacy space-containing dmdSec ID arrives from the XmlSerializer as *several* tokens, none of which equals any ID on its own — so one live reference read as several dangling ones and all of them were removed. The div silently lost its descriptive record while the section survived, orphaned.

Now judged the way `IdRefs` resolves: if the joined form names a real dmdSec the tokens are one reference and all stay; only a genuine multi-token list is pruned token by token.

Introduced by `62e9dfa`, while addressing a round-2 review comment on #213 — it reintroduced the bug class the IDREFS work existed to close. It also contradicted #214's headline promise that legacy documents stay fully editable.

### #215 — IDREFS resolved to the first candidate that *existed*, not one the caller could use
ADMID is a list whose order carries no meaning, so a file or directory may name a shared `rightsMD` before the section that actually describes it — the Archivematica shape our own `DeleteDiv` comments cite as one to support. Three failures followed, all reproduced:

- the parser stopped on the rights section, found no fixity, and handed the binary on **with no SHA256** — which `GetDiffImportJob` requires;
- an update resolved the rights section and threw `ArgumentOutOfRangeException` indexing `TechMd[0]` — an unhandled 500 on the upload path;
- a directory resolved no `premis:originalName` and got no path at all, taking every edit beneath it with it.

Each site now passes a predicate that rejects unusable candidates so resolution walks on: PREMIS present (parser), a techMD present (`MetadataManager`), an `originalName` present (`MetsCache`). `SetAmdSec` also writes a techMD rather than indexing an absent one.

Worth noting what changed in #213: before it, these ADMIDs resolved **nothing** — the wrong answer, but a safe and visible one. #213 replaced that with silently resolving the wrong section.

### #216 — a late failure in an add orphaned a FILE, and the retry minted a duplicate ID
`AddNewFile` promised in a comment that nothing is attached until the fallible metadata step has succeeded. `ProcessAllFileMetadata` attaches the FILE and amdSec itself, and the step *after* it is fallible — so a failure there stranded both with no div, the retry minted a second FILE with the same `xs:ID`, and from then on `SetFileAndFileGroup` threw on every update and delete of that path.

The descriptive step now runs first, while nothing is attached. The one thing that lands early is a dmdSec it may create, so a later failure removes it again.

### Verification

Each new test was checked by reverting its fix and confirming the test fails — four of the five #215 tests fail without the predicates (the fifth is the over-correction guard, which should pass either way), and both the #216 and #217 tests fail without theirs.

XmlGen 467 passing in Debug and Release, Storage.API 55, Preservation.API 106, Core 68, Common.Model 8.

### DIFF
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetadataManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetadataManager.cs
@@ -216,11 +216,18 @@ public class MetadataManager(PremisManager premisManager, PremisManagerExif prem
         // IdRefs handles both; FileAdmId is then the RESOLVED amdSec's actual ID (identical
         // to the rejoined tokens for legacy content), so IDs derived from it - the ClamAV
         // digiprovMD ID - always embed a real amdSec ID.
-        var amdSec = IdRefs.ResolveSingle(ctx.File.Admid, id => fullMets.Mets.AmdSec.FirstOrDefault(a => a.Id == id));
+        // The candidate must actually carry technical metadata. ADMID is a LIST and its order
+        // means nothing, so a file may name a shared rightsMD before the section holding its
+        // own techMD - the Archivematica shape. Accepting the first section that merely EXISTS
+        // would hand back the rights section, and the update would then write this file's
+        // PREMIS into it (or throw indexing TechMd[0]) - issue #215.
+        var amdSec = IdRefs.ResolveSingle(ctx.File.Admid,
+            id => fullMets.Mets.AmdSec.FirstOrDefault(a => a.Id == id && a.TechMd.Count > 0));
         if (amdSec == null)
         {
             return Result.Fail(ErrorCodes.BadRequest,
-                $"No amdSec found for ADMID '{IdRefs.Joined(ctx.File.Admid)}' of file {operationPath}");
+                $"No amdSec carrying technical metadata found for ADMID " +
+                $"'{IdRefs.Joined(ctx.File.Admid)}' of file {operationPath}");
         }
         ctx.FileAdmId = amdSec.Id;
         ctx.AmdSec = amdSec;
@@ -322,9 +329,25 @@ public class MetadataManager(PremisManager premisManager, PremisManagerExif prem
                 },
             };
         }
-        else
+        else if (ctx.AmdSec.TechMd.Count > 0)
         {
             ctx.AmdSec.TechMd[0].MdWrap.XmlData = new MdSecTypeMdWrapXmlData { Any = { premisXml } };
+        }
+        else
+        {
+            // Resolution only accepts an amdSec that already has a techMD, so this is
+            // unreachable today - but it used to throw here rather than degrade, and an
+            // unhandled index is a 500 on the upload path where everything else is a
+            // Result.Fail. Writing the techMD we were about to update is the repair.
+            ctx.AmdSec.TechMd.Add(new MdSecType
+            {
+                Id = ctx.TechId,
+                MdWrap = new MdSecTypeMdWrap
+                {
+                    Mdtype = MdSecTypeMdWrapMdtype.PremisObject,
+                    XmlData = new MdSecTypeMdWrapXmlData { Any = { premisXml } }
+                }
+            });
         }
     }
 }
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetadataManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetadataManager.cs
@@ -221,12 +221,14 @@ public class MetadataManager(PremisManager premisManager, PremisManagerExif prem
         // own techMD - the Archivematica shape. Accepting the first section that merely EXISTS
         // would hand back the rights section, and the update would then write this file's
         // PREMIS into it (or throw indexing TechMd[0]) - issue #215.
-        var amdSec = IdRefs.ResolveSingle(ctx.File.Admid,
-            id => fullMets.Mets.AmdSec.FirstOrDefault(a => a.Id == id && a.TechMd.Count > 0));
-        if (amdSec == null)
+        var amdSec = IdRefs.ResolveSingle(
+            ctx.File.Admid,
+            id => fullMets.Mets.AmdSec.FirstOrDefault(a => a.Id == id),
+            CarriesPremisObject);
+        if (amdSec == null || !CarriesPremisObject(amdSec))
         {
             return Result.Fail(ErrorCodes.BadRequest,
-                $"No amdSec carrying technical metadata found for ADMID " +
+                $"No amdSec carrying PREMIS technical metadata found for ADMID " +
                 $"'{IdRefs.Joined(ctx.File.Admid)}' of file {operationPath}");
         }
         ctx.FileAdmId = amdSec.Id;
@@ -236,6 +238,22 @@ public class MetadataManager(PremisManager premisManager, PremisManagerExif prem
 
         return Result.Ok();
     }
+
+    /// <summary>
+    /// True when this amdSec carries the file's PREMIS object — the thing the metadata paths
+    /// actually read and patch. Merely having SOME techMD is not enough: an Archivematica-shaped
+    /// ADMID can name a section whose techMD holds only FITS or EXIF output, and treating that
+    /// as usable stops resolution on the wrong section, then hands non-PREMIS XML to
+    /// <c>GetPremisComplexType()</c> to deserialise into a null it does not expect. This is the
+    /// same check the parser and the path cache use, so all three agree on what "usable" means.
+    /// </summary>
+    private static bool CarriesPremisObject(AmdSecType amdSec) =>
+        amdSec.TechMd.Any(techMd =>
+            techMd.MdWrap?.XmlData?.Any?.FirstOrDefault() is XmlElement element &&
+            (element.LocalName == "object" && element.NamespaceURI == PremisNamespace ||
+             element.GetElementsByTagName("object", PremisNamespace).Count > 0));
+
+    private const string PremisNamespace = "http://www.loc.gov/premis/v3";
 
     private static void SetFileAndFileGroup(ProcessingContext ctx, DivType? div, FullMets fullMets)
     {
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsCache.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsCache.cs
@@ -127,10 +127,18 @@ public static class MetsCache
             case Constants.DirectoryType:
             {
                 // ADMID is an IDREFS token collection; IdRefs resolves both legacy
-                // space-containing IDs and genuine multi-references.
-                var amdSec = IdRefs.ResolveSingle(child.Admid, id => index != null
-                    ? index.AmdSecById(id)
-                    : mets.AmdSec.FirstOrDefault(a => a.Id == id));
+                // space-containing IDs and genuine multi-references. The candidate must
+                // actually carry an originalName: ADMID is a LIST whose order means nothing,
+                // so a directory may name a shared rightsMD before the section describing it,
+                // and stopping at the first section that merely EXISTS would leave the div with
+                // no path at all — taking every edit beneath it down with it (issue #215).
+                var amdSec = IdRefs.ResolveSingle(child.Admid, id =>
+                {
+                    var candidate = index != null
+                        ? index.AmdSecById(id)
+                        : mets.AmdSec.FirstOrDefault(a => a.Id == id);
+                    return ExtractPremisOriginalName(candidate) != null ? candidate : null;
+                });
                 path = ExtractPremisOriginalName(amdSec);
                 if (path == null)
                 {
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsCache.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsCache.cs
@@ -132,14 +132,26 @@ public static class MetsCache
                 // so a directory may name a shared rightsMD before the section describing it,
                 // and stopping at the first section that merely EXISTS would leave the div with
                 // no path at all — taking every edit beneath it down with it (issue #215).
-                var amdSec = IdRefs.ResolveSingle(child.Admid, id =>
-                {
-                    var candidate = index != null
+                // The name of the last candidate tested, so the winner isn't parsed twice: this
+                // runs for every directory div on every cache build.
+                AmdSecType? lastTested = null;
+                string? lastTestedName = null;
+
+                var amdSec = IdRefs.ResolveSingle(
+                    child.Admid,
+                    id => index != null
                         ? index.AmdSecById(id)
-                        : mets.AmdSec.FirstOrDefault(a => a.Id == id);
-                    return ExtractPremisOriginalName(candidate) != null ? candidate : null;
-                });
-                path = ExtractPremisOriginalName(amdSec);
+                        : mets.AmdSec.FirstOrDefault(a => a.Id == id),
+                    candidate =>
+                    {
+                        lastTested = candidate;
+                        lastTestedName = ExtractPremisOriginalName(candidate);
+                        return lastTestedName != null;
+                    });
+
+                path = ReferenceEquals(amdSec, lastTested)
+                    ? lastTestedName
+                    : ExtractPremisOriginalName(amdSec);
                 if (path == null)
                 {
                     diagnostics?.Add(
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
@@ -193,14 +193,31 @@ public class MetsManager(
             Fptr = { new DivTypeFptr { Fileid = fileId } }
         };
 
-        // Nothing is attached to the METS until the fallible metadata step has succeeded,
-        // so a failed add leaves the document (and the cache) exactly as it was.
-        var metadataResult = metadataManager.ProcessAllFileMetadata(fullMets, childItemDiv, workingFile, localPath, true);
-        if (metadataResult.Failure)
-            return metadataResult;
+        // Every fallible step runs before anything is attached, so a failed add leaves the
+        // document and the cache exactly as they were. The order matters: ProcessAllFileMetadata
+        // ATTACHES the FILE and the amdSec (it fails, when it fails, before doing so), so it has
+        // to come after the descriptive-metadata step rather than before it - otherwise a
+        // failure there strands a FILE and an amdSec with no div, and the retry mints a SECOND
+        // FILE with the same xs:ID, after which the path can no longer be updated or deleted
+        // at all (issue #216).
+        //
+        // The one thing that does land early is a dmdSec, which the descriptive step may create;
+        // if the step after it fails, that section is removed again.
+        var dmdSecCountBefore = fullMets.Mets.DmdSec.Count;
 
         var dmdResult = PopulateDmdFromResource(fullMets, workingFile, childItemDiv, localPath);
-        if (dmdResult.Failure) return dmdResult;
+        if (dmdResult.Failure)
+        {
+            RemoveDmdSecsAddedSince(fullMets, dmdSecCountBefore);
+            return dmdResult;
+        }
+
+        var metadataResult = metadataManager.ProcessAllFileMetadata(fullMets, childItemDiv, workingFile, localPath, true);
+        if (metadataResult.Failure)
+        {
+            RemoveDmdSecsAddedSince(fullMets, dmdSecCountBefore);
+            return metadataResult;
+        }
 
         parentDiv.Div.Add(childItemDiv);
         CacheAddedDiv(fullMets, localPath, childItemDiv);
@@ -249,6 +266,18 @@ public class MetsManager(
 
         SortChildDivs(parentDiv);
         return Result.Ok();
+    }
+
+    /// <summary>
+    /// Undo the dmdSecs an abandoned add created. They are appended, so anything past the
+    /// count taken before the step belongs to it.
+    /// </summary>
+    private static void RemoveDmdSecsAddedSince(FullMets fullMets, int dmdSecCountBefore)
+    {
+        while (fullMets.Mets.DmdSec.Count > dmdSecCountBefore)
+        {
+            fullMets.Mets.DmdSec.RemoveAt(fullMets.Mets.DmdSec.Count - 1);
+        }
     }
 
     /// <summary>
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
@@ -1,4 +1,4 @@
-using System.Diagnostics;
+﻿using System.Diagnostics;
 using System.Xml;
 using DigitalPreservation.Common.Model;
 using DigitalPreservation.Common.Model.Results;
@@ -203,19 +203,19 @@ public class MetsManager(
         //
         // The one thing that does land early is a dmdSec, which the descriptive step may create;
         // if the step after it fails, that section is removed again.
-        var dmdSecCountBefore = fullMets.Mets.DmdSec.Count;
+        var dmdSecsBefore = fullMets.Mets.DmdSec.ToHashSet();
 
         var dmdResult = PopulateDmdFromResource(fullMets, workingFile, childItemDiv, localPath);
         if (dmdResult.Failure)
         {
-            RemoveDmdSecsAddedSince(fullMets, dmdSecCountBefore);
+            RemoveDmdSecsAddedSince(fullMets, dmdSecsBefore);
             return dmdResult;
         }
 
         var metadataResult = metadataManager.ProcessAllFileMetadata(fullMets, childItemDiv, workingFile, localPath, true);
         if (metadataResult.Failure)
         {
-            RemoveDmdSecsAddedSince(fullMets, dmdSecCountBefore);
+            RemoveDmdSecsAddedSince(fullMets, dmdSecsBefore);
             return metadataResult;
         }
 
@@ -269,14 +269,19 @@ public class MetsManager(
     }
 
     /// <summary>
-    /// Undo the dmdSecs an abandoned add created. They are appended, so anything past the
-    /// count taken before the step belongs to it.
+    /// Undo the dmdSecs an abandoned add created — identified by REFERENCE against a snapshot
+    /// taken beforehand, not by position. A count-and-truncate would remove whatever happens to
+    /// sit at the end, so anything another edit against the same in-memory FullMets legitimately
+    /// added in the meantime would be destroyed instead. MetsManager is not thread-safe today
+    /// (see the bug/metsmanager-thread-safety work), which is exactly why this should not add a
+    /// new way for concurrent edits to lose unrelated metadata.
     /// </summary>
-    private static void RemoveDmdSecsAddedSince(FullMets fullMets, int dmdSecCountBefore)
+    private static void RemoveDmdSecsAddedSince(FullMets fullMets, HashSet<MdSecType> dmdSecsBefore)
     {
-        while (fullMets.Mets.DmdSec.Count > dmdSecCountBefore)
+        var added = fullMets.Mets.DmdSec.Where(dmdSec => !dmdSecsBefore.Contains(dmdSec)).ToList();
+        foreach (var dmdSec in added)
         {
-            fullMets.Mets.DmdSec.RemoveAt(fullMets.Mets.DmdSec.Count - 1);
+            fullMets.Mets.DmdSec.Remove(dmdSec);
         }
     }
 
--- a/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
+++ b/src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
@@ -1,4 +1,4 @@
-﻿using System.Diagnostics;
+using System.Diagnostics;
 using System.Xml;
 using DigitalPreservation.Common.Model;
 using DigitalPreservation.Common.Model.Results;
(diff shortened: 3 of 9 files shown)


src/DigitalPreservation/DigitalPreservation.Mets/MetadataManager.cs
```diff
@@ -216,11 +216,18 @@ private static Result GetMetadataXml(ProcessingContext ctx, FullMets fullMets, D
         // IdRefs handles both; FileAdmId is then the RESOLVED amdSec's actual ID (identical
         // to the rejoined tokens for legacy content), so IDs derived from it - the ClamAV
         // digiprovMD ID - always embed a real amdSec ID.
-        var amdSec = IdRefs.ResolveSingle(ctx.File.Admid, id => fullMets.Mets.AmdSec.FirstOrDefault(a => a.Id == id));
+        // The candidate must actually carry technical metadata. ADMID is a LIST and its order
+        // means nothing, so a file may name a shared rightsMD before the section holding its
+        // own techMD - the Archivematica shape. Accepting the first section that merely EXISTS
+        // would hand back the rights section, and the update would then write this file's
+        // PREMIS into it (or throw indexing TechMd[0]) - issue #215.
+        var amdSec = IdRefs.ResolveSingle(ctx.File.Admid,
```
brianmac1967: This "usable candidate" check only verifies `TechMd.Count > 0`, weaker than the semantically-correct check (does it actually carry a `premis:object`) used by the two sibling fixes for the same issue (#215) in `MetsParser.cs` and `MetsCache.cs`.

On Archivematica-style METS, an ADMID can list a rights-only section first and, before the real PREMIS techMD, an amdSec whose TechMd holds only FITS/EXIF output. `TechMd.Count > 0` accepts that non-PREMIS section as "usable" and resolution stops there instead of walking on — the same bug class #215 fixed, reintroduced for this narrower input shape. Downstream, `ctx.PremisIncExifXml` gets set from non-PREMIS XML and `.GetPremisComplexType()!` in `ProcessFileFormatDataForFile` can deserialize to null and null-deref, or silently patch PREMIS fields with garbage.

src/DigitalPreservation/DigitalPreservation.Mets/MetsCache.cs
```diff
@@ -127,10 +127,18 @@ private static void Walk(
...
-                // space-containing IDs and genuine multi-references.
-                var amdSec = IdRefs.ResolveSingle(child.Admid, id => index != null
-                    ? index.AmdSecById(id)
-                    : mets.AmdSec.FirstOrDefault(a => a.Id == id));
+                // space-containing IDs and genuine multi-references. The candidate must
+                // actually carry an originalName: ADMID is a LIST whose order means nothing,
+                // so a directory may name a shared rightsMD before the section describing it,
+                // and stopping at the first section that merely EXISTS would leave the div with
+                // no path at all — taking every edit beneath it down with it (issue #215).
+                var amdSec = IdRefs.ResolveSingle(child.Admid, id =>
```
brianmac1967: The new "usability" predicate is applied uniformly to both the joined (legacy multi-token) resolution tier and the per-token tier of `IdRefs.ResolveSingle`. A legacy multi-token ID whose joined form matches a real-but-unusable section (no techMD / no `premis:originalName`) now falls through to per-token matching instead of correctly reporting the "no path"/"no PREMIS" diagnostic — an uncovered fallthrough path.

Concretely: a legacy directory div has a single space-containing ADMID that legitimately names its own amdSec, but that amdSec lacks `premis:originalName` (malformed/partial legacy content). This used to correctly report the diagnostic; now it falls through to resolving each half of the split legacy ID as an independent token — if either half coincidentally matches some other real amdSec ID elsewhere in the document, the div silently resolves to the wrong section's path instead of failing loudly.

src/DigitalPreservation/DigitalPreservation.Mets/MetsCache.cs
```diff
@@ -127,10 +127,18 @@ private static void Walk(
...
+                {
+                    var candidate = index != null
+                        ? index.AmdSecById(id)
+                        : mets.AmdSec.FirstOrDefault(a => a.Id == id);
+                    return ExtractPremisOriginalName(candidate) != null ? candidate : null;
```
brianmac1967: Minor: `ExtractPremisOriginalName(candidate)` is called once inside the `ResolveSingle` predicate to test each candidate, then a second time on the winning candidate to extract the actual path — re-parsing the same `XmlElement` twice. No correctness impact, but the predicate could return (and reuse) the already-extracted name to save a redundant traversal on every directory-div resolution.

brianmac1967 (commented):
Reviewed the #215/#216/#217 follow-up fixes. All 467 tests pass. Six findings below (three correctness bugs, one test-coverage gap, one robustness concern, one minor inefficiency) — none blocking, but #1, #4 and #6 are the same bug classes #215/#217 were meant to close, just reintroduced for narrower/mixed input shapes.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -251,6 +268,18 @@ private Result AddNewDirectory(DivType parentDiv, FullMets fullMets, WorkingDire
         return Result.Ok();
     }
 
+    /// <summary>
+    /// Undo the dmdSecs an abandoned add created. They are appended, so anything past the
+    /// count taken before the step belongs to it.
+    /// </summary>
+    private static void RemoveDmdSecsAddedSince(FullMets fullMets, int dmdSecCountBefore)
```
brianmac1967: `RemoveDmdSecsAddedSince` rolls back by comparing `fullMets.Mets.DmdSec.Count` to a snapshot taken before the fallible steps, rather than removing specific captured references. Fragile if the same `FullMets` instance is ever touched concurrently — there's an open thread-safety concern for `MetsManager` (`bug/metsmanager-thread-safety` branch). If two edits against the same in-memory `FullMets` interleave, one thread's failed-add rollback could delete a dmdSec another thread just legitimately added, silently dropping unrelated, valid descriptive metadata instead of only its own.

src/DigitalPreservation/DigitalPreservation.Mets/ModsManager.cs
```diff
@@ -276,6 +266,35 @@ private static bool TryRemoveRedundantDmd(DigitalPreservation.XmlGen.Mets.Mets m
...
+    /// time against <c>dmdSec.Id</c>. A legacy space-containing dmdSec ID arrives from the
+    /// XmlSerializer as SEVERAL tokens, none of which equals any dmdSec ID on its own, so a
+    /// verbatim per-token test reads one live reference as several dangling ones and strips
+    /// it — silently costing the div its descriptive record while orphaning the section
+    /// (issue #217). So: if the joined form names a real dmdSec, the tokens are one reference
+    /// and all of them stay; only a genuine multi-token list is pruned token by token.
+    /// </remarks>
+    private static void SweepDanglingDmdTokens(DigitalPreservation.XmlGen.Mets.Mets mets, DivType div)
+    {
+        if (div.Dmdid.Count > 1 && mets.DmdSec.Any(d => d.Id == IdRefs.Joined(div.Dmdid)))
```
brianmac1967: `SweepDanglingDmdTokens`'s fix only handles the pure-legacy case (the whole remaining `Dmdid` list joins to one real dmdSec ID). A div whose `Dmdid` mixes one legacy multi-token reference with a separate genuine single-token reference isn't tested and, per the same joined-vs-per-token tiering limitation as the MetsCache.cs finding above, can still have its legacy tokens incorrectly pruned.

E.g. `div.Dmdid = ["DMD_valid", "DMD_legacy/my", "file.pdf"]` where the last two tokens together name one live legacy dmdSec and the first names a separate live one. The joined-all-three form won't match any dmdSec ID, so the joined-tier early return doesn't fire, and the per-token fallback strips the two legacy tokens individually — silently orphaning the legacy dmdSec exactly as #217 originally described, just for a mixed-reference div.

src/DigitalPreservation/XmlGen.Tests/PhysicalDivsCacheTests.cs
```diff
@@ -359,6 +359,62 @@ public async Task Failed_File_Add_Leaves_Mets_And_Cache_Untouched_And_Retry_Succ
         AssertCacheMatchesRebuild(fullMets);
     }
 
+    [Fact]
+    public async Task A_File_Add_That_Fails_LATE_Leaves_No_Orphan_And_The_Retry_Adds_One_File()
```
brianmac1967, line 363: This #216 test only exercises the zero-rollback case: `PopulateDmdFromResource` fails immediately via the pre-existing `blocker` dmdSec, so 0 dmdSecs are ever created. The branch where a dmdSec is actually created by `PopulateDmdFromResource` (e.g. because `AccessRestrictions` was set) and `ProcessAllFileMetadata` fails afterward — requiring an actual rollback in `MetsManager.cs` lines 215-220 — is never exercised.

If a future edit breaks rollback specifically for the "dmdSec really was created" case, no test would catch it — the add would leak a dmdSec into the METS with the suite staying green, contrary to the PR's claim that each new test was verified to fail without its fix. Might be worth a second case where the blocker is something that lets `PopulateDmdFromResource` succeed but a later step fails.

brianmac1967:
Note for the team: heads up that this PR (#218) does **not** address the two dormant findings from the earlier PR #214 follow-up review — `MetsManager.PathForDiv`'s O(n) linear cache scan, and the encoded/legacy-ID pair being reconstructed independently in `FindConflictingChild`/`FindChildDivByPath` instead of through a shared helper. Neither is touched by this diff. #218 actually covers #215/#216/#217 (IDREFS candidate-selection fixes) instead — worth confirming those two PathForDiv/duplication items still have a tracking ticket somewhere, since they're not closed by this PR.

tomcrane:
All six addressed — five changed code, one is a deliberate non-fix. Findings 1, 2 and 5 turned out to share a root cause, and you were right that they were the same bug classes reintroduced.

## The shared cause

I applied the "usable candidate" predicate to **both tiers** of `IdRefs` resolution, which is wrong, and your finding 2 is the clearest statement of why.

The joined form is an **identity match**: it says *this one section is what the reference names*. If it resolves, that is the answer whether we can use it or not — returning it lets the caller report a precise diagnostic. Walking on resolves a *fragment* of a legacy ID against some unrelated section and answers with confident nonsense. Only the per-token tier is a genuine list whose order carries no meaning, and only there should an unusable candidate be skipped.

There are now three-argument `ResolveSingle` overloads implementing exactly that asymmetry, once, for both stacks — so the three call sites can't drift on it again.

**Finding 1** is the same thing from the other end: `TechMd.Count > 0` was weaker than the two siblings, and accepted a FITS/EXIF-only section. All three sites now agree that "usable" means carrying a `premis:object`, and there's a test asserting a FITS section is not overwritten with the file's PREMIS.

**Finding 5** — the mixed-reference sweep — now matches **runs** of consecutive tokens that name a real dmdSec, longest first, so a legacy ID wins over its own fragments. Your exact example is the test.

## The rest

**Finding 4** — rollback now identifies what it created by reference, not position. Your framing is the one I've put in the code: a count-and-truncate removes whatever sits at the end, so under interleaved edits against one `FullMets` it would destroy metadata another edit legitimately added.

**Finding 3** — fixed; the winner's `originalName` is no longer parsed twice. Worth taking given R1 was a performance regression in the same code path, and this ran for every directory div on every cache build.

**Finding 6 is the one I'm most glad you caught.** You're right that the rollback branch was never exercised — the test asserted a count that was zero either way, so it could not have failed. That is exactly the class of defect this whole review process exists to catch, and I introduced one while fixing the others. There is now a case where the descriptive step succeeds (creating a dmdSec) and the metadata step then fails.

## Two of my new tests didn't pin their fix

Verifying by reverting each fix, two passed anyway:

- the legacy-identity test's decoy section had no PREMIS, so it was unusable on either behaviour and the test could not distinguish them. The decoy now carries PREMIS, so the broken behaviour resolves to it and writes the file's metadata into an unrelated section. It now fails without the fix.
- the by-reference rollback test could not be made to fail without a seam that would exist only for testing — the snapshot is taken inside `AddNewFile`, so a test can't append between snapshot and rollback. **I removed it rather than leave something that looks like coverage.** The change stands on the reasoning in the doc comment, not on a test, and I'd rather say so than have a green test implying otherwise.

XmlGen 476 passing in Debug and Release, plus the rest of the solution.

tomcrane:
Good check, and the answer differs for the two.

**The encoded/legacy ID-pair duplication is already fixed** — in #214 itself, commit [`d33ac91`](https://github.com/digirati-co-uk/digital-preservation/commit/d33ac91), which merged to `main` as part of that PR. `PhysicalDivIdCandidates(localPath)` is now the single source of the `(Encoded, Legacy)` pair and both `FindChildDivByPath` and `FindConflictingChild` go through it; `FindConflictingChild` no longer takes the minted ID as a parameter either. So it isn't in #218 because it wasn't outstanding when #218 was branched, not because it was missed. Confirmed present on `main`.

**`PathForDiv`'s O(n) scan is genuinely not fixed, and you're right that it has no ticket.** It was deliberate, and the reasoning is in the xmldoc on the method — but your asking is itself the argument that the reasoning isn't discoverable from outside the file, which is a fair hit.

Two things worth recording while we decide, because the obvious fix is a trap:

The reverse scan is doing more than recovering a path. **Only physical divs are in the cache, so a miss is what tells us the div is LOGICAL.** Swapping it for `MetsCache.TryResolvePath` looks like the same answer for less work, but a logical div of `TYPE="Item"` carrying an `fptr` resolves the path of the file it paints — so it would be handed the `DMD_` id of the *physical* div for that path and the two would silently share a dmdSec. That is R7 from the audit, arriving by another route.

And there is no cheaper per-call answer: resolving a Directory div scans the amdSecs, an Item div scans the fileSec, so every option is O(n) for a single call. The real fix belongs to the bulk caller — hoist a reverse map or a physical-div set across the loop — and **there is no bulk caller yet**; in fact there is no caller at all. All four `*ByDivId` methods are unreferenced outside tests: the UI edits physical resources by path via `SetModsInformation`, and logical ranges by re-applying the whole structMap via `SetStructMap`.

I argued against a standalone ticket earlier on the grounds that "when someone does X, remember Y" can't be started or closed and tends to become backlog noise, and that the constraint is better placed where someone would change it. @tomcrane agreed at the time. Your question suggests that trade-off leans the other way for reviewers, so I'm happy to raise one — my preference would be an issue about **whether the `ByDivId` shape earns its keep at all**, with the O(n) note as an acceptance criterion, rather than a perf ticket for a method nothing calls. Say the word and I'll open it.

tomcrane:
## What changed since your review — #218

Two commits, so you can review the delta rather than the whole diff again. Neither is a touch-up: `c0837f4` changes resolution *semantics*, so it is worth a look on its own terms.

**[`c0837f4`](https://github.com/digirati-co-uk/digital-preservation/commit/c0837f4) — the six findings.**

| Your finding | What changed |
|---|---|
| 1 — weak `TechMd.Count > 0` predicate | `MetadataManager` now uses the same PREMIS-object check as the parser and cache, so all three agree on "usable" |
| 2 — predicate applied to both tiers | New three-argument `IdRefs.ResolveSingle` overloads (both stacks) making the joined tier authoritative and skipping only in the per-token tier |
| 3 — double `ExtractPremisOriginalName` | Winner's name reused rather than re-parsed |
| 4 — count-based rollback | Now identifies what it created by reference |
| 5 — mixed legacy + genuine reference | New `LiveTokenIndexes`: matches *runs* of consecutive tokens, longest first |
| 6 — rollback never exercised | New test where the descriptive step creates a dmdSec and a later step fails |

**[`ba19411`](https://github.com/digirati-co-uk/digital-preservation/commit/ba19411)** — one-line parameter type, SonarCloud CA1859. Nothing else.

### Two things I'd particularly like your judgement on

**`LiveTokenIndexes` is new logic, not a tweak.** It solves the problem you identified, which puts it squarely in the fixes-of-fixes pattern this codebase keeps producing — the last three rounds have all been review of a fix. Longest-run-first matching is the part to attack.

**I removed a test rather than leave it looking like coverage.** The by-reference rollback couldn't be made to fail without a seam existing purely for testing: the snapshot is taken inside `AddNewFile`, so a test can't append between snapshot and rollback. That change now rests on the reasoning in its doc comment, not on a test. You may reasonably think that's the wrong trade.

Also worth knowing: verifying each new test by reverting its fix caught **two of my own that passed anyway** — one had a decoy section that was unusable under either behaviour, so it couldn't distinguish them. Both are noted in the commit message.

### Housekeeping

This branch was **rebased**: some of these fixes were briefly committed to #220's branch by mistake and have been moved here. If any earlier inline comments render as outdated, that is why — the code they refer to is unchanged.

CI green, SonarCloud clean on new code, XmlGen 471 passing in Debug and Release.

src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
```diff
@@ -425,9 +425,18 @@ private void ProcessChildStructDivs(MetsFileWrapper mets, XElement parent,
...
+                    // exactly that. Taking the first candidate that merely EXISTS would stop on
+                    // the rights section and lose the fixity, leaving the binary with no SHA256
+                    // (issue #215).
+                    techMd = IdRefs.ResolveSingle(
+                        admId,
+                        id => lookupMaps.TechMdMap.GetValueOrDefault(id)
+                              ?? lookupMaps.AmdSecMap.GetValueOrDefault(id),
+                        candidate => candidate.Descendants(XNames.PremisObject).Any());
 
                     if (techMd == null)
```
brianmac1967, line 453: The #2 fix (MetsCache.cs tiering) introduced a new asymmetric `IdRefs.ResolveSingle`: for the identity-match tier (a single-token ADMID, or a legacy space-containing ADMID matched as a whole), it returns the resolved candidate **unconditionally** — even when it fails the `isUsable` predicate — by design, expecting the caller to recheck and report a precise diagnostic rather than silently proceeding. `MetadataManager.cs` does this explicitly (`if (amdSec == null || !CarriesPremisObject(amdSec))`). `MetsCache.cs` does it implicitly, since it always re-derives `path` afterward and that naturally comes back null for an unusable candidate.

This call site has neither — it only checks `techMd == null` here, then unconditionally treats any non-null `techMd` as PREMIS-bearing in the `else` branch below (extracting fixity/size/originalName, all silently null/absent if it isn't).

**Failure scenario:** a legacy platform METS file whose ADMID is a single space-containing ID (or even a single non-legacy ID) that resolves, as the whole attribute value, to an existing-but-unusable amdSec — e.g. a shared rights-only section with no `premis:object`. `ResolveSingle` returns this candidate unconditionally per the new design; since `techMd != null`, the code proceeds into the `else` branch and `digest` stays null with no diagnostic. This is verbatim the original #215 failure ("parser stopped on the rights section, found no fixity, handed the binary on with no SHA256"), just now reachable via the identity-match tier instead of the original per-token bug. No test in `IdRefsCandidateSelectionTests.cs` exercises this path for `MetsParser` — the existing rights-section-first test uses a genuine two-token ADMID list (the per-token tier, which correctly filters), not a single/legacy identity-match ADMID.

Suggest an equivalent `techMd == null || !techMd.Descendants(XNames.PremisObject).Any()` guard here, with a regression test analogous to MetsCache's `A_Legacy_Reference_To_An_Unusable_Section_Fails_Loudly_Rather_Than_Guessing` but driving through `MetsParser`/`GetMetsFileWrapper`.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -1,4 +1,4 @@
-using System.Diagnostics;
+﻿using System.Diagnostics;
```
brianmac1967: Minor/cosmetic: this line picked up a stray UTF-8 BOM (`﻿using System.Diagnostics;`), almost certainly an incidental artifact of the edit tool rather than intentional. No functional effect, but worth stripping before merge if noticed.

brianmac1967 (commented):
Re-review after the follow-up commit (`c0837f49f` + Sonar cleanup `ba19411d1`) addressing the six prior comments. Five of six are fixed correctly and non-superficially, each with a targeted regression test:

- MetadataManager.cs usable-candidate check → replaced with a proper `CarriesPremisObject` helper matching the sibling checks, plus a new regression test.
- MetsCache.cs tiering → fixed at its site; the new asymmetric `IdRefs.ResolveSingle` design works here because the caller always re-derives `path`, which naturally comes back null for an unusable candidate.
- MetsCache.cs double-parse → now caches the tested candidate's result and reuses it.
- MetsManager.cs rollback fragility → now removes specific captured references via `HashSet` reference identity (confirmed `MdSecType` has no `Equals`/`GetHashCode` override, so this is correct reference-identity semantics) instead of comparing counts.
- ModsManager.cs mixed-reference pruning → new `LiveTokenIndexes` correctly handles runs of consecutive live tokens; hand-traced against the exact mixed example from the original comment and it holds.
- PhysicalDivsCacheTests.cs coverage gap → new test genuinely forces `PopulateDmdFromResource` to create a dmdSec and then fails a later step, exercising the actual rollback branch.

One new finding below that I'd flag as a merge blocker, plus one cosmetic note.

tomcrane:
Both addressed in [`fd427c9`](https://github.com/digirati-co-uk/digital-preservation/commit/fd427c9). The blocker was a real gap in the design I introduced, not just at that call site.

## The parser recheck

You've identified the actual weakness in the asymmetric `ResolveSingle`: making the identity tier return an unusable match **only works if every caller rechecks**, and I built that contract without enforcing it anywhere. Two of three sites honoured it — one explicitly, one by accident of re-deriving `path` — and the third didn't. That is a design that relies on each caller remembering, which is exactly the shape that produced the duplicated-ID-pair finding on #214.

`MetsParser` now nulls a candidate carrying no `premis:object` before the `else` branch, so all three sites behave alike.

**One correction to the failure scenario**, having built it: the *digest* outcome is unchanged — null either way, because the `else` branch simply finds no fixity in a rights section. What genuinely differs is the virus-scan lookup. With a non-null `techMd`, `amdScope` becomes that rights section and the structural search finds **any** ClamAV-prefixed digiprovMD inside it — so a shared rights section carrying another file's scan reports it as this file's. That is the misattribution class again, and it is what the new test pins.

**The test needed correcting before it meant anything.** My first version gave the foreign scan an ID matching *this* file's conventional key — so the fallback found it legitimately and the test failed for the wrong reason. It now names a different file, which is the only shape where scoping to the section is what misattributes. Verified by reverting the guard.

## The BOM

Well spotted, and it was mine rather than incidental: `MetsManager.cs` had no BOM, and my scripted edits rewrote it as `utf-8-sig`. Stripped, and I checked every `.cs` file in the cumulative diff for the same thing — it was the only one.

## State

XmlGen 472 passing on this branch in Debug and Release; #220 has been restacked on top and is at 478. Thank you for three rounds on this one — the first round's tiering finding and this one are the same underlying problem seen from two ends, and I would not have got there from one of them alone.

brianmac1967 (approved):

PR #220, opened 2026-08-13
fix: virus scans accumulate as provenance events (#219)

### LINKED ISSUE #219: A virus scan overwrites whatever digiprovMD is first, destroying other provenance and hiding its own result
Found by the cumulative adversarial review of the #188 work, but **pre-existing** — `fe6738a`, well before any of it. Independent of the identifier changes.

## What happens

`MetadataManager.AddVirusXml` overwrites `AmdSec.DigiprovMd[0]` whatever that element happens to be:

```csharp
if (ctx.AmdSec.DigiprovMd.Count != 0)
{
    ctx.AmdSec.DigiprovMd[0].MdWrap.XmlData = new MdSecTypeMdWrapXmlData { Any = { ctx.VirusXml } };
}
```

A `digiprovMD` holds *digital provenance* — a PREMIS **event**, of which a virus scan is only one kind. If a file's amdSec already carries any other event (an ingestion event, a format-migration event, anything a third-party producer put there), a virus scan **replaces its contents in place while keeping its original ID**.

Two things go wrong at once, and both are silent:

1. **The existing provenance is destroyed.** Its `eventIdentifier`, type, date and outcome are gone, overwritten by the ClamAV event.
2. **The virus result then cannot be read back.** Both readers look for an ID with the ClamAV prefix — `MetsParser.cs:507` (`StartsWith`) and `MetadataManager.cs:235` (`Contains`) — and the element now carries the *other* event's ID. Re-parsing the saved METS yields **no `VirusScanMetadata` at all**.

Reproduced: a file amdSec carrying `<digiprovMD ID="digiprovMD_ingest_doc">`, then an upload with `VirusScanMetadata`. The ingest event's content is replaced, the element keeps its old ID, and the round-trip reports no virus scan.

## Related, same two lines of concern

`MetadataManager.cs:235` finds the existing virus element with `x.Id.Contains(prefix)`:

- `MdSecType.Id` is optional, so an amdSec containing an ID-less `digiprovMD` — legal METS, and something `GetAmdSecType` can itself produce, since its `digiprovId` parameter defaults to null — throws `NullReferenceException` out of `EditMets`.
- It uses `Contains` where the parser uses `StartsWith`, so the two disagree about what counts as a ClamAV element.

## Suggested fix

`AddVirusXml` should select the ClamAV-prefixed `digiprovMD` and create one only if there isn't one, rather than indexing position zero. The reader at `:235` should use the same null-safe prefix test as the parser, so writer and reader agree on which element is the virus event.

tomcrane:
Fixes #219. **Stacked on #218** (`fix/188-review-followups`) — the base is that branch, not `main`, so the diff here shows only this change and #218's in-flight review is undisturbed. GitHub will retarget to `main` when #218 merges.

Unlike #215/#216/#217 this one is genuinely **pre-existing** (`fe6738a`), and independent of the identifier work.

## What was wrong

`AddVirusXml` overwrote `AmdSec.DigiprovMd[0]` — whatever that element happened to be. A `digiprovMD` records digital provenance of *any* kind, so if a file's amdSec already carried another event, a virus scan replaced its contents in place while keeping its original ID. Two things went wrong at once, both silently: the earlier provenance was destroyed, and the scan itself became unreadable, because both readers look for an ID with the ClamAV prefix and the element now carried someone else's ID.

## The model this now implements

Per Tom: **digiprov events accumulate.** A scan appends its own event no matter what is already in the list — virus scans or otherwise — and the file's current virus status is the *latest* ClamAV scan. Anything we don't recognise is left strictly alone.

So:

- a scan **appends** a new `digiprovMD`, with its ID suffixed (`…_2`, `…_3`) if the plain one is taken, since `xs:ID` must stay unique;
- nothing already present is read, reordered or modified — not earlier scans, and not events written by tools we neither recognise nor need to understand, including ones with **no ID at all**, which METS permits and which used to throw;
- only an *actual* scan writes an event, so an ordinary metadata update no longer rewrites the last one;
- reading follows the same rule: the virus status is the **last** ClamAV event, not the first.

The parser still reports a single `VirusScanMetadata` per file — the latest — so nothing downstream sees a behaviour change.

## ⚠️ This changes two existing test expectations

`Overwriting_File_Updates_Virus_Scan_From_Clean_To_Infected` asserted *"The digiprovMD must be updated in place, not duplicated"* and *"still exactly one digiprovMD (not two)"*. That encoded the overwrite model. Both assertions now check the history instead: two events under distinct IDs, with the later one reported. Worth a look, since it is a deliberate reversal rather than an incidental edit.

## New tests

- repeated scans accumulate, and the latest wins;
- foreign provenance — including an ID-less event — survives a scan untouched;
- an upload carrying no scan adds no event.

XmlGen 470 passing, Storage.API 55, Preservation.API 106, Core 68, Common.Model 8.


tomcrane:
## Knock-on analysis: what the accumulate change touches

Traced every consumer of virus status rather than reasoning about it. Short answer: **no behaviour change for any document this platform has already written, and no new UI or validity effects.** The detail, since it's worth being able to check:

**Why existing documents are unaffected.** Under the old code a file could only ever have *one* ClamAV event, because each scan overwrote the last. So "first matching" and "last matching" select the same element. The parser change from `FirstOrDefault` to `LastOrDefault` is therefore a no-op on every document in existence, and only starts to matter for scans written from now on — where it correctly reports the most recent.

**UI surfaces, all driven by the same one accessor:**

- `Deposit.cshtml:197` — the red "Files with viruses" banner, from `RootCombinedDirectory.GetFilesWithVirus()`.
- `Browse.cshtml:454` and `_RenderCombinedDirectoryAsTableRows.cshtml:167` — the per-file indicator.

All three go through `CombinedFile.GetVirusMetadata()`, which prefers the deposit-side metadata (from the Brunnhilde output) and falls back to the METS. The METS side now yields the latest scan, which is what those surfaces were always meant to show.

**Deposit validity / preservation gating:** none. `GetFilesWithVirus` has no callers in `Preservation.API` or `DigitalPreservation.Workspace` — the banner is informational and dismissible, and nothing refuses to build an ImportJob or preserve a deposit containing an infected file. That's unchanged by this PR, but worth knowing as a policy question in its own right.

**Mismatch detection** (`CombinedFile.cs:84-107`, comparing deposit-side against METS-side `HasVirus`/`VirusFound`/`VirusDefinition`) is unaffected: after a pipeline run both sides describe that run.

**iiif-builder** doesn't read digiprovMD at all.

### Two things to be aware of

1. **`WorkingFile.GetVirusScanMetadata()` uses `SingleOrDefault()`**, so it throws if a file ever carries two `VirusScanMetadata`. The parser emits exactly one — the latest — so this is safe, but it's a latent constraint: any future change that surfaced scan *history* through `WorkingFile.Metadata` would hit it. Worth a comment there at some point.

2. **METS files now grow with each pipeline run.** Previously a re-scan reused one element; now every run appends one per file. A deposit scanned N times carries N events per file. Fine at normal cadence, and it is the point of an event history, but it is a real change in how these documents grow — and for a large deposit re-scanned repeatedly it is the one thing I'd keep an eye on.

Also noted while tracing: the conventional-key fallback in `MetsParser` (used only when the structural lookup fails) matches the unsuffixed ID, so on a document with accumulated events it would find the *first*, not the latest. It can only fire for documents whose structure didn't resolve — which are third-party, and so have no suffixed IDs of ours — but it is an inconsistency worth knowing about rather than discovering later.

src/DigitalPreservation/DigitalPreservation.Mets/MetadataManager.cs
```diff
@@ -169,32 +169,20 @@ private Result ProcessFileFormatDataForFile(ProcessingContext ctx, WorkingFile w
...
         }
 
-        if (virusEventComplexType is null) return;
-        ctx.VirusXml = PremisEventManagerVirus.GetXmlElement(virusEventComplexType);
+        ctx.VirusXml = PremisEventManagerVirus.GetXmlElement(premisEventManagerVirus.Create(virusScan));
 
         if (ctx.AmdSec == null) return;
 
-        AddVirusXml(ctx);
+        AppendVirusEvent(ctx);
```
brianmac1967, line 185: Since scans now accumulate as new events instead of overwriting in place, there's no dedup/idempotency check here. A retried scan submission (client timeout retry, or an at-least-once SQS redelivery in Pipeline.API) will mint a brand-new duplicate `digiprovMD` event every time instead of being a no-op, which the old overwrite model was incidentally.

Not a correctness bug for the *reported* result — the read side's `LastOrDefault` still surfaces the latest scan correctly on the primary path — but it means the amdSec can accumulate an unbounded number of near-duplicate ClamAV events for what was really one scan, with no way to distinguish a genuine rescan from a retried duplicate. Worth considering whether `ProcessVirusDataForFile`/`AppendVirusEvent` should dedupe against an identical very-recent event before appending.

src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
```diff
@@ -513,9 +513,13 @@ private void ProcessChildStructDivs(MetsFileWrapper mets, XElement parent,
...
+                // LAST, not first: scans accumulate as separate events, so the file's current
+                // virus status is the most recently appended one. Everything else in the
+                // amdSec - earlier scans, and provenance from tools we don't recognise - is
+                // simply not ours to interpret.
                 var amdScope = techMd?.Name == XNames.MetsAmdSec ? techMd : techMd?.Parent;
                 var digiprovMd = amdScope?.Elements(XNames.MetsDigiprovMD)
-                    .FirstOrDefault(d => (d.Attribute("ID")?.Value ?? "")
+                    .LastOrDefault(d => (d.Attribute("ID")?.Value ?? "")
                         .StartsWith(Constants.VirusProvEventPrefix, StringComparison.OrdinalIgnoreCase));
                 if (digiprovMd == null)
```
brianmac1967, line 612: The conventional-key fallback just below (line 531: `var clamavKey = $"{Constants.VirusProvEventPrefix}{techMd?.Attribute("ID")?.Value ?? admId}";` → `lookupMaps.DigiprovMdMap.TryGetValue(clamavKey, ...)`) wasn't updated for the new multi-event model. `DigiprovMdMap` is keyed by exact ID, and only the *first* accumulated scan ever gets the plain, unsuffixed conventional ID — later scans get `_2`, `_3` suffixes via `MetadataManager.UnusedDigiprovId`. So this fallback can only ever resolve to the first-ever scan, never a later one.

This path is reachable: write-side `GetMetadataXml` only requires `TechMd.Count > 0`, but this read-side resolution additionally requires a `premis:object` descendant on `techMd` before the primary `LastOrDefault` path above even runs. For a file whose techMD is e.g. MIX/textMD instead of PREMIS, `amdScope` resolves to null, the primary path never executes, and this fallback kicks in — resolving to the stale first scan. A file rescanned and found infected could still report as clean for that class of document, which is the exact bug this PR set out to fix, just on the fallback path. Might be worth threading `LastOrDefault`/latest-suffix logic through the fallback too, or having it look up all matching conventional-key variants and take the last.

brianmac1967 (commented):
Reviewed the P1 fix (virus scans overwriting foreign provenance). The core fix is genuinely correct: `AppendVirusEvent`/`UnusedDigiprovId` now append rather than overwrite, select by the ClamAV prefix instead of index 0, and preserve foreign/unrecognized provenance (confirmed by the new `Provenance_We_Do_Not_Recognise_Is_Left_Untouched_By_A_Scan` test). All 471 tests in XmlGen.Tests pass. Two inline findings below, plus one that isn't inline-able since it's in a file this PR doesn't touch:

**`PremisEventManager.cs:44`** — `PremisEventManagerVirus.Patch` is now unreferenced dead code. `MetadataManager.ProcessVirusDataForFile` was its only production caller, and it now always calls `Create` instead (see the `AppendVirusEvent`/`Create` change in `MetadataManager.cs`). No functional issue, but it's ~68 lines left behind with its own subtly different `EventDateTime` fallback (`DateTime.UtcNow` vs `Create`'s `DateTime.Now`) that could confuse a future reader into thinking it's still exercised. Worth deleting in this PR or a quick follow-up.

src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
tomcrane: Fixed in `33ce950` — and following it up surfaced a worse problem in the same scheme, so thank you for pulling the thread.

The fallback now takes the highest suffix rather than the plain key, matching the structural path's `LastOrDefault`. Your reachability analysis is right and my own earlier note on this was wrong: I'd written it off as "only third-party documents, which have no suffixed IDs of ours", having missed that the structural path requires a `premis:object` on the techMD while the *write* path only requires `TechMd.Count > 0`. So we can perfectly well append ClamAV events to an amdSec whose techMD is MIX, and then only ever read the first of them.

**The related hole:** chasing the suffix logic showed the mint itself was unsafe. `UnusedDigiprovId` only checked the file's own amdSec, but `…ADM_objects_x002F_a_2` is *both* the second scan of `objects/a` and the first scan of a file genuinely named `objects/a_2`. Two files, one `xs:ID`, in a change set whose entire purpose is valid IDs. The mint now checks the whole document, and there's a test with exactly that pair of filenames.

src/DigitalPreservation/DigitalPreservation.Mets/MetadataManager.cs
tomcrane: Real, and I don't think I should decide it alone — it's a question about what the provenance record is supposed to mean rather than a bug with an obvious fix.

You're right that the overwrite model was incidentally idempotent and the append model isn't, and that Pipeline.API's SQS consumption is at-least-once, so this is reachable rather than theoretical.

What makes it awkward is that we can't distinguish the two cases from the data. `VirusScanMetadata` carries `Source`, `HasVirus`, `VirusFound` and `VirusDefinition` — **no scan timestamp**. `Create` stamps `eventDateTime` at write time, so a redelivery produces a *different* timestamp and the events aren't identical. That leaves only content-based dedup, on `(HasVirus, VirusFound, VirusDefinition)`.

And content-based dedup has a real cost: a genuine periodic re-scan with unchanged definitions would produce no event, so we'd lose the evidence that we checked again. For preservation, "re-checked on 2026-08-13, still clean" is meaningful provenance in its own right — arguably the main reason to keep a history at all. Deduping it away would undo part of what this change is for.

So I'd frame the options as:

1. **Dedup on content** — cheap, but silently discards proof of a re-check.
2. **Carry the scan's own timestamp** in `VirusScanMetadata`, set by the scanner rather than the writer, and dedup on that. Correct, and it makes the event honest about *when the scan ran* rather than when we wrote it — but it touches the pipeline's metadata contract.
3. **Idempotency at the job layer**, where the redelivery actually happens, leaving METS writing dumb.

My instinct is (3) with (2) as a follow-on, since a duplicated *event* is a symptom of a duplicated *job*, and fixing it in the METS writer means every future writer has to remember to. But it's a call for @tomcrane rather than something to slip into this PR. Happy to raise it as its own issue with this analysis so it doesn't get lost.

tomcrane:
**Dead `Patch` — deleted** in `33ce950`. Confirmed unreferenced across the solution including tests. You're right that the `DateTime.UtcNow` / `DateTime.Now` divergence was the dangerous part: sixty-eight lines that look live and disagree with the live path about something as load-bearing as an event timestamp.

**The fallback finding — fixed**, and it was the most valuable of the three, because it corrected something I had already looked at and got wrong. See the inline reply: chasing it also turned up a duplicate-`xs:ID` hole in the suffix scheme itself, which is now fixed with a test.

**Dedup — replied inline, not fixed.** It's real and reachable, but I think it's a question about what the provenance record should mean rather than a bug with an obvious fix, and the obvious fix has a cost worth stating out loud. Left for @tomcrane; happy to raise it as its own issue.

State: XmlGen 472 passing in Debug and Release, plus the rest of the solution. Both new tests were checked by reverting their fix and confirming they fail. Note this PR still has **no CI** — the workflow only triggers for pull requests targeting `main`, and this one is stacked on #218's branch; it will run once #218 merges and GitHub retargets it.

src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
```diff
@@ -525,15 +576,7 @@ private void ProcessChildStructDivs(MetsFileWrapper mets, XElement parent,
...
-                    {
-                        var matchingKey = lookupMaps.DigiprovMdMap.Keys
-                            .FirstOrDefault(k => string.Equals(k, clamavKey, StringComparison.OrdinalIgnoreCase));
-                        if (matchingKey != null)
-                        {
-                            digiprovMd = lookupMaps.DigiprovMdMap[matchingKey];
-                        }
-                    }
+                    digiprovMd = LatestByConventionalKey(lookupMaps.DigiprovMdMap, clamavKey);
                 }
```
brianmac1967, line 621: **Cross-file misattribution (new, from this commit):** `LatestByConventionalKey` resolves against the *whole document's* digiprovMD IDs, not just this file's own amdSec. If file X gets scanned twice, `UnusedDigiprovId`'s whole-document mint can give X's 2nd-scan event the literal ID that equals file Y's plain conventional key (`ADM_objects_x002F_a_2` is both X's 2nd scan and Y's own plain key) — this exact collision is documented in `UnusedDigiprovId`'s own comment in MetadataManager.cs. If Y's techMD is never recognized as `premis:object` (forcing this fallback path — the PR's own new MIX-based test demonstrates this shape) and Y itself is never scanned, this line exact-matches X's 2nd-scan element and reports X's virus status as Y's. Neither new test combines "victim file never scanned" with "victim file hits the fallback path", so this isn't covered. Narrow reproduction path, but it's the same bug class this PR exists to close, one level removed.

**Also:** the inline comment a few lines above ("Case-insensitive but EXACT match") is now stale — it only describes the prefix-collision risk (`ADM_a` vs `ADM_a2`), not the suffix-collision class this method's own newer XML-doc comment (lines 39-46) actually warns about. Worth updating so the two comments don't disagree about what the code does.

src/DigitalPreservation/DigitalPreservation.Mets/MetadataManager.cs
```diff
@@ -167,34 +167,22 @@ private Result ProcessFileFormatDataForFile(ProcessingContext ctx, WorkingFile w
...
         }
 
-        if (virusEventComplexType is null) return;
-        ctx.VirusXml = PremisEventManagerVirus.GetXmlElement(virusEventComplexType);
+        ctx.VirusXml = PremisEventManagerVirus.GetXmlElement(premisEventManagerVirus.Create(virusScan));
 
         if (ctx.AmdSec == null) return;
 
-        AddVirusXml(ctx);
+        AppendVirusEvent(ctx, fullMets.Mets);
```
brianmac1967: Idempotency gap from the prior review is still unaddressed here: `AppendVirusEvent` unconditionally appends a new digiprovMD event on every call, with no dedup check anywhere in the chain up through `ProcessAllFileMetadata`'s callers. A retried scan submission (client retry, or at-least-once SQS redelivery) will mint a duplicate event each time. Understood this is the item deliberately deferred per the PR description — just flagging so it gets an explicit decision before merge rather than merging silently.

src/DigitalPreservation/DigitalPreservation.Mets/MetadataManager.cs
```diff
@@ -285,27 +294,58 @@ public AmdSecType GetAmdSecType(FileFormatMetadata premisFile, string admId, str
...
             return;
 
-        if (ctx.AmdSec.DigiprovMd.Count != 0)
+        ctx.AmdSec.DigiprovMd.Add(new MdSecType
         {
-            ctx.AmdSec.DigiprovMd[0].MdWrap.XmlData = new MdSecTypeMdWrapXmlData { Any = { ctx.VirusXml } };
+            Id = UnusedDigiprovId(ctx, mets),
+            MdWrap = new MdSecTypeMdWrap
+            {
+                Mdtype = MdSecTypeMdWrapMdtype.PremisEvent,
```
brianmac1967, line 314: Efficiency regression (new, from this commit): `UnusedDigiprovId` now scans `mets.AmdSec.SelectMany(...DigiprovMd)` — the *whole document* — on every file that gets a virus event, instead of just that file's own amdSec. For a large multi-file deposit (thousands of files, each rescanned multiple times over its lifetime), this turns per-file ID minting into O(total digiprovMD count) work, so batch processing degrades toward O(files × total_scans) instead of O(files × own_scan_count). The whole-document scope is necessary for correctness (per the comment above, xs:ID uniqueness is document-wide) — but worth considering whether this HashSet can be built once per batch/document rather than rebuilt from scratch per file.

brianmac1967 (commented):
Re-review after the follow-up commit addressing the prior three comments:

- **Conventional-key fallback (was: resolves to first scan, not latest)** — addressed for the same-file case via `LatestByConventionalKey`, with a dedicated new test. See inline comment below for a narrower, related cross-file edge case it introduces.
- **Dead `Patch` code** — addressed cleanly, fully removed with no remaining references.
- **Idempotency/dedup on scan submission** — still open. Per the PR description this is the intentionally deferred item ("raised for a decision"), not an oversight — flagging inline as a reminder it needs an explicit call before merge.

Two new items from the follow-up commit itself, inline below.

tomcrane:
Both new items fixed in `5a8b5fb`; the third is confirmed as an explicit decision rather than an oversight.

## Cross-file misattribution — fixed, and you found the real defect

You're right, and it's the sharper of the two: making the ID document-unique fixed one collision and created a subtler one. A **trailing** counter is indistinguishable from part of a path, which is exactly what my own comment on `UnusedDigiprovId` described without my noticing it cut both ways.

The counter now sits **between** the ClamAV prefix and the identifier — `digiprovMD_ClamAV_2_ADM_objects_x002F_a` rather than `…ADM_objects_x002F_a_2`. A conventional ID always has the identifier's own prefix (`ADM_`/`TECH_`) straight after the ClamAV prefix, never a digit, so a numbered event cannot be mistaken for any file's key. The parser now **constructs** each candidate from the file's own identifier and walks upwards, instead of scanning every ID in the document for a shape.

The test is your reproduction exactly: `objects/a` scanned twice with the second infected, alongside an unscanned `objects/a_2` forced onto the fallback path by a MIX techMD. It asserts the second file reports *no* scan. Verified by reverting to a trailing counter — it fails, reporting the neighbour's infection.

## Efficiency regression — fixed by the same change

Good catch, and the fix falls out of the above rather than needing a cache. Because the ID now embeds the file's own identifier, it cannot collide with another file's, so **uniqueness within its own amdSec is sufficient** — the whole-document scan is gone. No per-batch state needed, and `UnusedDigiprovId` is back to O(this file's own scans).

## Idempotency — decided, not deferred indefinitely

Tracked as **#219 → [#221](https://github.com/digirati-co-uk/digital-preservation/issues/221)**, and @tomcrane has made the call: **job-layer idempotency first, then carry the scan's own timestamp.** The reasoning for that order is that a duplicated event is a symptom of a duplicated *job*, so solving it in the METS writer means every future writer has to remember to.

Investigating it turned up something worth knowing: `VirusScanMetadata` **does** already inherit a `Timestamp` that `MetadataReader` populates — I was wrong to say otherwise in the issue — but it holds `DateTime.UtcNow` at metadata-*read* time, so it can't distinguish a redelivery either. And `PremisEventManagerVirus.Create` ignores it entirely, writing `EventDateTime = DateTime.Now.ToLongDateString()`: write time, **date-only**, locale-formatted rather than ISO 8601. So the one field you would use to order accumulated events is neither precise enough nor machine-comparable. That is now the first, self-contained step of the timestamp work.

Also fixed: the stale "EXACT match" comment above the fallback, which described only the prefix-collision risk and not this one.

XmlGen 477 passing in Debug and Release.

**One housekeeping note:** this branch was rebased earlier — some fixes for #218's review had been committed here by mistake and have been moved to that branch. If any of your earlier inline comments show as outdated, that is why; the code they refer to is unchanged.

tomcrane:
## What changed since your re-review — #220

One commit, [`8c4829b`](https://github.com/digirati-co-uk/digital-preservation/commit/8c4829b), addressing both new items. It is small, but it **changes the on-disk ID format**, so it is worth seeing as a format decision rather than only as a bug fix.

**Cross-file misattribution.** Later scan events are now numbered *between* the ClamAV prefix and the identifier:

```
first scan   digiprovMD_ClamAV_ADM_objects_x002F_a
second scan  digiprovMD_ClamAV_2_ADM_objects_x002F_a     (was: …ADM_objects_x002F_a_2)
```

A conventional ID always has the identifier's own prefix (`ADM_`/`TECH_`) straight after the ClamAV prefix, never a digit, so a numbered event can no longer be mistaken for the plain key of a file named `objects/a_2`. The parser CONSTRUCTS each candidate from the file's own identifier and walks upwards, rather than scanning the document for a shape.

**No migration consideration**: nothing in this work is in production, and no document has ever accumulated a second scan event, so no existing METS carries the old shape.

**Efficiency.** Fixed by the same change rather than by caching — because the ID now embeds the file's own identifier it cannot collide with another file's, so uniqueness within its own amdSec is sufficient and the whole-document scan is gone.

**Idempotency.** Confirmed as an explicit decision rather than an oversight: tracked in [#221](https://github.com/digirati-co-uk/digital-preservation/issues/221), with @tomcrane's call being job-layer idempotency first, then carrying the scan's own timestamp.

### One thing worth reading in #221

Chasing it turned up that `EventDateTime` is currently `DateTime.Now.ToLongDateString()` — write time, **date-only**, and locale-formatted rather than ISO 8601. So the one field you would use to order accumulated events is neither precise enough nor machine-comparable. That is independently worth fixing and is now the first step of the timestamp work. (`VirusScanMetadata` does already inherit a populated `Timestamp`; I said otherwise in the issue and was wrong. It holds read time, so it doesn't rescue dedup either.)

### Housekeeping

This branch was **rebased twice**: fixes belonging to #218 had been committed here by mistake and have been moved. The diff is now correctly scoped to four files. If earlier inline comments render as outdated, that is why — the code they refer to is unchanged.

The PR still targets #218's branch, so **CI has not run on it**; it will once #218 merges and GitHub retargets. Run locally instead: XmlGen 477 passing in Debug and Release, plus the rest of the solution.

src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
```diff
@@ -73,6 +73,50 @@ private static MetsLookupMaps BuildLookupMaps(XDocument xMets, XElement physical
...
+    /// </remarks>
+    private static XElement? LatestByConventionalKey(
+        IReadOnlyDictionary<string, XElement> digiprovMdMap, string? identifier)
+    {
+        if (identifier == null)
+        {
+            return null;
+        }
+
+        var latest = digiprovMdMap.Keys
```
brianmac1967: This read-path fallback got *less* efficient in this commit, not more — despite the comment at lines 103-105 explicitly claiming it avoids scanning every key in the document.

Both this plain-key lookup and each numbered-occurrence lookup in the loop below (line 108-110) do a full case-insensitive `.Keys.FirstOrDefault(...)` scan over the *entire* `digiprovMdMap` (all digiprovMD IDs in the whole document), because the dictionary uses the default ordinal comparer so O(1) `TryGetValue` isn't available for a case-insensitive match. For a file that hits this fallback (non-PREMIS techMD) and has been rescanned K times, this costs O(D×(K+1)) instead of the prior single O(D) pass. Narrow path — only reachable when techMD isn't recognized as `premis:object` — but worth either building/reusing a case-insensitive lookup dictionary, or normalizing case once rather than per-comparison.

src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
```diff
@@ -73,6 +73,50 @@ private static MetsLookupMaps BuildLookupMaps(XDocument xMets, XElement physical
...
+            .FirstOrDefault(key => string.Equals(
+                key, Constants.VirusProvEventPrefix + identifier, StringComparison.OrdinalIgnoreCase));
+
+        // Later scans are numbered BETWEEN the prefix and the identifier, so each candidate
+        // still names this file and nothing else. Walk upwards rather than scanning every key
+        // in the document: an event numbered N only exists if N-1 does.
+        for (var occurrence = 2; ; occurrence++)
+        {
+            var numbered = MetadataManager.NumberedDigiprovId(occurrence, identifier);
```
brianmac1967: This introduces a new coupling: the read-only `MetsParser` now calls `MetadataManager.NumberedDigiprovId()`, a static helper on the write-side class.

The repo's `CLAUDE.md` documents a deliberate separation here: `MetsParser` reads third-party METS with unpredictable structure and the long-term goal is a fully standalone library with no `XmlGen` dependency, minimally coupled to the writer side. This call doesn't cross that literal line (no `XmlGen` type passes between them, `NumberedDigiprovId` returns a plain string), so it's not a violation of the stated rule — but it does create a compile-time dependency of the parser on a helper defined in `MetadataManager`, working against the stated goal. A future split of `MetsParser` into its own project would need to extract this shared numbering helper into a common location first. Might be worth doing that extraction now while it's fresh, rather than later.

src/DigitalPreservation/XmlGen.Tests/MetsManagerWithPremis.cs
```diff
@@ -629,6 +630,222 @@ public async Task Overwriting_File_With_FileFormatMetadata_Updates_Premis_Withou
...
+            .Metadata.OfType<VirusScanMetadata>().Should().ContainSingle()
+            .Which.VirusDefinition.Should().Be(ClamAvDefinition);
+    }
+
+    [Fact]
+    public async Task Accumulated_Scan_Ids_Do_Not_Collide_With_Another_File_S_Conventional_Id()
+    {
+        // The suffix that makes a second scan's ID unique (…_2) is also the conventional ID of
+        // the FIRST scan of a file genuinely named "…_2". xs:ID is unique per document, so the
+        // mint has to look at the whole document rather than just this file's amdSec.
```
brianmac1967: This comment is now stale: it says minting "has to look at the whole document rather than just this file's amdSec," which is exactly what commit 8c4829b reversed — `UnusedDigiprovId` is now deliberately scoped to just `ctx.AmdSec` (per its own updated comment in MetadataManager.cs), since the new numbered-prefix scheme no longer needs document-wide uniqueness checking.

Not a runtime bug — the test's assertions still hold under the new scheme, since numbered IDs never collide with plain conventional IDs regardless of scope. But the comment (and arguably the test's rationale for existing) no longer describes why the current implementation is scoped the way it is, and will mislead a future reader trying to understand the minting logic.

brianmac1967 (commented):
Re-review after commit 8c4829b. **The core cross-file misattribution bug from the last round is now correctly fixed.** Moving the counter to sit between the ClamAV prefix and the identifier (`digiprovMD_ClamAV_2_ADM_…`) is structurally sound: xs:ID/NCName can never start with a digit, so a numbered scan ID can never collide with any file's plain conventional ID. Full suite passes (481 tests, 478 passed, 3 pre-existing skips, 0 failed), including a new well-targeted regression test for this exact scenario.

Idempotency/dedup on scan submission remains open exactly as before — deliberately deferred, unchanged by this commit.

Three small, non-blocking findings from this commit below.

brianmac1967 (approved):

tomcrane:
All three fixed in [`699a6ce`](https://github.com/digirati-co-uk/digital-preservation/commit/699a6ce). Two of them were comments asserting the opposite of what the code did, which is the failure mode I keep saying matters, so thank you for reading them against the implementation rather than taking them at face value.

**The fallback got slower, not faster.** You're right, and my comment claimed the very thing it had stopped doing. The map is keyed ordinally, so a case-insensitive match can't use `TryGetValue`, and probing per candidate meant a full scan for the plain ID and another for every numbered one — worse than the single pass it replaced. It is now one pass that classifies each key (`ScanOccurrence` returns 0 for the plain form, N for a numbered one, -1 for another file's), so the comment and the code agree again.

**The parser/writer coupling.** Agreed, and taking it now while it's fresh was the right call — `NumberedVirusProvEventId` has moved to `Constants`, which is where `VirusProvEventPrefix` already lives and which both sides already depend on. That also expresses the constraint better than either location did: the writer mints this shape and the parser reads it, so it belongs in the one place they share rather than on the side that happens to have written it first. Its doc comment now says so, and points at the `MetsParser`/`MetsManager` separation in `CLAUDE.md`.

**The stale test comment.** Correct — it justified the minting scope by document-wide uniqueness, which the numbering change had already reversed. Rewritten to say what the test now actually pins: that the two families of ID are disjoint *by construction*, which is precisely why minting only needs to look at the file's own amdSec.

**CI has finally run on this PR.** It never had one before — the workflow only triggers for pull requests targeting `main`, and this was stacked on #218's branch. Worth knowing that the retarget after #218 merged did **not** trigger it either: retargeting fires `pull_request` with action `edited`, and the workflow declares no `types:`, so it defaults to `opened, synchronize, reopened`. This push supplied the `synchronize`. Green, with SonarCloud clean.

XmlGen 478 passing in Debug and Release.

PR #224, opened 2026-08-14
fix: a pipeline job runs once, and a scan is recorded once (#221)

### LINKED ISSUE #221: Virus scans are no longer idempotent: an SQS redelivery appends a duplicate provenance event
Raised out of review of #220, where virus scans changed from overwriting one `digiprovMD` in place to appending a new provenance event per scan. That is the right model — a digiprovMD records an event, and events accumulate — but it removes an idempotency the old model had by accident.

## The problem

Under the overwrite model, submitting the same scan twice was a no-op: the second write landed on the same element. Under the append model it mints a second, near-identical event.

This is reachable rather than theoretical. `Pipeline.API` consumes its jobs from SQS, which is **at-least-once** — a redelivery, or a client retry after a timeout, produces a duplicate. Nothing distinguishes "this file was rescanned and the result was the same" from "we were told about one scan twice", so an amdSec can accumulate an unbounded number of near-duplicate ClamAV events for what was really a single scan.

The *reported* result stays correct throughout — the read side takes the latest event either way — so this is about the integrity and size of the provenance record, not about virus status being wrong.

## Why the obvious fix isn't obviously right

Dedup on content — `(HasVirus, VirusFound, VirusDefinition)` — is cheap, and it's the first thing that comes to mind. But it silently discards a genuine periodic re-scan whose definitions happen not to have changed, and "re-checked on 2026-08-13, still clean" is meaningful preservation provenance in its own right. Arguably it is the main reason to keep a history at all. Deduping it away would undo part of what the accumulate change is for.

We also can't tell the cases apart from what we store. `VirusScanMetadata` carries `Source`, `HasVirus`, `VirusFound` and `VirusDefinition` — **no scan timestamp**. `PremisEventManagerVirus.Create` stamps `eventDateTime` at *write* time, so a redelivery gets a different timestamp and the two events are not identical. Content is all that is left to compare, which is exactly the comparison that can't distinguish a re-check from a retry.

## Options

1. **Dedup on content** in `MetadataManager.ProcessVirusDataForFile` / `AppendVirusEvent`. Cheapest; loses proof of a re-check.
2. **Carry the scan's own timestamp** in `VirusScanMetadata`, set by the scanner rather than the writer, and dedup on that. Correct, and it makes the event honest about when the scan *ran* rather than when we happened to write it — but it changes the pipeline's metadata contract.
3. **Idempotency at the job layer**, where the redelivery actually happens, leaving METS writing dumb.

My inclination is **(3), with (2) as a follow-on**: a duplicated event is a symptom of a duplicated job, and solving it in the METS writer means every future writer has to remember to. (2) is worth doing on its own merits regardless — an event that records write time rather than scan time is slightly dishonest provenance.

Not urgent: nothing here is in production, and the failure mode is a fatter METS file rather than a wrong answer.

tomcrane:
Fixes #221.

Scans became provenance *events* in #220, which was right, but it removed an idempotency the old overwrite model had by accident: submitting the same scan twice now appends two near-identical events. `Pipeline.API` takes its jobs from SQS, which is at-least-once, so that is reachable rather than theoretical.

Fixed at both layers, in the order agreed on the issue — the job layer first, because a duplicated event is a symptom of a duplicated *job*, and solving it only in the METS writer means every future writer has to remember to.

## 1. Starting a job is a claim, not a status report

`RunPipelineStatus` moves a job out of `waiting` with a single conditional UPDATE and returns `Conflict` if it was not waiting:

```csharp
var claimed = await dbContext.PipelineRunJobs
    .Where(job => job.Deposit == depositId
                  && job.Id == request.PipelineDeposit.Id
                  && job.Status == PipelineJobStates.Waiting)
    .ExecuteUpdateAsync(...);
```

One statement, so two consumers racing on the same job cannot both read `waiting` and both proceed — read-then-write through the change tracker would leave exactly that window open. There is a test that runs two claims concurrently against real Postgres and asserts exactly one wins.

`Pipeline.API` abandons the delivery on `Conflict`, **and only on `Conflict`**. Any other failure is treated as it was before: it is far more likely to be a transient problem reaching Preservation API than a duplicate, and refusing to run on that basis would silently strand a job that nothing will retry.

This also subsumes the existing force-complete check in `SqsPipelineQueue`, and covers the in-process queue for free, since the claim lives on the Preservation API side of the boundary.

## 2. `eventDateTime` is about the scan

It was:

```csharp
EventDateTime = DateTime.Now.ToLongDateString()
```

— the time the METS was written rather than the time of the event, to a day's precision, in whatever format the host's culture produced (`16 June 2026`). Since #220 made scans accumulate, this is the field that *orders* them, so it now carries the scan's own time as an ISO 8601 UTC instant, taken from the last-modified time of the ClamAV log that the scan itself wrote — the same trick already used for the BagIt manifest.

Two things fall out of this that are worth knowing:

- the parser does `Convert.ToDateTime(eventDatetime.Value)`, which is culture-sensitive. On the old format that would have thrown on a container running in the wrong locale; ISO 8601 parses everywhere.
- **every provenance event this platform has ever written carries a bad timestamp.** How many of those are in production is an open question — the great majority of the 100,000+ preserved Archival Groups came from EPrints migrations, whose METS `MetsManager` never wrote and never touches, so they carry no event of ours at all. What is left is the count of Archival Groups whose METS *we* wrote, which is not zero and is worth establishing before the release rather than after. Whatever that number turns out to be, the fix does not depend on it: old values are read exactly as before and never suppress a new event, so a mixed corpus needs no remediation to be correct. See [Backwards compatibility](#backwards-compatibility) below.

## 3. With an honest scan time, a scan is recorded once

The claim closes the SQS duplicate. It cannot close this one: any file put through the METS writer while the previous run's ClamAV output is still sitting in the deposit gets a second event for a scan that happened once — re-adding a file to METS from the UI does it, and a redelivered job would have done it to every file in the deposit.

So `MetadataManager` skips a scan whose `eventDateTime` it already holds. Deliberately **not** deduplicated on content, which is the fix the issue warns against: *"re-checked a month later, still clean"* is provenance in its own right. The date is what distinguishes two scans, so a genuine re-scan is appended as normal even when it found precisely what the last one found.

The date is written to the second, because its source is an S3 last-modified time and that is all S3 stores. So one second is the smallest gap at which two scans of one file stay distinguishable, and `Two_Scans_One_Second_Apart_Are_Both_Recorded` pins that. Closer than that is out of reach rather than merely unlikely: a scan is a whole Brunnhilde/ClamAV run over the deposit and the deposit lock serialises runs, so two of them cannot start and finish inside the same second. Worth being explicit that a generated disambiguator would not help here — the dedup key has to be a deterministic function of the scan, and anything that varies between two reads of the same output is precisely the bug being fixed.

The existing rules from #219 are untouched: nothing already present is read for meaning, reordered, rewritten or removed, and events from tools we do not recognise are left strictly alone. Only events carrying our own `eventType` of `virus check` are considered, and that string now lives in `Constants` next to the ID helper, since the writer both mints it and reads it back.

## Backwards compatibility

A METS file written before this change has locale-formatted, date-only `eventDateTime` values. Those are read exactly as before. They will never match a new scan's ISO 8601 value, so an old event never suppresses a new one — the failure mode of a mixed document is an extra event, not a missing one.

## Tests

15 added. Every new behaviour was verified by reverting its fix and confirming the test fails:

| Behaviour | Fails without the fix |
|---|---|
| A job not in `waiting` cannot be claimed (×4 states) | ✔ |
| Two concurrent claims, one winner | ✔ |
| `eventDateTime` is the scan's own time | ✔ |
| …in ISO 8601, independent of host culture | ✔ |
| …falling back to now, not year 1, when unknown | ✔ |
| The same scan read twice is recorded once | ✔ |

Three are guard-rails rather than tests of the change, and pass either way — they exist to catch a fix that over-reaches. Confirmed they can fail: replacing the date comparison with "any virus-check event present" (the content-style dedup) makes `A_Later_Scan_Of_The_Same_File_Is_Still_Recorded` fail.

Full suite green in **Debug and Release** (identical counts in both): XmlGen 485 (+3 pre-existing skips), Preservation.API 117, Storage.API 55, Core 68, Common.Model 8, **Pipeline.API 4**.

### Second review round

`Pipeline.API.Tests` did not exist, so the claim branch — this PR's headline behaviour — had no test. It does now, covering both halves of the rule, each verified by breaking it:

| Break | What fails |
|---|---|
| Remove the Conflict branch | `A_Delivery_That_Cannot_Claim_The_Job_Is_Abandoned` |
| Widen it to `claim.Failure` | `A_Delivery_That_Fails_To_Start_For_Any_Other_Reason_Still_Runs` (×3) |
| Restore the read-time fallback in `MetadataReader` | `A_Scan_Whose_Time_We_Do_Not_Know_Is_Left_Unstamped` (×2) |

Also from that round: a Conflict is no longer logged at Error level in two places now that it is an expected outcome; `RunPipelineStatus` no longer loads the job row before claiming it; and the PREMIS namespace is declared once in `DigitalPreservation.Mets` rather than six times.

The e2e METS comparison is unaffected — `replaceKnownDifferencesInMETSFile` already normalises `eventDateTime` away.

Docs updated in the docs repo (`mets-profiles`, commit `6b16515`): 02b now documents the ISO 8601 instant and the one-scan-one-event rule, and the stale `16 June 2026` example is gone.


src/DigitalPreservation/Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs
```diff
@@ -133,7 +133,24 @@ private async Task<Result<LogPipelineStatusResult>> UpdateAnyJobStatus(
...
-        await UpdateJobStatus(request, PipelineJobStates.Running, cancellationToken);
+        // Claiming the job IS the start: Preservation API moves it out of "waiting" only once, and a
+        // Conflict means someone else already did. SQS is at-least-once, so the same start message can
+        // be delivered more than once for one job; running it again would re-scan the deposit and
+        // append a second virus-scan provenance event for a scan that only ever happened once (#221).
+        // Any other failure is treated as it was before - it is far more likely to be a transient
+        // problem reaching Preservation API than a duplicate, and refusing to run on that basis would
+        // silently strand a job that nothing will retry.
+        var claim = await UpdateJobStatus(request, PipelineJobStates.Running, cancellationToken);
+        if (claim.ErrorCode == ErrorCodes.Conflict)
```
brianmac1967, line 155: This Conflict-vs-other-failure branch - the mechanism behind the PR description's claim that Pipeline.API "abandons the delivery on Conflict, and only on Conflict" - has zero automated test coverage. All 15 new tests referenced in the PR live in `Preservation.API.Tests` (the DB-claim atomicity) and `XmlGen.Tests` (eventDateTime/dedup); there's no `Pipeline.API.Tests` project anywhere in the repo. A regression here (e.g. swapping the condition, or a different error code being mistaken for Conflict) would ship undetected despite the PR's detailed test-coverage table implying this logic was verified.

Separately, see the general PR comment - I think the stated rationale for this branch (preserving a non-Conflict failure so "something" can retry it) doesn't correspond to a real retry path, since `SqsPipelineQueue` deletes the SQS message unconditionally before this code runs.

src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/RunPipelineStatus.cs
```diff
@@ -32,14 +32,21 @@ public async Task<Result> Handle(RunPipelineStatus request, CancellationToken ca
         var entity = await dbContext.PipelineRunJobs.SingleAsync(
```
brianmac1967: This `SingleAsync` load now runs unconditionally before the `Running`/`ClaimJob` branch below, but `entity` is never used on that path - it's a wasted DB round-trip specifically on the claim path this PR is trying to make as tight as possible. The correctness guarantee comes entirely from `ExecuteUpdateAsync`'s WHERE clause in `ClaimJob`, so this load contributes nothing there. Might be worth moving it after the `Running` check, so the claim path stays a single round-trip.

brianmac1967 (commented):
Reviewed against the detailed write-up in the PR description, cross-checking each specific claim against the actual diff rather than taking the prose at face value. The core correctness mechanisms check out: the atomic `ExecuteUpdateAsync` claim (verified with a real concurrent-Postgres test that passes), the ISO 8601 write format, the exact-timestamp dedup logic in `MetadataManager`, the untouched #219 rules, and the backwards-compatibility argument all match what's described. Full suite green in Debug and Release, independently re-run.

Two things worth a second look before merge - both in files this PR doesn't touch, so posted here rather than inline:

**1. The "abandons on Conflict, and only on Conflict" framing doesn't match what the code does.** The description argues this distinction matters because refusing to run on a non-Conflict failure "would silently strand a job that nothing will retry" - implying non-Conflict failures deliberately preserve the SQS message for a retry. But `SqsPipelineQueue.cs` (`DequeueRequest`, ~line 80) deletes the SQS message **unconditionally**, immediately on receipt, before `ExecutePipelineJob.Handle` - and therefore the Conflict check - ever runs. So the message is already gone regardless of outcome; there's no retry path downstream that this distinction is protecting. The Conflict-vs-other-failure branch itself may still be the right call, but the stated *reason* for it describes a safety mechanism that isn't actually there in this code. Worth either correcting the description's reasoning, or confirming I've misread `SqsPipelineQueue` and there's a re-queue path elsewhere I didn't find.

**2. Possible real bug: a missing UTC normalization on the new timestamp's source.** `Storage.cs`'s `GetStream()` (~line 581) returns S3's raw `LastModified` without the `.ToUniversalTime()` call that the *other two* `LastModified` reads in that same file apply - both of which carry an explicit comment referencing AWS SDK issue #1885 and `DateTime.Kind` ambiguity. The new `eventDateTime` write path (`MetadataReader.cs` -> `PremisEventManager.EventDateTimeFor`) consumes exactly this unnormalized value and calls `.ToUniversalTime()` once, itself, at write time. If the SDK's `LastModified.Kind` isn't already `Utc` on this particular call path - the exact ambiguity the sibling code was written to guard against - the "ISO 8601 UTC instant" this PR is built around could be silently off by the host's local UTC offset. `MetadataReaderVirusTimestampTests.cs` wouldn't catch this either way, since its fake always constructs `DateTimeKind.Utc` directly rather than exercising the real SDK's behaviour on this path.

Minor, not worth an action: the description says "XmlGen 484"; an independent re-run shows 485 passed (488 total incl. 3 pre-existing skips) - suggests the coverage numbers weren't re-verified before being written down. Doesn't affect the substance of the PR.

tomcrane:
Thanks — both worth chasing down. One of them I think reads the comment backwards, one is a real inconsistency I've fixed, and the test count doesn't reproduce.

**1. The retry reasoning — I think this is an inversion.**

The claim isn't that a non-Conflict failure preserves the message for a retry. It's the opposite: *nothing* retries it, therefore we must not abandon on a transient error. "Refusing to run on that basis would silently strand a job that nothing will retry" is an argument *for* proceeding, and your `SqsPipelineQueue` finding is exactly what makes it true — `DequeueRequest` deletes the message at line 80, on receipt, before `Handle` runs. So there is no retry path, and a job abandoned on a transient error is abandoned for good.

That said, if the sentence can be read the other way once it can be read that way again, so I've anchored it in the code that makes it true rather than leaving it implicit (`830ba5d`):

```
// Any other failure is treated as it was before - it is far more likely to be a transient
// problem reaching Preservation API than a duplicate, and there is no retry to fall back on:
// SqsPipelineQueue.DequeueRequest deletes the message the moment it reads it, before we get
// here, so a job abandoned on a transient error is abandoned for good. Bailing out on
// anything short of a definite duplicate would silently strand it.
```

**2. The UTC normalization — right about the inconsistency, but it isn't a bug.**

You're right that `GetStream` (line 581) was the only one of the three `LastModified` reads in `Storage.cs` without `.ToUniversalTime()` and the aws-sdk-net#1885 comment. But the written value is unaffected, because `.ToUniversalTime()` is still applied exactly once on this path — just one hop later, in `EventDateTimeFor`:

```
MetadataReader:        virusScanTimestamp = brunnhildeAVResult.Value.LastModified;   // passed through unchanged
PremisEventManager:    XmlConvert.ToString(scanned.ToUniversalTime(), ...Utc)
```

Nothing between the two touches the value. So whatever the SDK's `Kind` is on this call, this path applies the identical conversion the sibling reads apply: if it comes back `Local` both convert correctly, and if it comes back `Unspecified`-but-really-UTC both are wrong together. There's no differential for the offset to appear in.

Still worth fixing as consistency — the odd one out costs every reader the double-take it cost you, and normalising at the read is the better place for it. Done in `830ba5d`. `MetadataReader` is the only consumer of this value in the codebase (`MediaController`, `ContentController` and `FedoraClient` all take the stream and discard the timestamp), and both its uses want a UTC instant: the virus scan time, and `bagitTimestamp`, which is initialised from `DateTime.UtcNow` and so was already being mixed with UTC. `.ToUniversalTime()` is a no-op on a `Utc`-kind value, so the call in `EventDateTimeFor` stays as-is — `Timestamp` has a non-S3 fallback path.

Full suite still green in Release after both changes.

**3. The test count — 484 is what I get, in both configurations.**

Re-ran on the branch head just now: XmlGen `Passed: 484, Skipped: 3, Total: 487` in Release, and identical in Debug. The branch is level with `main` (0 commits behind), so it isn't a stale-base difference. I can't account for 485/488 — if it reproduces for you, the extra test's name would be interesting, because a test that appears for you and not for me is a more interesting problem than the count.

src/DigitalPreservation/DigitalPreservation.Mets/PremisEventManager.cs
```diff
@@ -41,6 +41,28 @@ public EventComplexType Create(VirusScanMetadata virusScanMetadata)
...
+    private static string EventDateTimeFor(VirusScanMetadata virusScanMetadata)
+    {
+        // Metadata.Timestamp carries the scan's own time (the ClamAV log's last-modified time - see
+        // MetadataReader). Fall back to now only when we genuinely have nothing better, rather than
+        // writing a default DateTime and claiming the scan happened in the year 1.
+        var scanned = virusScanMetadata.Timestamp == default
+            ? DateTime.UtcNow
+            : virusScanMetadata.Timestamp;
+
+        return XmlConvert.ToString(scanned.ToUniversalTime(), XmlDateTimeSerializationMode.Utc);
```
brianmac1967, line 63: The dedup this PR relies on keys on `eventDateTime`'s string form, ultimately sourced from S3's `LastModified` header - which only has 1-second resolution. Two genuinely distinct scans of the same file whose ClamAV log uploads happen to land in the same UTC second would produce an identical `eventDateTime`.

`MetadataManager.AlreadyRecorded` would then treat the second, real scan as already-recorded and silently drop its provenance event - contradicting the PR description's premise that "two scans of one file cannot share an instant." Narrow trigger (needs two scans within the same second), but a real gap in the design rather than an edge case the tests exercise - none of the new tests use two timestamps closer together than whole seconds apart. Worth a maintainer decision: accept this as acceptable residual risk given how narrow it is, or add a disambiguator (e.g. include a counter/nonce alongside the timestamp) so a same-second collision can't silently eat a real event.

src/DigitalPreservation/DigitalPreservation.Workspace/MetadataReader.cs
```diff
@@ -168,8 +168,19 @@ private async Task FindMetadata()
 
         var brunnhildeFiles = brunnhildeSiegfriedOutput is { Files.Count: > 0 } ? brunnhildeSiegfriedOutput.Files : [];
 
+        // When the scan actually ran, not when we happened to read its output. A virus scan becomes a
+        // PREMIS event, and an event's date is the date of the event - so it has to survive being read
+        // again later, or a redelivered pipeline job produces two events that look like two scans
+        // (issue #221). The ClamAV log is written by the scan itself, so its last-modified time is the
+        // closest thing to a scan time we have without changing Brunnhilde's output.
+        var virusScanTimestamp = timestamp;
```
brianmac1967: If `brunnhildeAVResult` succeeds but `LastModified` comes back as `default` (a narrow but not impossible SDK/backend edge case), `virusScanTimestamp` silently falls back to `timestamp` (read-time "now") rather than a real scan time - and critically, this fallback value isn't `default(DateTime)`, so it doesn't trip `MetadataManager.AlreadyRecorded`'s explicit `Timestamp == default` guard.

That means every redelivery of the same job under this condition would mint a fresh "now" `eventDateTime` each time, so the date-based dedup would never match and the exact duplicate-event bug this PR sets out to fix could reappear - silently, with nothing failing loudly. `MetadataReaderVirusTimestampTests`'s fakes only ever construct `DateTimeKind.Utc` timestamps directly and don't exercise a default-`LastModified`-but-`Success:true` path, so this isn't currently covered.

src/DigitalPreservation/DigitalPreservation.Mets/Constants.cs
```diff
@@ -29,6 +29,16 @@ public static class Constants
...
 
+    /// <summary>
+    /// The premis:eventType we write for a virus scan, and the only value by which we recognise one
+    /// of our own scan events again. Here rather than on either side for the same reason as
+    /// <see cref="NumberedVirusProvEventId"/>: the writer mints it and the writer later reads it
+    /// back to tell whether a scan is already recorded, and the two must not drift.
+    /// </summary>
+    public const string VirusCheckEventType = "virus check";
+
+    public const string PremisNamespace = "http://www.loc.gov/premis/v3";
```
brianmac1967, line 45: Minor/cosmetic: this duplicates a pre-existing private `const string PremisNamespace` already in scope in `MetadataManager.cs` (~line 297), and there's a third near-identical pattern via `XNames.premis.NamespaceName` in `MetsCache.ExtractPremisOriginalName`. Not a bug, but a future PREMIS namespace/version change now has multiple places to remember to update - including two in the very file where the duplication is easiest to have avoided. Might be worth having `MetadataManager` reference this new constant instead of keeping its own copy, while touching this area.

src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/RunPipelineStatus.cs
brianmac1967: Still unresolved from the last review round: this `SingleAsync` load still runs unconditionally before the `Running`/`ClaimJob` branch below, and `entity` is still unused on that path - `ClaimJob` runs its own independent `ExecuteUpdateAsync`. Still a wasted DB round-trip on the claim path this PR is hardening, and `SingleAsync` will still throw on a missing row here where `ClaimJob`'s own query would gracefully report `Conflict` instead.

brianmac1967 (commented):
Re-reviewed after the follow-up commit (`830ba5de`, "normalise S3 LastModified at the read, not at the write"). Both substantive findings from the last round are properly resolved:

- **The SQS-delivery framing** is now correct - the comment above `Handle` accurately describes that `SqsPipelineQueue` deletes the message unconditionally before this code runs, verified again against the actual delete call.
- **The missing `.ToUniversalTime()` on `Storage.cs`'s `GetStream()`** is fixed properly, not cosmetically - now matches the sibling `LastModified` reads in the same file, and I traced every downstream consumer of `GetStream` to confirm nothing else regressed.

Two items from the original review remain open and untouched by this commit: the Conflict-bailout branch in `ExecutePipelineJob.cs` still has no test coverage (no `Pipeline.API.Tests` project exists in the repo), and the wasted `SingleAsync` load on the claim path in `RunPipelineStatus.cs` is still there (comment below). The "XmlGen 484" vs actual test-count discrepancy in the description is unchanged too - still a harmless nit, not a code issue.

One more thing worth flagging that isn't inline-able, since it's in code this PR doesn't touch: **`ExecutePipelineJob.cs`'s `UpdateAnyJobStatus` (~line 126-129)** logs at **Error** level on any non-success `LogPipelineRunStatus` call, including the Conflict outcome this PR now treats as an expected, frequent, correctly-handled result (a legitimate duplicate SQS delivery). On top of the deliberate `Warning` already logged in `Handle` for this case, every routine duplicate delivery - which the PR description itself says is "reachable rather than theoretical" - will also emit an Error-level log from this pre-existing generic branch and another from `PipelineJobRunner`. If Error-level logs feed alerting/dashboards, this means correctly-handled, by-design duplicate deliveries will page on-call or inflate error-rate metrics. Worth a follow-up to special-case Conflict down to Warning/Information in `UpdateAnyJobStatus` too, now that it's a normal outcome rather than a failure.

brianmac1967 (approved):

src/DigitalPreservation/Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs
tomcrane: Fixed in `0227501`. There is now a `Pipeline.API.Tests` project, and this branch is what it exists for.

Both halves of the rule are covered, because the second half is the one that is easy to lose to a tidy-up — `if (claim.Failure)` reads like a simplification and would silently drop jobs on any transient Preservation API blip, with nothing to retry them. Each verified by breaking it:

| Break | What fails |
|---|---|
| Remove the Conflict branch entirely | `A_Delivery_That_Cannot_Claim_The_Job_Is_Abandoned` |
| Widen it to `claim.Failure` | `A_Delivery_That_Fails_To_Start_For_Any_Other_Reason_Still_Runs` (×3) |

The abandon test asserts `GetDeposit` is never called, so it pins the thing that actually matters: abandoning early enough that Brunnhilde is never reached and no second event can be written.

src/DigitalPreservation/DigitalPreservation.Mets/PremisEventManager.cs
tomcrane: Real, and I'm leaving it — but you're right that the description overstated it, and it now says what is actually true.

`eventDateTime` is written to the second because its source is an S3 last-modified time and that is all S3 stores. So one second is the resolution, and I've pinned it with `Two_Scans_One_Second_Apart_Are_Both_Recorded` rather than leaving the boundary implicit.

Two reasons for accepting rather than disambiguating:

**The trigger is out of reach, not merely narrow.** A scan is a whole Brunnhilde/ClamAV run over the deposit's `objects/` folder, and the deposit lock serialises runs against one deposit. Two scans cannot start *and* finish inside the same second.

**A generated disambiguator would reintroduce the bug it is meant to guard.** The dedup key has to be a deterministic function of the scan, because the case being defended against is *the same output read twice*. Anything that varies between two reads — a counter, a GUID — makes every redelivery unique again, which is #221 exactly. A disambiguator would have to be derived from the log's content, and at that point it only helps when the content differs, which for two scans one second apart it essentially always would anyway. So it would buy nothing for the case it is aimed at while risking the case that is real.

The comment on the new test records that reasoning, including what would have to change for the conclusion to change.

*(Edited to reword one term — it reads very differently in British English than intended.)*

src/DigitalPreservation/DigitalPreservation.Workspace/MetadataReader.cs
tomcrane: This was the best finding of the round — fixed in `0227501`.

You're right, and the reason it's nasty is that the fallback value *passes* the `Timestamp == default` guard. It looks like a real scan time, so `AlreadyRecorded` runs, compares, and never matches — because it's a different "now" every read. Dedup that appears to be working and isn't.

Now it returns `default` instead, which routes to the explicit "no scan time" branch that prefers a duplicate event to a lost scan. Same outcome in the end, but as a decision taken in the open.

One thing that came out of writing the test, which I'd missed: `!= default` isn't a safe check any more, because of the `.ToUniversalTime()` I added in `830ba5d`. On a host west of UTC that turns an unset `DateTime` into `0001-01-01T05:00` — meaningless as a date, but no longer equal to `default`, so it would have sailed through the guard and written a scan dated to the year 1. It tests for plausibility now instead. Both shapes are covered:

```
[InlineData(0L)]                          // unset
[InlineData(5L * TimeSpan.TicksPerHour)]  // unset, after ToUniversalTime shifted it
```

Both fail against the old code, confirmed by reverting.

src/DigitalPreservation/DigitalPreservation.Mets/Constants.cs
tomcrane: Fair, and it was worse than you flagged — the duplication was mine, added in this PR, in the file I'd just touched. Fixed in `0227501`, and I took the rest while I was there:

- `MetadataManager`'s private `const` deleted, both uses now `Constants.PremisNamespace`
- the four `namespaces.Add("premis", …)` literals in `PremisEventManager`, `PremisManager`, `PremisManagerExif` and `StorageHelpers`
- `XNames.premis` now derives its `XNamespace` from the same constant

`DigitalPreservation.Mets` now spells the namespace out exactly once, in `Constants.cs`, with a comment saying that's the point. The generated `XmlGen.Premis.V3.cs` attributes are left alone — that file is generated from the schema and shouldn't be hand-edited.

src/DigitalPreservation/Preservation.API/Features/Deposits/Requests/RunPipelineStatus.cs
tomcrane: Fixed properly this time in `0227501` — sorry, I acknowledged it in the last round and then didn't act on it.

The load has moved below the `Running` branch, so the claim path is `SingleOrDefaultAsync` (the deposit) plus the `ExecuteUpdateAsync` itself, and nothing else.

Your `SingleAsync`-throws point is the better half of this and I'd underrated it: a delivery for a job that no longer exists should be abandoned, not turned into a 500. It now gets the same clean `Conflict` as any other lost claim.

The one thing the old load bought was a precise log line — `claimed == 0` on its own can't tell "not waiting" from "not there". The rare branch does a cheap `AnyAsync` to say which, so the diagnostic is actually better than before, and it costs a round trip only on the path that was already exceptional.

tomcrane:
All five addressed in `0227501` — replies on each thread. Summary, and one where I've pushed back rather than changed the code.

**The `MetadataReader` default-`LastModified` fallback was the find of the round.** Substituting read-time "now" is worse than substituting nothing, exactly as you say, and the reason is that the fallback value *passes* the `Timestamp == default` guard — so `AlreadyRecorded` runs, compares, and never matches, because it's a different "now" every read. Dedup that looks like it's working and isn't, which is the failure mode this PR exists to remove. It returns `default` now.

Writing the test surfaced a second problem I'd introduced myself: after `830ba5d`, `!= default` is no longer a safe test for "unset", because `.ToUniversalTime()` shifts an unset `DateTime` to `0001-01-01T05:00` on a host west of UTC — meaningless as a date, but no longer equal to `default`, so it would have written a scan dated to the year 1. Both shapes are covered now, and both fail against the old code.

**`Pipeline.API.Tests` now exists.** You were right that the PR's headline behaviour was the one thing with no test behind it, and that the coverage table implied otherwise. Both halves of "abandon on Conflict, and only on Conflict" are covered, each verified by breaking it in that direction.

**The `SingleAsync` load** — fixed properly this time; I acknowledged it last round and then didn't act on it, which is on me.

**Error-level logging on a routine Conflict** — a good catch that follows directly from this PR changing what Conflict *means*. Fixed at both sites you identified.

**The same-second collision I've left alone, deliberately, and corrected the description instead.** You're right that "two scans of one file cannot share an instant" was an overstatement; the resolution is one second, because the source is an S3 last-modified time. That's now pinned by a test rather than left implicit. Two reasons for not disambiguating:

- The trigger is unreachable rather than narrow. A scan is a full Brunnhilde/ClamAV run over the deposit, and the deposit lock serialises runs against one deposit — two cannot start *and* finish inside one second.
- **A generated disambiguator would reintroduce the bug.** The dedup key has to be a deterministic function of the scan, because the case being defended against is the same output read twice. A counter or GUID makes every redelivery unique again, which is #221 verbatim. Any workable disambiguator would have to derive from the log's content — and then it only helps when the content differs, which for two scans a second apart it would anyway.

Full suite green in Debug and Release: XmlGen 485, Preservation.API 117, Storage.API 55, Core 68, Common.Model 8, Pipeline.API 4.

On the count discrepancy: XmlGen is now 485/488 on this branch, which happens to be the numbers you reported for the previous commit. I still can't reproduce 485 on `830ba5d`, so I don't think that resolves it — flagging the coincidence rather than claiming it as an explanation.

PR #225, opened 2026-08-14
test: AG-derived METS navigability, in every build configuration (#222)

### LINKED ISSUE #222: The gating CI job runs Release, so [Conditional("DEBUG")] assertions never run in it
Raised from the cumulative #188 audit (finding T4). **The diff is one line; the decision behind it isn't**, which is why it's here rather than in a PR.

## The situation

`.github/workflows/build.yml` runs the gating test job in Release:

```yaml
env:
  BUILD_CONFIG: "Release"
...
- run: dotnet test $SOLUTION --filter 'Category!=Manual' --configuration $BUILD_CONFIG --no-restore --no-build
```

and `build-push-images` has `needs: test-dotnet`. Debug is only built in the non-gating `sonar-cloud-analysis` job.

That means every `[Conditional("DEBUG")]` assertion **compiles out of the job that decides whether we ship**.

## Measured, not assumed

Breaking the Archival-Group export so directory divs get no `premis:originalName` (`MetsFromArchivalGroup.AddResourceToMets`, one line):

| Configuration | Result |
|---|---|
| Debug | **8 failed**, 470 passed |
| Release | **1 failed**, 477 passed |

`MetsFromArchivalGroup.AssertNavigable` is doing seven of those eight. It is load-bearing for a production code path — building METS from an existing Archival Group — and in the job we actually gate on, it does not exist.

Worth noting the converse, so the scope is clear: `MetsManager.AssertCacheConsistent` is **not** in this position. Breaking cache maintenance fails the same six tests in both configurations, because `PhysicalDivsCacheTests` covers it with ordinary assertions. So this is about one assertion, not the whole pattern.

## Options

1. **Add a second test job running Debug.** Cheapest — a few lines of YAML and roughly 90 seconds of CI. Keeps Release as the gate (it is what ships) and gets the assertions exercised too.
2. **Replace the load-bearing assertions with ordinary tests.** More work, but it removes the dependence on a debug-only mechanism for coverage. Arguably the right shape: an assertion is a self-check, and coverage should not be a side effect of one.
3. **Do nothing**, accepting that a class of regression in the AG-export path is caught only by whoever runs the suite locally in Debug.

My inclination is **1 now and 2 for `AssertNavigable` specifically**, since (1) is nearly free and closes the gap immediately, while (2) is what stops us relying on this again. But it is a question about how the project gets its assurance, so it wants a team view rather than a quiet PR.

## Context

Not urgent and nothing is broken: this is about what future changes are protected by, not a defect in current behaviour. It came up because the #188 work touched exactly this path, and the audit's test-suite pass measured the gap rather than inferring it.

tomcrane:
Closes #222.

`MetsFromArchivalGroup.AssertNavigable` is `[Conditional("DEBUG")]`, so the check that an exported Archival Group produces a navigable structMap only ever ran in a Debug build. This replaces that accidental coverage with an ordinary test, which is the resolution proposed on the issue.

An earlier draft of this description said the assertion "never ran where it mattered" because "the CI job that gates merges builds Release". Prompted by review, I checked the actual configuration rather than assuming it, and both halves of that were wrong:

- **`sonar-cloud-analysis` builds Debug.** It runs bare `dotnet build` / `dotnet test` with no `--configuration`, so the assertion *has* been executing on every PR. Only `test-dotnet` is Release.
- **No CI check gates merges at all.** The `main` ruleset (`gh api repos/.../rules/branches/main`) carries exactly three rules — `deletion`, `non_fast_forward`, and `pull_request` with `required_approving_review_count: 0`. There is no `required_status_checks` rule, so neither job blocks a merge.

So the assertion was running, in a job that blocks nothing, where a red test sits behind a green quality-gate badge. That is a weaker gap than "never ran" and a better argument for this change than the one I originally made: an ordinary test in `test-dotnet` is coverage you can actually see fail.

## What the measurement actually showed

I broke navigability *only* — dropping the `premis:originalName` that directory divs resolve through, which leaves every ID valid and unique — and counted:

| | Debug | Release |
|---|---|---|
| XmlGen.Tests failures | **8** | **1** |

The single Release failure was `MetsManagerTests.Can_Delete_Directories_From_Mets`, an incidental catch from a delete test that happens to lean on the AG-built structure.

The interesting part is what *passed*. `MetsIdIntegrityTests.A_Mets_Built_From_An_Archival_Group_Has_Valid_Unique_Resolvable_Ids` — the one test named for this construction path — passed in Release with the structMap unnavigable, because **IDs stay valid, unique and resolvable when paths stop resolving.** The #188 merge gate says nothing at all about navigability. The gap was real, not theoretical arithmetic.

## The test

`PhysicalDivsCacheTests.A_Mets_Built_From_An_Archival_Group_Is_Fully_Navigable_By_Path` builds a METS from the existing `Samples/archivalGroup.json` fixture, loads it back the way any reader would, and asserts:

- `PathDiagnostics` is empty — the same invariant `AssertNavigable` checks;
- every container and binary in the Archival Group is reachable by its deposit-relative path;
- the nested case (`objects/folder-b/folder-bb/…`) explicitly, because that is where a path-building bug shows first.

Expected paths are **derived from the Archival Group** rather than listed, so the test cannot drift from the fixture and states the contract independently of the code that builds the structMap.

One exclusion, which the first run of the test established rather than something I assumed: the METS file itself does not appear in its own structMap, so it is filtered out via `MetsUtils.IsMetsFile`.

**Verified by re-applying the break: the new test fails in Release.** Full suite green in Debug and Release (XmlGen 479, Preservation.API 106, Storage.API 55, Core 68, Common.Model 8).

## What this does not change

The assertion stays. With the coverage in a real test it goes back to being what it was meant to be — a development-time self-check — rather than the only thing standing behind this path. `MetsManager.AssertCacheConsistent` is untouched and should stay conditional: it rebuilds the whole cache on every navigation, so making it unconditional would recreate the quadratic regression R1 removed, and its invariant is covered plainly by the other tests in this file.

I am **not** proposing a Debug CI job. Tom's point on the issue is that `MetsFromArchivalGroup` only runs when an Archival Group with no METS file is exported, which is not expected in production — it was useful during development. That is the same fact that defused the duplicate-ID finding during the #188 audit: no production item was ever made this way. So the only place where a debug assertion carried real coverage is a path we do not expect to run in production, and with this test the residual risk is that the assertions themselves rot unnoticed. That does not seem worth a second CI job.

Sized to go into the release with the rest of the #188 work.

### DIFF
--- a/src/DigitalPreservation/XmlGen.Tests/PhysicalDivsCacheTests.cs
+++ b/src/DigitalPreservation/XmlGen.Tests/PhysicalDivsCacheTests.cs
@@ -1,12 +1,15 @@
+using System.Text.Json;
 using DigitalPreservation.Common.Model;
 using DigitalPreservation.Common.Model.Transit;
 using DigitalPreservation.Common.Model.Transit.Extensions.Metadata;
 using DigitalPreservation.Mets;
 using DigitalPreservation.Mets.StorageImpl;
+using DigitalPreservation.Utils;
 using DigitalPreservation.XmlGen.Mets;
 using FluentAssertions;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Storage.Repository.Common.Mets;
 
 namespace XmlGen.Tests;
 
@@ -21,6 +24,7 @@ public class PhysicalDivsCacheTests
 {
     private readonly MetsManager metsManager;
     private readonly FileSystemMetsStorage metsStorage;
+    private readonly MetsFromArchivalGroup metsFromArchivalGroup;
 
     private const string TestDigest = "eb634d64ce8e6be5195174ceaef9ac9e19c37119f3b31618630aa633ccdbf68f";
 
@@ -40,6 +44,7 @@ public class PhysicalDivsCacheTests
         var premisEventManager = new PremisEventManagerVirus();
         var metadataManager = new MetadataManager(premisManager, premisManagerExif, premisEventManager);
         metsManager = new MetsManager(parser, metsStorage, metadataManager);
+        metsFromArchivalGroup = new MetsFromArchivalGroup(metsManager, parser, metadataManager);
     }
 
     // -----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/DigitalPreservation/XmlGen.Tests/PhysicalDivsCacheTests.cs
```diff
@@ -732,4 +737,82 @@ public async Task Legacy_Mets_With_Spaces_In_Ids_Is_Cached_And_Editable_By_Path(
...
+    private static List<string> DepositPathsIn(ArchivalGroup archivalGroup)
+    {
+        var root = archivalGroup.Id!.LocalPath;
+        var paths = new List<string>();
+
+        void Collect(Container container)
+        {
+            foreach (var binary in container.Binaries)
+            {
+                if (MetsUtils.IsMetsFile(binary.Name ?? "")) continue;
```
brianmac1967: This exclusion predicate doesn't quite mirror the production one it's meant to stand in for. Here it's `MetsUtils.IsMetsFile(binary.Name ?? "")` - a loose match on just the bare name, default `mustBeStandardName: false`. Production (`MetsFromArchivalGroup.AddBinariesToMets`) uses `IsMetsFile(localPath, true)` - an exact match on the full relative path.

They only agree today because the fixture's METS file happens to sit at the deposit root, named exactly `mets.xml`. Two ways this could silently diverge later: a fixture update adds a non-METS binary named e.g. `old-mets-backup.xml` - this test would drop it from `expected` without failing, weakening coverage quietly; or the METS binary moves into a subdirectory - production's exact-path match would stop excluding it, but this test's name-based check still would, so the derived path set would no longer match what `MetsFromArchivalGroup` actually produces. Might be worth matching production's exact predicate here (`IsMetsFile(Relative(binary.Id!), true)` or similar) so the test can't quietly drift from the code it's meant to describe.

brianmac1967 (commented):
Verified every claim in the PR description and the issue-#222 thread it restates against the actual diff - including two full break/rebuild/test cycles of my own in both Debug and Release. All of it holds up: the diff really is just the one new test file (83 lines, zero production code touched), the new test genuinely derives expected paths from the fixture rather than hardcoding them, breaking navigability makes it fail in Release exactly as claimed, `AssertCacheConsistent` is confirmed untouched and still `[Conditional("DEBUG")]`, and the full-suite counts (XmlGen 479, Preservation.API 106, Storage.API 55, Core 68, Common.Model 8) matched exactly on an independent re-run in both configurations.

One inline finding below, plus one that can't be inline since it's in a file this PR doesn't touch:

**`.github/workflows/build.yml`** - the core premise ("the CI job that gates merges builds Release, so this assertion never ran where it mattered") only accounts for the `test-dotnet` job. `sonar-cloud-analysis`, in the same workflow, runs `dotnet build` / `dotnet-coverage collect "dotnet test"` with no `--configuration` flag - i.e. Debug by default. Whether that job is a *required* branch-protection check isn't visible from the repo, so this doesn't invalidate the fix (replacing a conditional assertion with an ordinary test is strictly better regardless of whether the gap was fully or partially real) - but the framing "never runs in a job that gates merges" may be only partially accurate, worth a quick confirmation of branch-protection config rather than taken at face value.

brianmac1967 (approved):

src/DigitalPreservation/XmlGen.Tests/PhysicalDivsCacheTests.cs
tomcrane: Good catch, and fixed in `a3b886e` — this is now `IsMetsFile(Relative(binary.Id!), true)`, the same expression `AddBinariesToMets` uses, against the same string.

Your second divergence is the one that would have bitten hardest, and it is worth spelling out why: both failure modes are *silent*. The assertion is `Contain(expected)` — a superset check — so an entry wrongly dropped from `expected` never fails anything, it just quietly stops being tested. A derived expectation is supposed to be immune to exactly that, and mine had a hand-written predicate sitting in the middle of it doing the deriving. Comment added recording why the loose form is not good enough, so the next person does not simplify it back.

The METS file is still excluded, which the test passing proves: if the predicate had stopped matching, `mets.xml` would now be in `expected` and absent from the cache.

tomcrane:
Thanks — the inline finding was right and is fixed in `a3b886e` (replied on the thread). The workflow point turned out to be worth more than a footnote, so I went and read the actual configuration instead of assuming it. Both halves of my framing were wrong, and not in the direction I expected.

**You're right that `sonar-cloud-analysis` builds Debug.** Bare `dotnet build` / `dotnet test`, no `--configuration`. So `AssertNavigable` *has* been executing on every PR all along. It wasn't invisible.

**But no CI check gates merges at all.** `main` isn't protected in the classic sense — it's a ruleset, and it has exactly three rules:

```
$ gh api repos/digirati-co-uk/digital-preservation/rules/branches/main
deletion
non_fast_forward
pull_request  { required_approving_review_count: 0, ... }
```

No `required_status_checks` entry. So neither `test-dotnet` nor `sonar-cloud-analysis` blocks anything; the only requirement is that a change arrives as a PR, with zero required approvals.

So the accurate version is: the assertion was running, in a job that blocks nothing, where a red test sits behind a big green quality-gate badge that most people read as "Sonar is happy". That's a weaker gap than "never ran" — and a better argument for this change than the one I made, since an ordinary test in `test-dotnet` is coverage you can actually see fail. Description corrected, with the evidence in it.

**One thing that fell out of checking, which I think matters more than either of the above.** The Sonar job also runs with no `--filter 'Category!=Manual'`, so **it runs the Manual tests too.** Numerically, on #227:

| Job | Config | XmlGen total |
|---|---|---|
| `test-dotnet` | Release, Manual excluded | 481 |
| `sonar-cloud-analysis` | Debug, Manual included | 483 |

`#227`'s branch has exactly two Manual tests in XmlGen (`MetsManagerPathFixtureTests`, `SurveyMetsCorpus`). The +2 is them.

`CLAUDE.md` says "CI skips 'Manual' category", which is true of `test-dotnet` and false of the other job — and I've been writing Manual tests against that assumption, including a corpus survey that reads an env-var path and a 1,600-file timing loop on #229. The survey passes there (it no-ops without `METS_SURVEY_PATH`, which is the behaviour I wanted, though I'd have preferred to have designed it for that reason rather than got it by luck). Nothing is broken today, so I'm not folding a fix into this PR, but "Manual means it won't run in CI" is not a safe thing to keep believing. I'll raise it separately.

tomcrane:
Raised as #230, with the measurement and the branch-ruleset finding.

Issue #223, opened 2026-08-13
Decide METS editability by conformance to a profile, not by the mets:agent name


tomcrane:
Leeds wants to edit deposits whose METS we did not create. We currently cannot, and the reason is a single equality check.

This idea has been discussed and written up, but only inside a document about something else — [`docs/issues/188/issue-188-reassessment.md` § "New consideration: conformance-based editability"](https://github.com/digirati-co-uk/digital-preservation/blob/main/docs/issues/188/issue-188-reassessment.md#L87) (line 87 onwards). This issue exists to give it a home, because several findings from other work have been parked against a "conformance-checker backlog" that has no ticket to hang on.

## Today's rule

A METS file is editable if and only if its `mets:agent` name is exactly ours:

```csharp
mets.Editable = mets.Agent == Constants.MetsCreatorAgent;   // MetsParser.cs:245, :261
```

The same gate is enforced at the storage layer, which refuses to load a `FullMets` at all — `FileSystemMetsStorage.cs:64` and `S3MetsStorage.cs:107`.

Two things are wrong with it:

1. **It excludes everything we did not write.** Third-party METS uses arbitrary ID schemes — Goobi `PHYS_0001`, Archivematica `file-<uuid>`, EPrints `eprint_10315_370441` — so no navigation that assumes our `PHYS_` + path convention could ever work on them anyway. That was one of the motivations for the #188 Step 1 path cache.
2. **It is a weak guarantee even for our own files.** "The agent string matches" is not the same claim as "this document has the structure our editing code relies on". A file we wrote, then someone hand-edited, still passes.

## Proposed rule

**A METS file is editable if it conforms to a documented profile** — [`02b-METS-Written-by-the-Platform.md`](https://github.com/digirati-co-uk/digital-preservation-docs/blob/mets-profiles/documentation/02b-METS-Written-by-the-Platform.md) in the docs repo, generalised to admit files we did not write.

To first order, editable ≈ **the Step 1 path cache builds completely and unambiguously**, plus the structural invariants `MetsManager` assumes:

- a physical structMap whose Directory divs resolve (ADMID → `premis:originalName`) and whose Item divs resolve (fptr → fileSec → `FLocat/@xlink:href`) to unique, normalisable paths;
- one OBJECTS file group, or an unambiguous equivalent;
- SHA256 fixity per file;
- referential integrity of ADMID / DMDID / FILEID;
- no two divs claiming the same path.

[...]

PR #227, opened 2026-08-14
tool: survey real METS for path-navigability, and why it fails (#223)

tomcrane:
A `Manual`-category harness that runs the platform's own path cache over a directory of real METS files and reports what it made of each one — and where it made nothing, which shape of diagnostic stopped it. Point it at a corpus with `METS_SURVEY_PATH`; it reads and reports, writes nothing.

Groundwork for #223, and the tool that can answer "how many of the Archival Groups at Leeds would conform?" without anyone guessing.

## Why a what-if

A structMap the cache will not accept as physical stops it before it examines a single div, so one rejection hides everything behind it. Where a document has a single structMap that is untyped, or typed in a different case, the harness types it *in memory* and runs again. Nothing is written back.

## What it says about the METS we already have

| Document | As it stands | With the structMap typed |
|---|---|---|
| our own fixture | **6 cached, 0 diagnostics** | — |
| EPrints (sample, 4 files) | 0 cached, `no PHYSICAL structMap` | 0 cached, 4 × `div has no recognised TYPE` |
| **EPrints (real, 410 files)** | 0 cached, `no PHYSICAL structMap` | 0 cached, **410 × `div has no recognised TYPE`** |
| Archivematica | 0 cached, `no PHYSICAL structMap` | **38 cached**, 11 × `directory div has no ADMID` |
| Goobi 2026 | 0 cached, 29 × `div has no recognised TYPE` | — |

Three things worth pulling out:

**EPrints is two conventions away from fully navigable.** Both documents are flat — a root div and one div per file, no directory divs — and every `FLocat/@xlink:href` is already deposit-relative (`objects/650510_001.tif`). They carry no `premis:originalName`, but with no directory divs nothing needs it. What blocks them is that METS makes `TYPE` optional on both the structMap and the div, and EPrints omits both. Tolerating an untyped div that has an `fptr` takes the real 410-file document from 0 entries to **410 entries and 0 diagnostics** — I measured that with a temporary patch, which is not in this PR.

**Archivematica is blocked by letter case.** It writes `TYPE="physical"`; we compare against `"PHYSICAL"` ordinally. Behind that one comparison, 38 of its paths already resolve.

**Goobi 2026 would be a false positive.** Its divs are `TYPE="page"`, so div-tolerance would admit it — but its `FLocat` hrefs are absolute IIIF URLs, so the "paths" cached would be `https://…`. Navigability needs *deposit-relative* paths, and the cache does not currently check that. Worth knowing before tolerance is loosened.

## What this does not claim

Navigability is not editability. `SetFileAndFileGroup` requires a single `OBJECTS` fileGrp and these documents have `reference`, `original`, `DEFAULT`; there is no div for `objects` itself, so operations needing a parent div have nothing to attach to; and writing into a document that uses none of our conventions needs a policy, not just a parser change. The `mets:agent` gate sits on top of all of it, and that is the part #223 replaces.

Also: five documents is not a survey. That is what the tool is for.

Full suite green, CI-equivalent filter (`Category!=Manual`) unaffected: XmlGen 478, Preservation.API 106, Storage.API 55, Core 68, Common.Model 8.

### DIFF
--- a/src/DigitalPreservation/XmlGen.Tests/Experimental/Parsing/SurveyMetsCorpus.cs
+++ b/src/DigitalPreservation/XmlGen.Tests/Experimental/Parsing/SurveyMetsCorpus.cs
@@ -2,6 +2,7 @@ using System.Xml;
 using System.Xml.Linq;
 using System.Xml.Serialization;
 using DigitalPreservation.Mets;
+using FluentAssertions;
 using Xunit.Abstractions;
 
 namespace XmlGen.Tests.Experimental.Parsing;
@@ -43,6 +44,11 @@ public class SurveyMetsCorpus(ITestOutputHelper output)
             }
         }
 
+        // The one thing a survey can assert: if a corpus was pointed at, it was read. An empty
+        // result otherwise looks exactly like a corpus in which nothing was worth reporting,
+        // when in fact it usually means the path is wrong or holds no METS.
+        verdicts.Should().NotBeEmpty($"{PathVariable} ('{root}') should hold METS files to survey");
+
         output.WriteLine("");
         output.WriteLine($"=== {verdicts.Count} document(s) ===");
         foreach (var group in verdicts.GroupBy(v => v).OrderByDescending(g => g.Count()))
@@ -56,7 +62,7 @@ public class SurveyMetsCorpus(ITestOutputHelper output)
     /// <c>mets-objects</c> element, and a document we cannot deserialise is reported rather than
     /// thrown - a survey that stops at the first unreadable file surveys nothing.
     /// </summary>
-    private IEnumerable<(DigitalPreservation.XmlGen.Mets.Mets? Mets, string Label)> ReadMetsDocuments(string file)
+    private static IEnumerable<(DigitalPreservation.XmlGen.Mets.Mets? Mets, string Label)> ReadMetsDocuments(string file)
     {
         var name = Path.GetFileName(file);
         XDocument? document = null;
--- a/src/DigitalPreservation/XmlGen.Tests/Experimental/Parsing/SurveyMetsCorpus.cs
+++ b/src/DigitalPreservation/XmlGen.Tests/Experimental/Parsing/SurveyMetsCorpus.cs
@@ -187,7 +187,9 @@ public class SurveyMetsCorpus(ITestOutputHelper output)
         if (diagnostic.Contains("has no ADMID")) return "directory div has no ADMID";
         if (diagnostic.Contains("has no fptr")) return "item div has no fptr";
         if (diagnostic.Contains("no FLocat href")) return "item div's FILEID resolves to no FLocat href";
-        if (diagnostic.Contains("already", StringComparison.OrdinalIgnoreCase)) return "two divs claim one path";
+        // Matched via the constant, not a copy of the wording: this bucket was dead for exactly that
+        // reason, looking for an "already" that the diagnostic has never said.
+        if (diagnostic.Contains(MetsCache.DuplicatePathFragment)) return "two divs claim one path";
         return diagnostic;
     }
 
--- a/src/DigitalPreservation/XmlGen.Tests/Experimental/Parsing/SurveyMetsCorpus.cs
+++ b/src/DigitalPreservation/XmlGen.Tests/Experimental/Parsing/SurveyMetsCorpus.cs
@@ -1,0 +1,193 @@
+using System.Xml;
+using System.Xml.Linq;
+using System.Xml.Serialization;
+using DigitalPreservation.Mets;
+using Xunit.Abstractions;
+
+namespace XmlGen.Tests.Experimental.Parsing;
+
+/// <summary>
+/// A read-only survey of real METS files: what does the platform's path cache make of each one,
+/// and where it makes nothing, why?
+/// </summary>
+/// <remarks>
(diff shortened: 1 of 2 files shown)


src/DigitalPreservation/XmlGen.Tests/Experimental/Parsing/SurveyMetsCorpus.cs
```diff
@@ -0,0 +1,199 @@
...
+
+    private static string Shape(string diagnostic)
+    {
+        if (diagnostic.Contains("no PHYSICAL structMap")) return "no PHYSICAL structMap";
+        if (IsUntypedDiv(diagnostic)) return "div has no recognised TYPE";
+        if (diagnostic.Contains("no premis:originalName")) return "directory div has no premis:originalName";
+        if (diagnostic.Contains("has no ADMID")) return "directory div has no ADMID";
+        if (diagnostic.Contains("has no fptr")) return "item div has no fptr";
+        if (diagnostic.Contains("no FLocat href")) return "item div's FILEID resolves to no FLocat href";
+        if (diagnostic.Contains("already", StringComparison.OrdinalIgnoreCase)) return "two divs claim one path";
```
brianmac1967: This bucket is dead code. `MetsCache`'s actual duplicate-path diagnostic (`MetsCache.cs:90`) reads `"Divs 'X' and 'Y' both resolve to path 'z'"` - it never contains the substring `"already"`, so `diagnostic.Contains("already", ...)` never matches and this branch can't fire.

Concretely: point `METS_SURVEY_PATH` at a corpus containing a document where two divs resolve to the same deposit-relative path, and that diagnostic falls through to the raw `return diagnostic;` case instead of being grouped under "two divs claim one path" - each occurrence gets reported as its own distinct, ungrouped line (with div IDs baked into the text, so even repeated collisions of the same shape won't group with each other either). Doesn't affect any of the five documents surveyed in this PR (none hit this diagnostic type), but it means one of the seven intended diagnostic shapes silently doesn't work - worth fixing before this is pointed at a larger real corpus where duplicate-path collisions are more likely to show up. Probably wants to match on `MetsCache.DuplicatePathFragment` ("both resolve to path") instead of the string "already".

brianmac1967 (commented):
Reviewed this against real fixtures already in the repo, not just the description. Every code-behaviour claim checks out: genuinely read-only (no write/save call anywhere in the file), the what-if TYPE mutation is confirmed in-memory-only (mutates the deserialized object, re-runs Populate, nothing serialized back), correctly excluded from CI via `[Trait("Category", "Manual")]` (verified by running both `Category=Manual` and `Category!=Manual`), and the case-sensitive `TYPE=="PHYSICAL"` comparison plus the absence of any deposit-relative-vs-absolute-URL check both exist exactly as described in `MetsCache.cs`, not just this new tool.

On the corpus results table: 4 of the 5 rows correspond to fixtures already committed under `XmlGen.Tests/Samples/` (`EPrints.10315.METS.xml`, the Archivematica sample, `goobi-2026.xml`, and the "our own fixture" row). Pointed `METS_SURVEY_PATH` at them and ran the actual harness - every number reproduced exactly, including the specific diagnostic counts (38 cached / 11 ADMID diagnostics for Archivematica, 29 TYPE diagnostics for Goobi). The real 410-file EPrints export obviously isn't in the repo and can't be independently verified, but the diagnostic pattern claimed for it follows mechanically from the same code path already confirmed at 4 files, so it's consistent even though the exact count is unverifiable here. Full suite counts also matched exactly on an independent re-run. Description doesn't oversell anything I could find, aside from the one issue below.

One genuine bug, unrelated to the corpus claims, found while checking the diagnostic-bucketing logic - see inline comment.

src/DigitalPreservation/XmlGen.Tests/Experimental/Parsing/SurveyMetsCorpus.cs
tomcrane: Correct, and fixed in `414b255`. Good spot — this bucket has never fired.

I've matched on `MetsCache.DuplicatePathFragment` rather than a fresh guess at the wording, as you suggest. That constant exists precisely so the message and anything recognising it cannot drift; it was `internal`, so this file could not reach it and duplicated the idea badly instead. It is `public` now, with a note saying what it is for: a caller reading `PathDiagnostics` can tell a collision from the other six shapes only by this text, so recognising it is not purely an internal concern.

Verified rather than assumed, on a corpus built to trigger it — an AG-derived METS with one Item div duplicated, so two divs share a FILEID and resolve to one path, plus the unmodified document as a control.

Before:

```
 duplicate-paths.xml
     cached     : 10, diagnostics: 1
              1  Divs 'PHYS_objects_x002F_folder-b_x002F_folder-bb_x002F_minutes-laqm-22-april-2020.pdf' and 'PHYS_DUP_objects_…' both resolve to path 'objects/folder-b/folder-bb/minutes-laqm-22-april-2020.pdf'
```

After:

```
 control.xml
     cached     : 10, diagnostics: 0
 duplicate-paths.xml
     cached     : 10, diagnostics: 1
              1  two divs claim one path
```

Which also shows the second half of what you described: the div IDs baked into the raw text meant repeated collisions of the same shape each got their own line rather than grouping. That was the whole point of bucketing, so it would have been a poor thing to discover while pointing this at a large corpus.

Issue #228, opened 2026-08-14
Cleanups from the architecture review: a filter that isn't one, and an unguarded performance claim


tomcrane:
Two findings from the August architecture review that are small, carry no behaviour change, and do not depend on the step 3 migration or on any decision about #223. Raised separately from #226 because these are about the shape of the code rather than about ID opacity; #226 covers the four places that read meaning out of an ID string, and also proposes promoting `MetsIds` into production.

## 1. `IdRefs.ResolveSingle`'s `isUsable` parameter is not a filter

`ResolveSingle` takes a predicate, and then the callers re-apply it to what came back:

```csharp
// MetadataManager.cs:212
var amdSec = IdRefs.ResolveSingle(ctx.File.Admid, id => ..., CarriesPremisObject);
if (amdSec == null || !CarriesPremisObject(amdSec))
```

`MetsParser.cs:512` does the same and documents it — *"That makes the recheck the CALLER's job, here as at the other two sites."*

The asymmetry is deliberate and correct: the identity tier (the whole attribute value, which can only match in a legacy document) returns its match whether or not the caller can use it, so a broken reference fails loudly instead of resolving a fragment of a legacy ID against something unrelated. Only the per-token tier skips unusable candidates. That behaviour should not change.

But a parameter named like a filter, whose result may fail the filter, is the wrong shape — and it is why this took four review rounds to settle. The obviously-correct formulation already exists in the same file: `ResolveAll` returns the referenced elements with no predicate. Callers then write `candidates.FirstOrDefault(usable)`, and the precise diagnostic the asymmetry protects falls out for free — empty means the reference dangles, non-empty but nothing usable means the section exists but carries no PREMIS.

Two symmetric steps (canonicalise the reference, then select) instead of one asymmetric one. Deletes the two 3-arg overloads and the remarks explaining the trap. Needs a raw-string `ResolveAll` for the parser side.

One casualty to note: `MetsCache.cs:137-150` uses the predicate callback to memoise `ExtractPremisOriginalName` so the winning candidate is not parsed twice. That micro-optimisation goes, at the cost of one extra parse per directory div.

## 2. Nothing guards the performance claim the cache exists for

[...]

PR #229, opened 2026-08-14
chore: IDs stay opaque to code, and two decisions move from prose into structure

### LINKED ISSUE #226: METS IDs should be opaque to code: four places that still read meaning out of an ID string
METS IDs are meant to be **opaque to code and legible to people**: the ID carries the file's path deliberately — it is readable without this platform, and it inherits uniqueness from the path for free — but no code may infer anything from the string. Paths come from `FLocat/@xlink:href` and `premis:originalName`; references are followed by matching the raw string. This is now written up as a normative rule in [02d — Opaque to code, legible to people](https://github.com/digirati-co-uk/digital-preservation-docs/blob/mets-profiles/documentation/02d-METS-Identifiers.md#opaque-to-code-legible-to-people).

Because the ID really does contain the path, that is a **discipline rather than a property the format enforces** — nothing stops a reader parsing it. An audit of the platform's own code found five places that read meaning out of an ID string. Four are small and independent of any legacy corpus; this issue covers those four. The fifth is the legacy navigation tier, which is parked against #188 step 3 along with the virus-scan conventional-key fallback (see #188).

## 1. `MetadataManager.cs:347` — re-deriving an identifier from an ID it just minted

```csharp
var identifier = ctx.DigiprovId[Constants.VirusProvEventPrefix.Length..];
```

`ctx.DigiprovId` was built a few lines earlier from a known identifier; this strips the prefix back off to recover it for the occurrence numbering. Entirely self-inflicted — carry the identifier on `ProcessingContext` alongside `DigiprovId` and use it directly. No behaviour change.

## 2. `MetsParser.cs:611` — classifying a scan event by ID prefix

```csharp
.StartsWith(Constants.VirusProvEventPrefix, StringComparison.OrdinalIgnoreCase)
```

This selects the virus-scan `digiprovMD` *within an amdSec we have already resolved structurally*. The writer's actual guarantee for such an event is `MDTYPE="PREMIS:EVENT"` with a `premis:eventType` of `virus check` — not the ID prefix — so matching on the event's own type is both more honest and more tolerant of provenance we did not write. (This is exactly the rule the Playwright helper `getLatestVirusCheckEvent` now uses, in digital-preservation-e2e-tests#94.)

Note this is a *different* code path from the conventional-key fallback in the same file, which locates an amdSec when a file's `ADMID` resolves to nothing usable. That one is genuinely an ID-convention fallback and stays until step 3.

## 3. `ModsManager.cs:177` — deriving a dmdSec ID stem from a div's ID

```csharp
var idPart = localPath != null
    ? localPath.ToMetsId()
    : div.Id.RemoveStart(Constants.PhysIdPrefix);
```

The `localPath` branch is correct and is what #214 introduced. The fallback splits two unrelated cases:

- **a logical div** — `div.Id` is the client's own NCName-validated ID, so using it as a stem is reasonable, and stripping `PHYS_` off it is simply a no-op that reads as if it expects a physical ID;
- **a physical div whose path would not resolve** — here it really is inferring a path fragment from an ID, on precisely the documents where that inference is least safe.

Suggest splitting the two so the logical case says what it means, and the unresolvable-physical case goes to `UnusedDmdId` rather than guessing.

## 4. `FullMets.HasDuplicatePaths` — a dead flag with a comment that claims otherwise

Set in `MetsCache.cs:34`, declared in `FullMets.cs:33`, and **read by no production code** — only by test assertions in `CacheMutationConsistencyTests`. Meanwhile `MetsManager.cs:331` states that when a duplicate is recorded "every trust gate sees it", which is not true: every gate tests `PathDiagnostics.Count`.

Not ID-opacity as such, but found in the same sweep and in the same neighbourhood: delete the property, correct the comment, and reduce the tests to asserting the diagnostics. (This was also reported in the August cumulative review and not actioned.)

## Also worth doing while in here

`XmlGen.Tests/MetsIds.cs` is a tidy five-method funnel — `MetsIds.Phys(path)`, `.File(path)`, `.Adm(path)`, `.Tech(path)`, `.Dmd(path)` — that production does not have: prefixes are concatenated with `ToMetsId()` at 12+ sites across `MetsManager`, `MetadataManager`, `ModsManager` and `MetsFromArchivalGroup`. Promoting it into `DigitalPreservation.Mets` is roughly line-neutral, gives minting one authority, and makes any future change to the ID scheme (including step 3's) a one-file edit.

## Not in scope

- The legacy `PHYS_` + raw-path navigation tier (`PhysicalDivIdCandidates`) and `MetsParser.ScanOccurrence` — both serve pre-#214 documents and are tracked against #188 step 3.
- Anything requiring a migration of existing METS.

## Why bother

None of this is a live defect. The value is that "no production code infers anything from an ID string" is a checkable definition of done for the identifier work — sharper than a line count, and the thing that lets the platform edit third-party METS (#223) without caring what anyone else's IDs look like. Four small sites is close enough to that line to be worth closing.

tomcrane:
Closes #226 and #228 — the "cheap unconditional fixes" from the architecture review. One PR because they touch overlapping files and two would conflict.

Every one of these was a decision that was **right, but held in a comment instead of in the code**. No behaviour change except where called out.

## IDs are not parsed (#226)

| Site | Was | Now |
|---|---|---|
| `MetadataManager` | stripped the prefix back off an ID it had just minted, to recover the identifier | carries the identifier on `ProcessingContext` |
| `MetsParser` | recognised a virus-scan event by ID prefix | matches `premis:eventType`, which is the writer's actual guarantee |
| `ModsManager` | `div.Id.RemoveStart(PhysIdPrefix)` as a dmdSec stem | uses the div's own ID **whole**, encoded |
| `FullMets.HasDuplicatePaths` | written, never read, with a comment claiming otherwise | deleted |

Two of those deserve a note.

**The parser change is also more tolerant.** Matching on event type rather than ID prefix means provenance from tools we do not recognise is passed over instead of being mistaken for a scan because of its name. It is the same rule the Playwright helper now uses (digital-preservation-e2e-tests#94).

**The `ModsManager` change is a real behaviour change.** With no path to derive from, the minted dmdSec ID is now `DMD_` + the div's encoded ID. On a legacy div (`PHYS_objects/my file.pdf`) the old code produced `DMD_objects/my file.pdf` — **a brand new invalid `xs:ID`**, which is the thing #188 exists to stop. A test pinned that behaviour, and it was minting the invalid ID itself in order to have a spaced `DMDID` to resolve; it now puts the legacy reference on the div the way a pre-#188 document carries it, which is what it always meant to test. A new test asserts the minted ID is a valid NCName.

`HasDuplicatePaths` was reported in the August cumulative review and not actioned then.

## One authority for minting

`MetsIds` moves from `XmlGen.Tests` into `DigitalPreservation.Mets`, and the dozen hand-concatenated `Constants.XPrefix + path.ToMetsId()` sites across `MetsManager`, `MetadataManager`, `ModsManager` and `MetsFromArchivalGroup` route through it. The tests had the abstraction production lacked. Both sides now derive IDs the same way, and a change to the scheme — including step 3's — is a change to one file.

## A filter that was not one (#228)

`IdRefs.ResolveSingle` took an `isUsable` predicate, and every caller re-applied it to the result, because the identity tier returns its match usable or not. `MetsParser` even documented that the recheck was the caller's job.

Resolve, then select:

```csharp
IdRefs.ResolveAll(admid, lookup).FirstOrDefault(CarriesPremisObject)
```

This **keeps the asymmetry, structurally**: a joined-form or single-token match is an identity match and is therefore the only candidate, so an unusable one yields nothing rather than sending the caller on to match a fragment of a legacy ID against an unrelated section; a genuine multi-token list has no meaningful order, so every candidate is offered. Both 3-arg overloads and the remarks explaining the trap are deleted; a raw-string `ResolveAll` is added for the parser side.

`MetsCache`'s memoisation goes with them — it now stops at the first candidate that yields a name, so nothing is parsed twice and nothing past the winner is parsed at all. That is strictly less work than before.

The existing IDREFS tests, which were written to cover the asymmetry, pass unchanged. That is the evidence the semantics survived.

## The performance claim has a number again (#228)

A `Manual` benchmark reports the per-file cost of adding files at 200/400/800/1,600. It asserts no time — a wall-clock threshold in CI is a flaky test — but it makes the measurement the cache machinery exists for reproducible on demand:

```
   files    total ms     ms/file
     200          29        0.14
     400          43        0.11
     800         157        0.20
    1600         605        0.38

per-file cost at 1600 vs 200: ×2.65
```

0.38ms per file against the **0.44ms** recorded when the ID index was added — so the improvement is intact. The ×2.65 residual is `SortChildDivs` re-sorting a growing sibling list, which predates this work and is deliberately left alone; ×8 would mean the quadratic behaviour was back. All of that is in the doc comment, so the number explains itself to whoever runs it next.

## Verification

Debug **and** Release green: XmlGen 479, Preservation.API 106, Storage.API 55, Core 68, Common.Model 8.

One overlap to know about: this adds `Constants.VirusCheckEventType`, and so does #224. Both make the same change, so whichever merges second rebases cleanly or resolves to identical text.


src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
```diff
@@ -605,10 +607,15 @@ private void ProcessChildStructDivs(MetsFileWrapper mets, XElement parent,
...
+                // The writer's guarantee is a premis:event whose eventType is "virus check"; the
+                // ID convention is only a naming habit, and reading meaning out of an ID is the
+                // thing IDs are not for (see 02d, "Opaque to code, legible to people"). Matching
+                // on the event type is also more tolerant: provenance from tools we do not
+                // recognise is passed over rather than mistaken for a scan because of its name.
                 var amdScope = techMd?.Name == XNames.MetsAmdSec ? techMd : techMd?.Parent;
                 var digiprovMd = amdScope?.Elements(XNames.MetsDigiprovMD)
-                    .LastOrDefault(d => (d.Attribute("ID")?.Value ?? "")
-                        .StartsWith(Constants.VirusProvEventPrefix, StringComparison.OrdinalIgnoreCase));
+                    .LastOrDefault(IsVirusCheckEvent);
```
brianmac1967, line 618: Switching virus-event matching from an ID-prefix check to the content-based `IsVirusCheckEvent` here removes an implicit guarantee that used to hold at line 630 (just below this hunk, unchanged): `.Descendants(XNames.PremisEvent).SingleOrDefault()`.

Previously, a digiprovMD's ID had to start with the platform's own `digiprovMD_ClamAV_` prefix to be considered at all, so anything from a third-party or malformed METS was safely skipped regardless of its internal structure. Now matching is by content alone: if a third-party METS (which `MetsParser` is explicitly designed to tolerate, per this repo's own conventions) has a digiprovMD wrapping *more than one* `premis:event`, and one of those happens to have `eventType == "virus check"`, this now matches it via `IsVirusCheckEvent` - and the `SingleOrDefault()` two lines below throws `InvalidOperationException` instead of gracefully skipping unrecognised provenance. A tolerated case has become a parser crash. Might be worth `.FirstOrDefault()` there, or an explicit multi-event guard, now that content rather than ID shape decides what counts as "ours".

src/DigitalPreservation/DigitalPreservation.Mets/ModsManager.cs
```diff
@@ -184,6 +182,26 @@ public static void SetRecordInfo(this ModsDefinition modsDefinition, RecordInfo
...
+    /// is a no-op for the NCName-validated IDs a client supplies for logical ranges.
+    /// </remarks>
+    private static string MintDmdId(
+        DigitalPreservation.XmlGen.Mets.Mets mets, DivType div, string? localPath)
+    {
+        if (localPath != null)
+        {
+            return MetsIds.Dmd(localPath);
+        }
+        return div.Id.HasText() ? Constants.DmdIdPrefix + div.Id.ToMetsId() : UnusedDmdId(mets);
```
brianmac1967: This fallback still hand-concatenates `Constants.DmdIdPrefix + div.Id.ToMetsId()` instead of routing through the new `MetsIds` authority - one minting site left outside the PR's claimed "single production authority for minting IDs" (`MetsIds.Dmd` above, on line 200, correctly does route through it for the `localPath != null` case).

Not a runtime bug today - `MetsIds` has no localPath-less overload to call here - but it's exactly the maintenance risk the consolidation was meant to eliminate: a future change to the DMD ID scheme now has to remember two places instead of one. Might be worth adding a `MetsIds` overload for this ID-only case while the rest of the call sites are being migrated.

src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
```diff
@@ -152,6 +114,34 @@ public static IReadOnlyList<T> ResolveAll<T>(IReadOnlyList<string> tokens, Func<
         }
```
brianmac1967, line 114: Minor efficiency regression: the generic `ResolveAll<T>`'s multi-token branch just above (lines 108-113: `tokens.Select(lookupById).Where(...).Distinct().Cast<T>().ToList()`) eagerly materialises every candidate via `.ToList()` before the caller applies `FirstOrDefault(isUsable)` at each call site. The old `ResolveSingle(tokens, lookup, isUsable)` lazily stopped at the first usable candidate instead.

For an ADMID/DMDID list with many tokens where an early one is already usable, the new code still resolves and evaluates every remaining token's `lookupById` before selection happens - wasted work on every adopting call site (`MetadataManager`, `MetsParser`, `MetsCache`). Bounded in practice by typically-short IDREFS lists, so likely not worth blocking on, but worth knowing about if a document with unusually long ADMID lists shows up.

brianmac1967 (commented):
Extensively verified against the description - full diff review, targeted test runs in Debug and Release, ran the Manual benchmark, ran the 40 IDREFS tests, and grepped for residual hand-concatenation. Nearly everything holds up:

- The `MetadataManager`/`ModsManager`/`FullMets.HasDuplicatePaths` ID-opacity claims are all confirmed, including the `ModsManager` regression fix being a genuine fix (not coverage deletion) - the repurposed legacy test and new NCName test both pass and exercise real code paths.
- **Most importantly**: this PR's `IdRefs` "resolve, then select" refactor structurally fixes the `MetsParser` identity-tier bug flagged in an earlier PR review (#218) - usability filtering is now inseparable from the resolve call itself, so it's no longer possible to get an unusable identity-match result back without a check. Confirmed by reading the current code.
- The 40 existing IDREFS tests were actually run: 40/40 pass unchanged, as claimed.
- The benchmark was actually run: Release numbers (1.70ms/file @1600, x2.36 scaling) are the same order of magnitude and shape as claimed (0.38ms/file, x2.65), well clear of the x8 quadratic-regression signal. `SortChildDivs` confirmed untouched.
- Full suite counts matched exactly in both Debug and Release, independently re-run.

Three findings below, all non-blocking. One (`MetsParser.cs`) is worth taking seriously as a real robustness regression even though it's not what the PR set out to fix.

brianmac1967 (approved):

src/DigitalPreservation/DigitalPreservation.Mets/MetsParser.cs
tomcrane: You're right, and it is worse than you framed it — fixed in `a042899`, with a test that throws `System.InvalidOperationException : Sequence contains more than one element` without it.

Two things I hadn't followed through on when I made this change.

**The tolerance loss goes beyond that one line.** PREMIS allows several `eventOutcomeInformation` elements on a single event, so the `SingleOrDefault()` calls *inside* the event were the same hazard, reachable for the same reason — matching by content means the event's internals are now third-party too. The whole block is `FirstOrDefault` now.

**And "the only event" was the wrong question anyway.** The digiprovMD was selected because it *contains* a virus-check event, so the right element is that event. With an unrelated event listed first, `FirstOrDefault()` would have quietly reported an ingestion's outcome as the file's virus status — a wrong answer rather than a crash, which is worse. The test puts a failing `ingestion` event ahead of a passing `virus check` and asserts `HasVirus == false`, so it fails on both the throw and the silent mis-selection.

```csharp
var premisEvents = digiprovMd?.Descendants(XNames.PremisEvent).ToList() ?? [];
var virusEvent = premisEvents.LastOrDefault(IsVirusCheckEvent) ?? premisEvents.FirstOrDefault();
```

The fallback keeps the conventional-key path working, since that matches our own ID prefix on documents whose event may predate our writing `eventType` at all.

**One more thing this made me check, which the description should have said.** Matching on `eventType` doesn't only make us *skip* foreign provenance — it makes us *read* it. The Archivematica sample in this repo has 38 digiprovMDs containing `virus check`, and all 38 sit in the same amdSec as their techMD, so they resolve structurally. Before this PR they were invisible; now `MetsParser` reports Archivematica's own ClamAV results as the files' virus status. I think that is right — it is true provenance and the whole point of not reading meaning out of IDs — but it was an unannounced consequence, so `Can_Parse_Archivematica_METS` now asserts it explicitly rather than leaving it to be discovered.

src/DigitalPreservation/DigitalPreservation.Mets/ModsManager.cs
tomcrane: Agreed — fixed in `a042899`. `MetsIds.DmdFromDivId(div.Id)` now, with the overload you suggested:

```csharp
public static string DmdFromDivId(string divId) => Dmd(divId);
```

Delegating rather than repeating the concatenation, so the two entry points genuinely share one implementation and the scheme stays a single edit — a second `Constants.DmdIdPrefix + …ToMetsId()` body would have recreated the problem in the file that exists to prevent it. Separate name because the call site should say which stem it is passing; a path and a div ID are not interchangeable even though the encoding is.

src/DigitalPreservation/DigitalPreservation.Mets/IdRefs.cs
tomcrane: Leaving this one, and I think the eagerness is actually load-bearing rather than merely tolerable.

`MetsManager.cs:1307` does this:

```csharp
var dmdSecs = IdRefs.ResolveAll(div.Dmdid, id => mets.Mets.DmdSec.FirstOrDefault(d => d.Id == id))
    .Where(d => !referencedElsewhere.Contains(d.Id));
foreach (var dmdSec in dmdSecs)
    mets.Mets.DmdSec.Remove(dmdSec);
```

The lookup reads `mets.Mets.DmdSec` and the loop removes from it. That is safe only because `ResolveAll` has already materialised: the `Where` streams a snapshot that owes nothing to the collection being mutated. Deferred, each `MoveNext` would re-run `FirstOrDefault` over a collection shrinking underneath it, with `Distinct` carrying state across the mutations. Not obviously an exception, which is what makes it a bad thing to have.

On the cost: `lookupById` is an `MetsIdIndex` dictionary hit, and the branch only runs for genuine multi-token lists — a legacy space-containing ID resolves in the joined tier with one lookup. So the waste is a few O(1) probes on short lists.

And the early exit you're describing is what `ResolveSingle(tokens, lookup, isUsable)` did. That overload is what this PR deletes, because interleaving selection with resolution is exactly what let an unusable identity-tier match fall through to matching a *fragment* of a legacy ID against an unrelated section — the `MetsParser` bug from #218 you noted above as structurally fixed. Buying the probes back would mean giving that up, so I'd rather pay them.

Happy to revisit if a document with genuinely long ADMID lists turns up; it would be a change to this one method.

tomcrane:
Two fixed, one declined with reasons — replies on each thread, all in `a042899`.

**The `MetsParser` finding was right, and bigger than flagged.** Chasing it turned up two things I hadn't followed through when I made that change:

- The tolerance loss isn't confined to that line. PREMIS permits several `eventOutcomeInformation` elements on one event, so the `SingleOrDefault()` calls *inside* the event were the same hazard for the same reason — matching by content means the internals are third-party too. The whole block is `FirstOrDefault` now.
- "The only event" was the wrong question. The digiprovMD is selected because it *contains* a virus-check event, so that is the event to read. With an unrelated event listed first, taking the first would have reported an ingestion's outcome as the file's virus status — a wrong answer rather than a crash, which is the worse failure. The new test puts a **failing** `ingestion` event ahead of a **passing** `virus check`, so it catches both.

**And a behaviour change the description should have declared.** Matching on `eventType` doesn't only make us skip foreign provenance, it makes us read it. The Archivematica sample in this repo has 38 digiprovMDs containing `virus check`, all siblings of their techMD, so they resolve structurally — invisible before this PR, reported as the files' virus status after it. I think that's correct, and it's the point of not reading meaning out of IDs, but it shouldn't have arrived unannounced. `Can_Parse_Archivematica_METS` now asserts it.

**`ResolveAll`'s eagerness I've kept**, and on inspection it's load-bearing rather than incidental: `MetsManager.cs:1307` removes dmdSecs while enumerating a query whose lookup reads the very collection being removed from. That's safe only because the result is already materialised. Made lazy, each `MoveNext` would re-run the lookup over a shrinking collection with `Distinct` holding state across the mutations — not reliably an exception, which is what makes it a bad thing to have. The early exit you describe is also precisely what `ResolveSingle(tokens, lookup, isUsable)` bought, and that overload is what this PR deletes to fix the identity-tier bug you noted above.

Thanks for running the benchmark — your Release numbers (1.70ms/file at 1600, ×2.36) against mine (0.38ms, ×2.65) are a good illustration of why that test asserts a *shape* and not a time. The ratio is the signal; the absolute figure is whatever the machine of the day says.

Full suite green in Debug and Release: XmlGen 480, Preservation.API 106, Storage.API 55, Core 68, Common.Model 8.

Note for merge order: #224 and #229 both add `VirusCheckEventType` to `Constants.cs` with different doc comments, so this branch will need a one-hunk conflict resolution once #224 lands. I've trial-merged all four locally — comment text only, no semantic disagreement, and the whole set builds and passes together.

tomcrane:
Merged `main` in and resolved the `Constants.cs` conflict (`3454538`) — this is now MERGEABLE and green again. Both #224 and #229 had added `VirusCheckEventType` with the same value and different doc comments, and #224 also added `PremisNamespace`.

I kept both constants and took neither comment verbatim, because #224's had been overtaken by this branch: it said the event type is *"the only value by which we recognise one of our own scan events again"*, and after this PR that is no longer what it does — `MetsParser` matches any producer's virus-check event, ours or not, which is how Archivematica's 38 scans became visible. The resolved comment names both readers and what each wants from it.

Worth knowing that the conflict was why no checks appeared on `a042899`: GitHub can't build the merge ref for a conflicting PR, so the `pull_request` workflow never fired. Not a CI failure, just nothing to run.

Verified the rest of the sequence too — #227 then #229 merges clean, builds, and passes: XmlGen 488, Preservation.API 117, Storage.API 55, Core 68, Common.Model 8, Pipeline.API 4.

PR #232, opened 2026-08-19
feat: migrate preserved METS to legal xs:ID values (#188 step 3)

tomcrane:
Step 3 of #188: migrate the METS documents already preserved, and stop editing one making it worse.

Steps 1 and 2 stopped the platform *minting* invalid IDs. Nothing yet fixes what is already there — and today an edit to a pre-#214 Archival Group actively degrades it, because the new entries get encoded IDs while the existing ones keep their raw form. Editing is currently the only thing that creates mixed-generation documents.

## What decides that a document needs migrating

**Whether it contains an ID that is not a legal NCName** — not its age, and not who wrote it.

That predicate converges (after migration it is false), is idempotent, and never touches an ID that is already legal. The last part matters more than it sounds:

- client-supplied logical range IDs become public IIIF Range URIs via `ManifestBuilder.MakeRange`, and renaming one would change a published identifier for nothing;
- a pre-#214 document whose paths happened to hold no illegal characters needs no new version at all.

The `mets:agent` creator name narrows the population but is the filter, not the trigger. Measured over the whole held sample corpus the two agree completely: every third-party document (EPrints, Archivematica, all six Goobi variants) has **zero** invalid IDs anyway; every document the platform wrote before #214 has 8–78.

## The rewrite

`MetsIds.Normalise(id)` strips a known prefix (`PHYS_`, `FILE_`, `ADM_`, `TECH_`, `DMD_`, or the ClamAV event prefixes, whose remainder is itself an ID), encodes the stem, and puts the prefix back. A test pins that this is character-for-character what `MetsIds.Phys` and its siblings mint, so the two cannot drift.

It never reads a path and never asks what an ID means — it re-encodes the string it was given. That is what keeps the migration inside the "opaque to code, legible to people" rule (02d), which is the one place it would have been most tempting to break.

`MetsIdNormaliser` applies that across a document and follows every reference. `MetsGraph` walks the generated XmlGen classes reflectively, reading the schema's own `XmlAttributeAttribute` declarations rather than a hand-transcribed list of ~30 types — a visitor would be a second copy of an answer the schema already gives, silently wrong the day a type is missed. Two things a name-based sweep would get wrong and this does not:

- IDREFS arrive from the XmlSerializer already split on whitespace and a legacy ID contains spaces, so the joined form is tried first — same tiering, same reason, as `IdRefs`;
- `smLink/@xlink:from` is an IDREF while `smArcLink/@xlink:from` is an xlink *label*. Only the first moves.

Collisions (a rewritten ID landing on one that was already legal) are detected and the rewrite dropped with a warning, leaving the document unmigrated rather than ambiguous. `IMetsManager.NormaliseIds` rebuilds the path cache afterwards and refuses, writing nothing, if that turns up a diagnostic the document did not already have.

## Migrating one Archival Group

**The files do not need to move.** A deposit created against an existing Archival Group *without* export copies only the METS (`CreateDepositBase.EnsureMets` to `ExportArchivalGroupMetsOnly`). The diff still comes out right, because `WorkspaceManager` merges the METS physical structure into the combined tree, so every other file is present with the digest the group already holds, and `GetDiffImportJob`'s deposit-presence check applies only to adds and patches.

```
create deposit (no export) -> normalise -> if unchanged, stop
                           -> diff -> gate -> execute (event suppressed) -> verify
```

**The gate is the load-bearing part.** The import job must be exactly one binary to patch, and it must be the METS. On a METS-only deposit, a file the Archival Group holds but the METS does not mention would be listed for **deletion** rather than failing the diff. That one assertion closes it.

Result: one new OCFL version whose only new content is `mets.xml`, even for a 400-file object.

## The Activity Stream field

`ImportJob.SuppressActivityStreamEvent` keeps a maintenance version out of the published Change Discovery feed. Its readers — `iiif-builder` among them — treat an entry as "rebuild what you derived from this", and a rename of identifiers gives them nothing to rebuild.

It suppresses the **entry**, not the **record**. `ArchivalGroupEvent` rows are still written, with `Suppressed` set. That is deliberate and load-bearing: `StorageImportJobsProcessor` takes its watermark from the latest event date, so skipping rows would leave it re-reading the same window of Storage activities for ever — which a bulk migration, where every job is suppressed, would cause immediately. The watermark now lives in `PreservationContext.GetLatestArchivalGroupEvent()` with that invariant written down, and both the collection and page handlers filter identically, since between them they decide the page boundaries.

Two tests cover it, including one that would fail if only one side filtered.

## Migrating on write

`FeatureFlags:NormaliseMetsIdsOnWrite` (off by default) makes `MetsManager.WriteMets` migrate whatever it is about to write. `WriteMets` is the single point every mutation passes through — `HandleSingleChange`, `AddItemsToMets`, `DeleteItems`, `SetLogicalStructMap`, `RemoveLogicalStructMap`, `SetModsInformation` — which is why it hangs off there rather than off each of them. With it on, anything anyone edits fixes itself, and the corpus converges by attrition from the top while the bulk tool works up from the bottom.

**Deliberately not an import-job gate.** Refusing to preserve a deposit whose METS has invalid IDs was considered and rejected: it fails at the most expensive moment, after the work is done, and it fails machine clients (Goobi, the EPrints ingest, the Deposit Archiver) that have no way to act on "normalise the IDs first". Once nothing can produce an invalid ID and nothing writes one back unchanged, the same check becomes worth adding to `GetDiffImportJob` in a different role — a regression assertion that should never fire.

A flag rather than permanent behaviour for two reasons: the first real runs of the normaliser should happen in a controlled batch rather than while somebody waits for a page to save, and a write is *refused* if normalisation fails (which should be unreachable — paths are not touched — but would otherwise persist a document we have just decided we do not understand).

Consequence to have decided in advance: someone who edits one rights statement gets a METS diff touching every ID in the document, preserved under their name with a genuine Activity Stream event. Invisible at the import-job level, visible to anyone diffing two OCFL versions of `mets.xml`.

**Also fixed here:** `HandleSingleChange` discarded the `WriteMets` result and returned `Ok` regardless — survivable while a write could only fail on storage, not once it can be refused.

## UI

Behind `FeatureFlags:ShowNormaliseMetsIds`:

- **Actions → Normalise METS IDs** on a deposit. Says which of the two outcomes happened, because "already conforms" is the answer for most deposits and is not the same as "done".
- A **"Maintenance only — keep this version out of the Activity Stream"** checkbox beside Preserve, so a migration done by hand leaves the same clean history as one done in bulk.

## src/mets-id-migration

Python, driving the Preservation API — the METS rewriting stays in `MetsManager`, so there is no second METS writer. Nothing writes to S3, Fedora or OCFL directly.

```
survey --check-completeness   # read-only: walks the deposits query, decides per Archival Group
report                        # counts by state, and which characters make IDs illegal
list [--csv]                  # the actual list, for doing by hand
migrate --dry-run             # rehearse: everything up to the gate, then discard the deposit
migrate --limit 1 / migrate
verify                        # read-only
```

`survey` and `list` change nothing and answer the question that decides everything else: how many are there? If the answer is small — and on production it is expected to be — the UI route does the same job by hand and the bulk migration is not needed. A SQLite ledger makes the campaign resumable and is the only record of what the identifiers used to be, since after migration the old IDs exist only in the previous OCFL version.

## Things worth a reviewer's attention

- **`MetsGraph` uses reflection.** Deliberate, and argued in the class docs, but it is the least conventional thing here.
- **The METS-only deposit trick.** I traced it through `EnsureMets`, `CombinedDirectory.ToContainer` and `PopulateDiffTasks` and it holds, but it deserves a second pair of eyes, and it is what `migrate --dry-run` exists to prove per Archival Group on dev.
- **Do we want normalise-on-write on at all**, and if so when? Recommendation is after a campaign has been through the same documents.

## Testing

Full suite green: XmlGen 522, Preservation API 119, Storage 55, Core 68, Common.Model 8, Pipeline 4.

Normalisation is tested against the five **real** preserved documents in `Samples/` (`liddle`, `mets-sample-001`, `response-book`, `simple-image`, `wow`) as well as the fixtures — normalised, checked, and re-normalised as a no-op. Two of those already carry a dangling `div/@ADMID`, pre-existing and not the migration's business, so the test asserts "no more dangling than before" rather than zero.

Not yet tested on dev.

## Notes

- EF migration `AddSuppressActivityStreamEventToImportJobAndEvent`: two additive booleans, both defaulting false.
- `appsettings.Development*` is gitignored, so the new flags are `false` in the tracked files; turn them on locally to try it.
- Docs: `docs/issues/188/issue-188-step3-migration.md`, `src/mets-id-migration/README.md`, CLAUDE.md entry.


src/DigitalPreservation/DigitalPreservation.Mets/MetsIdNormaliser.cs
```diff
@@ -0,0 +1,304 @@
...
+/// <para>
+/// Nothing here reads a path, or asks what an ID means. Each ID is re-encoded as a string by
+/// <see cref="MetsIds.Normalise"/>, which is what makes it safe to run against a document whose
+/// conventions we only partly recognise, and what keeps the migration inside the rule that IDs are
+/// opaque to code.
+/// </para>
+/// <para>
+/// One METS feature is deliberately not covered: <c>BEGIN</c>/<c>END</c> hold an IDREF when the
+/// accompanying <c>BETYPE</c> says <c>IDREF</c>. The platform never writes that, and this only ever
+/// runs against a document the platform wrote (<c>GetFullMets</c> refuses any other agent), so the
```
brianmac1967: **Summary:** This comment claims BEGIN/END IDREF rewriting is safely skipped because "this only ever runs against a document the platform wrote (`GetFullMets` refuses any other agent)" — but `GetFullMets` performs no such check.

**Failure scenario:** A foreign-agent METS (Archivematica/EPrints/Goobi, explicitly supported per the platform's `MetsParser`) reaching `POST /deposits/{id}/mets/normalise` directly (via the UI or API, bypassing the Python migration tool's own `survey.py` FOREIGN-agent filter) and containing an `area` `BEGIN`/`END` IDREF has its target ID renamed but the `BEGIN`/`END` value left stale — silently writing back a METS with a dangling structural reference. Traced the full call chain (`DepositsController.NormaliseMetsIds` → `WorkspaceManager.NormaliseMetsIds` → `NormaliseMetsIdsHandler` → `metsManager.GetFullMets` → `S3MetsStorage.GetFullMets`) and confirmed no agent/creator guard exists at any layer.

src/DigitalPreservation/DigitalPreservation.Mets/MetsIdNormaliser.cs
```diff
@@ -0,0 +1,304 @@
...
+        return MetsIds.Normalise(value);
+    }
+
+    /// <summary>
+    /// Rewrite an IDREFS attribute, which the XmlSerializer has already split on whitespace. A
+    /// legacy ID containing spaces arrives as several tokens, so the joined form is tried first -
+    /// exactly as <see cref="IdRefs"/> does when resolving one, and for the same reason: no legal
+    /// ID contains a space, so a joined match can only be a legacy ID and is the right answer.
+    /// </summary>
+    private static int RewriteTokens(
```
brianmac1967, line 327: **Summary:** `RewriteTokens` matches multi-token IDREFS legacy values by rejoining tokens with `IdRefs.Joined` (always exactly one space), which fails to match a legacy ID containing irregular whitespace (e.g. two consecutive spaces), silently leaving the reference unrewritten with no warning.

**Failure scenario:** An amdSec has `ID="ADM_dir  file.pdf"` (two spaces) and a div references it via `ADMID="ADM_dir  file.pdf"`. The tokenizer splits the reference into `["ADM_dir","file.pdf"]` and `Joined()` rejoins as `"ADM_dir file.pdf"` (one space), which never matches the two-space dictionary key from `PlanRewrites`. Since both fragments (`ADM_dir`, `file.pdf`) are individually legal NCNames, the fallback per-token validity check also passes, so the `ADMID` is left pointing at the now-renamed, nonexistent old ID — with zero warning recorded in `MetsIdNormalisationReport`.

src/DigitalPreservation/DigitalPreservation.Mets/MetsIdNormaliser.cs
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    /// <summary>
+    /// Every ID that needs to change, old to new, with the collisions that would make the result
+    /// invalid ruled out first.
+    /// </summary>
+    private static Dictionary<string, string> PlanRewrites(
+        List<object> nodes, MetsIdNormalisationReport report)
+    {
+        var existing = new HashSet<string>(StringComparer.Ordinal);
```
brianmac1967, line 97: **Summary:** `PlanRewrites` tracks existing/invalid IDs as string sets rather than per-node, so two distinct elements that already (invalidly) share one ID value get silently merged onto the same new ID.

**Failure scenario:** Two separate elements both carry the corrupt duplicate ID `"1 2"`. `PlanRewrites` computes one rewrite mapping for that string, and the main loop in `Normalise` applies it to both nodes by value match — producing two elements with the identical new legal ID. The duplicate-ID corruption survives normalisation, contradicting `IMetsManager.NormaliseIds`'s doc comment that it "fails, writing nothing, if the result would not be a valid document." Nothing downstream (`RemoveColliding`, `MetsCache.Populate`, or the Python tool's `verify_migrated`) checks for post-rewrite ID uniqueness.

src/mets-id-migration/app/ids.py
```diff
@@ -0,0 +1,192 @@
...
+
+#: The mets:agent name the platform writes into documents it created. Only these are ours to
+#: migrate: EPrints, Archivematica and Goobi documents carry their own ID schemes, which are legal
+#: NCNames and were never affected by #188.
+METS_CREATOR_AGENT = "University of Leeds Digital Library Infrastructure Project"
+
+# XML 1.0 (5th edition) NameStartChar and NameChar, less the colon: an xs:ID must be an NCName.
+# Spelt out rather than approximated because the difference matters in both directions - an
+# accented letter is legal and must not be flagged, an ampersand is not and must be.
+_NAME_START = (
```
brianmac1967: **Summary:** This hand-rolled NCName-legality regex diverges from the platform's actual `XmlConvert`-based check in `MetsIds.cs`, with no shared test tying the two together.

**Failure scenario:** Enumerating every BMP codepoint against both implementations found 19,486 codepoints (e.g. U+0132–0133, U+017F, ranges in Latin Extended-B/Greek/Cyrillic) that this regex accepts as legal NCName-start characters but that .NET's `XmlConvert.IsStartNCNameChar` (used by `MetsIds.IsValidId`, the actual platform logic in `MetsManager.NormaliseIds`) rejects. A legacy METS ID starting with one of these codepoints would be judged "already valid, no migration needed" by this survey tool while the platform itself still considers it invalid — the migration campaign would silently skip an Archival Group that still has an illegal xs:ID. No test in `tests.py` or the C# suite cross-checks the two definitions.

brianmac1967 (commented):
Automated review of PR #232 (issue #188 step 3 METS ID migration). 8 findings verified against the code, most severe first. Full methodology: 8 finder angles × independent verification per candidate.

src/DigitalPreservation/Preservation.API/Features/Deposits/DepositsController.cs
```diff
@@ -244,6 +244,67 @@ public async Task<IActionResult> AddItemsToMets([FromRoute] string id, [FromBody
...
+    /// </para>
+    /// </remarks>
+    [RequireFeatureFlag("EnableMetsIdNormalisation")]
+    [HttpPost("{id}/mets/normalise", Name = "NormaliseDepositMetsIds")]
+    [ProducesResponseType<MetsIdNormalisationReport>(200, "application/json")]
+    [ProducesResponseType<ProblemDetails>(400, "application/json")]
+    [ProducesResponseType<ProblemDetails>(404, "application/json")]
+    [ProducesResponseType<ProblemDetails>(401, "application/json")]
+    [ProducesResponseType<ProblemDetails>(409, "application/json")]
+    public async Task<IActionResult> NormaliseMetsIds([FromRoute] string id)
```
brianmac1967, line 268: **Summary:** This endpoint calls `workspaceManagerFactory.CreateAsync(deposit)`, which eagerly builds a `CombinedDirectory` (S3 file listing + full METS parse via `MetsParser`) that is never used, then the handler does an entirely separate S3 fetch + parse of the same METS via `MetsManager.GetFullMets`.

**Failure scenario:** Every call to this endpoint — including once per Archival Group during the bulk ID-migration campaign (hundreds to low-thousands of groups per the migration tool's own README) — pays for two S3 reads and two full METS parses of the same document instead of one, doubling latency/cost on the hot path the whole migration campaign is built around.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -44,8 +55,52 @@ public async Task<Result<MetsFileWrapper>> CreateStandardMets(Uri metsLocation,
...
+    /// is set.
+    /// </summary>
+    /// <remarks>
+    /// This is the single point every mutation passes through - HandleSingleChange, AddItemsToMets,
+    /// DeleteItems, SetLogicalStructMap, RemoveLogicalStructMap and SetModsInformation all end here
+    /// - which is why the migration hangs off it rather than off each of them. Building a fresh
+    /// document (CreateStandardMets, MetsFromArchivalGroup) also comes through, and normalising one
+    /// of those is a no-op, because nothing minted since issue #214 is invalid.
+    /// </remarks>
     public async Task<Result> WriteMets(FullMets fullMets)
```
brianmac1967, line 69: **Summary:** `WriteMets` unconditionally re-runs `NormaliseIds` when `MetsManagerOptions.NormaliseMetsIdsOnWrite` is enabled, with no check for whether the document was already normalised — so `NormaliseMetsIdsHandler`'s explicit `NormaliseIds` call followed by `WriteMets` triggers a guaranteed no-op second full graph walk when both are enabled together.

**Failure scenario:** An operator leaves `NormaliseMetsIdsOnWrite` enabled (plausible post-migration per the tool's own README) while also calling the explicit `NormaliseMetsIds` endpoint. Every such call now does two full reflection-based `MetsGraph` traversals instead of one, wasting a complete graph walk for zero effect (guaranteed `report.Changed == false` on the second pass).

src/mets-id-migration/app/survey.py
```diff
@@ -0,0 +1,195 @@
...
+from typing import Iterator
+
+from lxml import etree
+from logzero import logger
+
+from app import api, ids, settings
+from app.ledger import CANDIDATE, CONFORMS, FOREIGN, NO_METS, Ledger
+
+
+def archival_group_paths(newest_first: bool = False) -> Iterator[str]:
```
brianmac1967: **Summary:** `archival_group_paths()` pages through the entire deposits table from the beginning on every survey run with no resume cursor/bookmark; already-known paths are only skipped after being fetched.

**Failure scenario:** A second or third incremental survey run over a growing deposits table (which the migration campaign itself grows, since `migrate` creates one deposit per Archival Group) re-walks the full history of deposit pages just to re-skip already-recorded rows. Cost scales with total deposits ever created rather than new deposits since the last run, making each rerun progressively more expensive as the corpus and the campaign's own deposits accumulate. The tool's own README acknowledges this: "A full survey of development means walking every deposit it has ever had."

src/DigitalPreservation/Preservation.API/Features/Activity/Requests/GetArchivalGroupsOrderedCollectionPage.cs
```diff
@@ -25,11 +25,17 @@ public class GetArchivalGroupsOrderedCollectionPageHandler(
     
     public async Task<Result<OrderedCollectionPage>> Handle(GetArchivalGroupsOrderedCollectionPage request, CancellationToken cancellationToken)
     {
-        var totalItems = await dbContext.ArchivalGroupEvents.CountAsync(cancellationToken: cancellationToken);
-        
+        // Published events only, and the same filter as the count in
```
brianmac1967: **Summary:** The identical `.Where(e => !e.Suppressed)` filter is duplicated three times across two files (once in `GetArchivalGroupsOrderedCollection.cs`, twice here), with this handler's own comment admitting a mismatch between copies would misplace pagination's 'next' link.

**Failure scenario:** A future change to the suppression rule (e.g. also excluding soft-deleted events) gets applied to only some of the three copies, desynchronizing the count and page queries exactly as the comment on lines 28-30 warns — producing a 'next' link on the last page, or omitting one that should exist. The same PR already introduces a shared-extension pattern (`PreservationContext.GetLatestArchivalGroupEvent`) that could have unified all three sites but wasn't applied here.

src/DigitalPreservation/Preservation.API/Features/ImportJobs/Requests/ExecuteImportJob.cs
```diff
@@ -71,6 +79,7 @@ public async Task<Result<ImportJobResult>> Handle(ExecuteImportJob request, Canc
                 LastUpdated = now,
                 DateSubmitted = now,
                 SourceVersion = storageImportJobResult.SourceVersion,
+                SuppressActivityStreamEvent = request.ImportJob.SuppressActivityStreamEvent,
```
brianmac1967, line 82: **Correctness/safety:** `ImportJob.SuppressActivityStreamEvent` is honored here with no authorization, feature-flag, or content check that the job is actually METS-only.

Any caller of `POST /deposits/{id}/importjobs` (UI, Goobi, EPrints ingest, ad-hoc scripts) can set `suppressActivityStreamEvent: true` on a job that adds/deletes real binaries; the content change is applied but never announced via the Activity Stream, so iiif-builder never rebuilds the manifest, with no detection or recovery. The sibling `POST /deposits/{id}/mets/normalise` endpoint is properly gated with `[RequireFeatureFlag("EnableMetsIdNormalisation")]` but this one is not. The migration tool's own client-side-only "refuse any diff that isn't a single METS patch" check (`src/mets-id-migration/app/migrate.py::_refuse_unless_mets_only`) is the sole safety net for this exact endpoint, creating a TOCTOU gap between `get_diff_import_job` and `execute_import_job` with no server-side backstop.

src/DigitalPreservation/DigitalPreservation.Mets/MetsManager.cs
```diff
@@ -79,8 +171,11 @@ private async Task<Result> HandleSingleChange(Uri workingRoot, string? depositET
             var editMetsResult = EditMets(workingBase, deletePath, fullMets);
             if (editMetsResult.Success)
             {
-                await WriteMets(fullMets);
-                return Result.Ok();
+                // The write result was previously discarded and Ok returned regardless. That was
+                // survivable while WriteMets could only fail on storage, but it can now refuse a
+                // document whose IDs would not migrate, and reporting success for a file that was
+                // not written is the worst of the available outcomes.
+                return await WriteMets(fullMets);
```
brianmac1967, line 195: **Correctness (latent, gated behind `NormaliseMetsIdsOnWrite`):** `WriteMets` can now fail here (when `NormaliseIds` refuses the document - e.g. duplicate IDs, or a rewrite that would make a div unresolvable by path) where previously it could only fail on storage errors.

However, several existing callers outside this diff perform their real S3 mutation and deposit-tracking state update **before** calling into this path, with no rollback on failure: `DigitalPreservation.Workspace/Requests/DeleteItems.cs` (S3 delete + JSON removal, then `WriteMets`), `DeleteObject.cs` (S3 delete, then `HandleDeleteObject` -> `WriteMets`), and `UploadFileToDeposit.cs` (S3 `PutObject` + `AddToDepositFileSystem`, then `HandleSingleFileUpload` -> `WriteMets`).

Once `NormaliseMetsIdsOnWrite` is turned on (the PR's own docs describe this as the intended eventual production state), a delete/upload against a pre-existing METS with a duplicate-ID or unnavigable-after-rewrite problem deletes/uploads the S3 object first, then this refuses - leaving S3 and METS permanently out of sync with no compensating action. Currently dormant since the flag defaults to `false` in all appsettings, but worth a rollback/compensation path before the flag is ever flipped on.

brianmac1967, line 195: **Correctness (latent, gated behind `NormaliseMetsIdsOnWrite`):** relatedly, `CreateDepositBase.CreateFolderInMets` (not touched by this diff) calls `metsManager.HandleCreateFolder(...)` with a bare `await` and never inspects the `Result` this method now propagates from `WriteMets`; its own callers likewise ignore the return value.

If normalisation ever fails during deposit creation, `CreateDepositRequest` reports the deposit as created successfully even though its METS is silently missing the `metadata`/`metadata/ad-hoc` folder entries it should have. Worth auditing all callers of `HandleCreateFolder`/`HandleSingleFileUpload`/`HandleDeleteObject` for unhandled `Result` now that failure is a reachable outcome.

brianmac1967 (commented):
Automated review pass (recall-biased, high effort). 10 findings posted inline, ranked most-severe first.

src/DigitalPreservation/DigitalPreservation.UI/Pages/Deposits/ImportJobs/ImportJob.cshtml
```diff
@@ -81,6 +81,26 @@ ViewData["section"] = "deposits";
                     else
                     {
                         <form class="single-submit" method="post" asp-page-handler="ExecuteDiffDirect">
+                            @if (Model.ShowNormaliseMetsIds)
```
brianmac1967, line 84: **Correctness/UX:** the "keep out of Activity Stream" checkbox is gated only by a display feature flag (`ShowNormaliseMetsIds`), not by whether the diff being executed is actually METS-only, and there is no un-suppress/republish mechanism once `ArchivalGroupEvent.Suppressed` is set.

An operator could tick this on a diff that also changes real binaries (left over from a migration session, or by habit); the version preserves with the event permanently hidden from the Activity Stream, recoverable only via a brand-new import job.

src/mets-id-migration/app/migrate.py
```diff
@@ -0,0 +1,186 @@
...
+    for key in ("binariesToAdd", "binariesToDelete", "binariesToRename",
+                "containersToAdd", "containersToDelete", "containersToRename"):
+        entries = import_job.get(key, [])
+        if entries:
+            raise MigrationRefused(
+                f"{key} is not empty ({len(entries)} entries, e.g. "
+                f"{entries[0].get('id')}); refusing to preserve")
+
+
+def _is_mets_file(slug: str) -> bool:
```
brianmac1967, line 147: **Maintainability/drift risk:** `_is_mets_file` re-implements the platform's `MetsUtils.IsMetsFile` permissive `.xml` + contains-"mets" matching in Python, tied to the C# original only by a code comment, with no automated cross-check test - unlike the NCName-range logic elsewhere in this tool, which is pinned by a generated `ncname_ranges.json` cross-checked by `NCNameRangesTests.cs` on the .NET side.

A future change to `MetsUtils.IsMetsFile` isn't forced to be mirrored here, which would silently weaken the one check that identifies which file in a diff is "the METS file" for the METS-only-diff safety gate governing Activity Stream suppression.

src/DigitalPreservation/DigitalPreservation.Mets/MetsIdNormaliser.cs
```diff
@@ -0,0 +1,413 @@
...
+        var report = new MetsIdNormalisationReport();
+        var nodes = MetsGraph.Nodes(mets).ToList();
+
+        var plan = PlanRewrites(nodes, report);
+        if (report.DuplicateIds.Count > 0)
+        {
+            // Refused before anything is touched, so the caller gets the document as it was.
+            return report;
+        }
+        if (plan.ById.Count == 0 && !AnyInvalidReference(nodes))
```
brianmac1967, line 51: **Efficiency:** `AnyInvalidReference` runs a second full reflective pass over every node (via `SingleIdRefs`/`TokenIdRefs`/`UntypedIdRefs` reflection helpers) even when `plan.ById.Count == 0` (nothing to rewrite by ID).

Since `MetsManager.WriteMets` runs `Normalise` on every mutation once `NormaliseMetsIdsOnWrite` is enabled (per this class's own doc comment, "the single point every mutation passes through" - the intended eventual production state, not just for the one-off migration), once the corpus converges and most writes have nothing left to rewrite, every ordinary save of a large METS document (can hold thousands of divs) permanently pays this second reflection-based tree walk it didn't need before this PR.

src/mets-id-migration/app/api.py
```diff
@@ -0,0 +1,253 @@
...
+        raise RuntimeError(f"No access token from AAD: {token.get('error_description', token)}")
+    headers["Authorization"] = f"Bearer {token['access_token']}"
+    return headers
+
+
+def _url(path: str) -> str:
+    return f"{settings.PRESERVATION_API}/{path.lstrip('/')}"
+
+
+def _request(method: str, path: str, what: str, **kwargs) -> requests.Response:
```
brianmac1967, line 72: **Reuse/efficiency:** `_request` hand-rolls a fixed-pause retry loop instead of using `requests`/`urllib3`'s built-in `Retry` + `HTTPAdapter` (the standard idiom for retrying idempotent calls with backoff), and no `requests.Session()` is used anywhere, so every call opens a fresh connection with no pooling.

At production scale (tens of thousands of Archival Groups per `settings.py`'s own documentation), every HTTP call pays full TCP/TLS setup instead of reusing a pooled connection, and transient 429/503 responses are retried on a fixed sleep rather than honoring `Retry-After`.

src/DigitalPreservation/DigitalPreservation.Mets/MetsIds.cs
```diff
@@ -38,4 +40,90 @@ public static class MetsIds
...
+    ];
+
+    /// <summary>
+    /// <c>digiprovMD_ClamAV_</c>, optionally followed by the occurrence counter that
+    /// <see cref="Constants.NumberedVirusProvEventId"/> puts there. What follows is another ID
+    /// (see <see cref="VirusProvEvent"/>, which concatenates rather than encoding), so it is
+    /// normalised in its own right.
+    /// </summary>
+    private static readonly Regex VirusProvEventPrefixPattern = new(
+        $"^{Regex.Escape(Constants.VirusProvEventPrefix)}(?:[0-9]+_)?",
```
brianmac1967, line 82: **Correctness (narrow edge case):** this pattern greedily consumes any `[0-9]+_` immediately after the ClamAV prefix as an occurrence number, assuming that digit run is always the occurrence tier. Safe for documents this platform wrote (since `DigiprovIdentifier` always starts with a letter), but `NormaliseIds`/`WriteMets` don't verify a METS's creating agent - a foreign document with an id shaped like `digiprovMD_ClamAV_2020_report.pdf` (where `2020_report.pdf` was meant as one atomic remainder) would have `2020_` silently mis-consumed, producing a different normalised id than encoding the whole remainder as one token, with no warning raised.

src/mets-id-migration/app/survey.py
```diff
@@ -0,0 +1,383 @@
...
+
+def check_completeness(ledger: Ledger, partial: bool = False) -> None:
+    """
+    Answer a question the survey cannot answer about itself: did the deposits query show us
+    everything?
+
+    The survey finds Archival Groups through deposits, and an Archival Group whose deposit rows were
+    hard-deleted therefore does not appear. The Activity Stream is the other record of the same
+    population - it has an entry for every create and update - so comparing the two says whether
+    anything was missed. It reads the stream and nothing else; the survey itself is unaffected by
```
brianmac1967, line 346: **Minor correctness:** `_path_under_root(object_id)` can return `None` when an activity's object id doesn't contain a `/repository/` segment, and that `None` is added to `from_stream` unfiltered here. The completeness summary later prints `len(from_stream)` without filtering, so a malformed/unexpected activity id inflates the operator-facing "Archival Groups mentioned in the activity stream" count by one per occurrence (the actual `missing` set computation does correctly filter `None`, so only the printed count is affected - cosmetic, but could confuse an operator cross-checking survey completeness).

src/mets-id-migration/app/api.py
```diff
@@ -0,0 +1,253 @@
...
+    enough to investigate with; rerunning the command is the retry.
+    """
+
+    def __init__(self, what: str, detail: str, status_code: int | None = None):
+        super().__init__(f"{what}: {detail}")
+        #: The HTTP status, or None when there was no response at all (timeout, connection lost).
+        self.status_code = status_code
+
+
+def _headers() -> dict[str, str]:
```
brianmac1967, line 47: **Reuse:** `_headers()` re-implements iiif-builder's MSAL client-credentials token acquisition (`ConfidentialClientApplication` + `acquire_token_silent`/`acquire_token_for_client`) inline rather than reusing/extracting the existing implementation in `src/iiif-builder/app/preservation_api.py` (`get_preservation_headers()`).

Two independent OAuth2 client-credentials implementations against the same Preservation API can drift - a bug fix or token-cache improvement made in one place silently doesn't propagate to the other.

Issue #237, opened 2026-08-21
MetsParser cannot see EPrints record identifiers (MODS vocabulary in the METS namespace)


tomcrane:
Found while checking the #223 EPrints-tier assumptions against twelve recent production EPrints migrations (2026-08-21; recorded in 02e "Quirks measured in the production corpus" in the docs repo).

## The corpus fact

Every recent EPrints-migrated METS carries record identifiers — the EPrints id, the EMu number, and the **`id.library.leeds.ac.uk` PID** — but in the **METS namespace**, inside a dmdSec whose wrapper claims MODS:

```xml
<mets:dmdSec ID="DMD_eprint_36675">
  <mets:mdWrap MDTYPE="MODS">
    <mets:xmlData>
      <mods:titleInfo><mods:title>Growing</mods:title></mods:titleInfo>
      <mets:recordInfo>
        <mets:recordIdentifier source="EPrints">36675</mets:recordIdentifier>
        <mets:recordIdentifier source="id.library.leeds.ac.uk">xvc3d1gg</mets:recordIdentifier>
        <mets:recordIdentifier source="EMu">441458</mets:recordIdentifier>
      </mets:recordInfo>
      ...
```

`MetsParser.cs:864` reads only `mods:recordIdentifier`, so all of these are invisible to the platform: EPrints documents parse as having no record info at all. (Titles work because `titleInfo`/`title` are the one part EPrints prefixed correctly.)

## Whose bug is it?

EPrints'. The document is XSD-valid only because `mets:xmlData` is `xs:any processContents="lax"` and the METS schema has no `recordInfo` declaration to check it against — but `MDTYPE="MODS"` claims the content is MODS, and MODS defines these elements in the MODS namespace. Validated *as MODS*, the content fails (wrong namespace, no `mods:mods` root). Our parser is correct by the book.

But the parser's whole design is tolerance for exactly this kind of third-party sloppiness, and the identifiers matter — the PID in particular links the item to the Identity Service from inside the preserved METS.

## The fix

[...]

Issue #236, opened 2026-08-21
MetsParser reads any premis:storage regardless of storageMedium, and throws on two


tomcrane:
Found while checking the #223 EPrints-tier assumptions against twelve recent production EPrints migrations (2026-08-21; recorded in 02e "Quirks measured in the production corpus" and RFC 008 in the docs repo).

## The corpus fact

Every file's techMD in recent EPrints-migrated METS carries a storage assertion pointing at the EPrints server it was migrated from:

```xml
<premis:storage>
  <premis:contentLocation>
    <premis:contentLocationType>URL</premis:contentLocationType>
    <premis:contentLocationValue>file:///usr/share/eprints/archives/digitallibrary/documents/disk0/00/03/66/75/01/LEEUA_1959.010_01.tif</premis:contentLocationValue>
  </premis:contentLocation>
</premis:storage>
```

No `storageMedium`. It is historical provenance, not a live location.

## The two defects

The two stacks disagree about how to read storage:

1. **`MetsParser` ignores `storageMedium`.** `PremisManager.cs:57` (the editing stack) reads a `StorageLocation` only from a `premis:storage` whose `storageMedium` is our agent name — correct, since the platform's own assertion is marked that way when written (`PremisManager.cs:118`). `MetsParser.cs` (~556 for files, ~422 for directories) takes **any** single `contentLocation` — so today the platform records the dead EPrints server path as these files' storage location.

2. **`SingleOrDefault()` throws on two storage elements.** Both parser sites use `Descendants(PremisContentLocation).SingleOrDefault()`, which raises `InvalidOperationException` when a techMD holds two `contentLocation`s. "Theirs plus ours" is exactly what the exported METS of an *edited* EPrints item will contain, because the platform adds its own marked `premis:storage` alongside on export. So the moment #223 enables editing of EPrints documents, re-parsing an exported edited item crashes the parser.

## The fix

In `MetsParser`, resolve storage the way `PremisManager` already does: prefer the `premis:storage` whose `storageMedium` is our agent name; treat any number of others as present-but-not-ours (preserved, not read as *the* location); never throw on multiples. A foreign-only document (like every EPrints migration) then yields no storage location, which is the truth.

## Sequencing

**This gates #223 Phase 4** (enabling conformance-based editing). It is worth fixing before then anyway — defect 1 is live today, just harmless-looking.

PR #238, opened 2026-08-21
feat: the METS editability judge (Schematron + twin .NET/Python implementations)

tomcrane:
Issue #223 Phase 3: the runnable editability judge — one contract, one Schematron rule source, two twin implementations.

## What this is

**`src/mets-editability/CONTRACT.md`** implements the docs repo's `02e-METS-Editability.md` as a concrete spec: four verdicts (`editable` / `editable-with-normalisation` / `navigable-read-only` / `not-editable`), a shared finding-code vocabulary, the structMap selection rules, the deposit-relative guard, and — for the EPrints tier — the first save's mutation list as a dry run. Both implementations are written from the contract, not from each other; when they disagree, the contract decides which is wrong.

**`src/mets-editability/`** (Python, standalone, lxml only):
```
python mets_editability.py judge path/to/mets.xml [--json]
```
Deliberately platform-free — the intended seed of the future published METS library.

**`DigitalPreservation.Mets.Conformance`** (.NET): same contract, reusing `MetsIds.IsValidId` (the XmlConvert NCName authority) and nothing else, so it stays extractable.

**`schematron/*.sch`** holds the XML-visible tier rules once. Python executes the `.sch` directly (lxml ISO Schematron); .NET runs the compiled XSLT (committed under `schematron/compiled/`, regenerated by `tools/compile_schematron.py`, pinned by a behavioural-equivalence test). Native code on both sides covers what XPath 1.0 cannot: path resolution and uniqueness, the deposit-relative guard, NCName legality (`ncname_ranges.json` pinned byte-for-byte to the migration tool's copy, itself pinned to XmlConvert by `NCNameRangesTests`).

## Acceptance

Both suites enforce the same table over the real sample corpus — the measured #223 conformance data re-expressed as verdicts:

| Document | Verdict |
|---|---|
| `simple-image.mets.xml`, `wow.mets.xml` | `editable` |
| `path-fixture-spaces.xml` | `editable` + `LEGACY_IDS` note |
| `EPrints.10315.METS.xml` | `editable-with-normalisation` + the six 02e mutations in order |
| Archivematica | `navigable-read-only` (38 files; `DIRECTORY_DIV_NO_ADMID` ×11) |
| `goobi-wc-b29356350.xml` | `navigable-read-only` |
| `goobi-2026.xml` | `not-editable` (`HREF_NOT_DEPOSIT_RELATIVE` ×89) |

Also validated against the twelve recent production documents from the corpus check (not committed): the ten EPrints migrations judge `editable-with-normalisation`, the two legacy platform documents judge `editable` + `LEGACY_IDS`, and the corpus quirks surface as notes (`FOREIGN_STORAGE_LOCATION` → #236, `METS_NAMESPACE_RECORD_INFO` → #237).

Python: 26 tests. .NET: 23 tests. Full solution builds clean; the new test project runs in ordinary CI.

## Not in scope

No platform behaviour changes: the agent-name gate, `MetsParser`, `MetsCache` are untouched. Wiring the judge into the platform (replacing the gate, the reading tolerances + deposit-relative guard, the on-save restructure, #236/#237) is Phase 4, gated on #232 bedding in and Leeds' agreement on RFC 008.


tomcrane:
## Adversarial review complete — all fourteen findings fixed (`387e45f`)

A four-angle adversarial review (line-by-line diff, removed-behaviour audit against the platform's actual parser, cross-file seam tracing, altitude) with adversarial verification produced ten capped findings plus four below-cap items. All fixed:

**The judge was more permissive than the machinery it vouches for** — the review's core theme:
- `C_PHYSICAL_FPTR_HAS_FILEID`: `MetsParser.GetRequiredFileId` NREs on any physical fptr without `@FILEID`; the judge accepted area-only fptrs. Now a common rule, scoped to the judged map (logical maps keep area-only pointers — the platform writes them there itself).
- `C_ONE_FLOCAT`: the parser takes `.Single()` FLocat and its href; the judge read only the first. Now every file needs exactly one FLocat with an href (this replaces `E_FILE_HAS_HREF` — one fact, one code).
- `E_SHA256`/`P_SHA256` require a non-empty **digest value**, not just the algorithm label.
- `E_MIXED_FILEGRP_USE` gains a presence-mix clause (XPath `!=` against a missing `@USE` is silently false).

**Contract violations**: `NO_FILES` (a zero-file document — the platform's own new-deposit skeleton — was refused with an empty reasons list); `DUPLICATE_PATH`/`fileCount` now count distinct files, not references.

**Chosen-map scoping** (found independently by all four angles): the Schematron structural rules are now scoped to the judged structMap via shared entity definitions, and `P_PHYSICAL_STRUCTMAP` asserts the *chosen* map is exactly `PHYSICAL` — an unchosen sibling can neither demote the verdict nor carry a tier.

**Twin/CLI robustness**: `judge.py` returns `PARSE_FAILED` for unreadable files (matching .NET) instead of a traceback; folder mode drops the Linux-hostile case-sensitive glob and reports what it skipped; `ncname.py` anchors with `\Z` (Python `$` matches before a trailing newline; XmlConvert doesn't); `PLATFORM_AGENT` is pin-tested against `Constants.cs`.

**CI**: new `test-mets-editability` job — recompiles the Schematron and fails on any diff (so the .NET-embedded compiled rules can never silently go stale against the sources Python executes), then runs the Python suite; the workflow path filter now covers `src/mets-editability/**`.

**Performance** (below-cap finding, made acute by the scoping fix): `xsl:key` + structMap-aggregated rules took the 355-file production document from 18s to **0.15s**.

Suites: Python 52 tests, .NET 44, all green; acceptance verdicts and all twelve production sample documents unchanged.

Refuted by verification, for the record: the P_AGENT/any-agent concern (by-design — conformance, not agent identity, decides), the `AppendOnce` round-trip worry, the claimed `E_FILE_HAS_HREF`/`FILE_NO_HREF` duplication (disjoint reach), and mutations-as-prose (a deliberate contract choice).

Issue #233, opened 2026-08-21
Deferred cleanups from the #188 step 3 reviews (DI extension, typed flags, MetsGraph lock, double walk)


tomcrane:
Four low-severity cleanups surfaced during the review rounds on #232 (issue #188 step 3) and deliberately deferred so as not to churn that PR right before release. None is load-bearing; all are safe to sweep in one follow-up PR after #232 ships.

### 1. `AddMetsManager` registration extension

Three services wire up the METS manager with the same copy-pasted pair of lines — `Configure<MetsManagerOptions>(FeatureFlags section)` + `AddSingleton<IMetsManager, MetsManager>()`:

- `Preservation.API/Program.cs` (~line 87)
- `DigitalPreservation.UI/Program.cs` (~line 96)
- `Pipeline.API/Program.cs` (~line 124)

Per the codebase's `Add*` convention (docs/adr/0000-service-clients.md), fold this into a single `services.AddMetsManager(configuration)` extension in `DigitalPreservation.Mets`, so a fourth consumer can't wire it up half-right.

### 2. Typed feature-flag options

The new flags are read as raw config strings at the point of use, with the magic strings repeated across code, tests and appsettings:

- `configuration.GetValue<bool>("FeatureFlags:EnableMetsIdNormalisation")` in `Preservation.API/Features/ImportJobs/ImportJobsController.cs` (~line 85)
- `FeatureFlags:ShowNormaliseMetsIds` looked up identically in two UI page models: `Pages/Deposits/Deposit.cshtml.cs` (~line 55) and `Pages/Deposits/ImportJobs/ImportJob.cshtml.cs` (~line 19)

A small bound options class (`IOptions<FeatureFlagOptions>`) would give compile-time safety against a typo'd key and one place to see all the flags.

### 3. `MetsGraph` plan-cache lock contention

`MetsGraph.PlanFor` (`DigitalPreservation.Mets/MetsGraph.cs`, ~line 146) takes a global `lock (PlansLock)` on **every** type lookup, even on a cache hit — and the walk calls `PlanFor` once per node, so a large METS takes that lock thousands of times per normalise. Not a safety bug (the lock is what makes it thread-safe); pure contention/overhead. Fix: replace `Dictionary` + lock with `ConcurrentDictionary.GetOrAdd`, making the hot path lock-free.

### 4. Double walk on the steady-state (conforming) path

[...]

PR #240, opened 2026-08-26
fix: suppression gates tolerate the platform's own scaffold folders (#188 step 3)

tomcrane:
## Problem

First METS ID migration trial on dev (2026-08-26) was refused by the suppression gate:

> suppressActivityStreamEvent is only for changes to how an object is recorded - a single METS patch - but it adds, deletes or renames content.

A deposit created against an Archival Group preserved **before LPII-9** gets `metadata/` and `metadata/ad-hoc/` written into its METS by `CreateDepositBase` (`GetDepositBase` after an export). So the diff for such a group is the METS patch **plus** `ContainersToAdd` for those two empty folders, and can never be a pure METS patch. Every pre-LPII-9 group would fail the same way, on dev and prod alike - neither survey counted that population.

## Fix

Both gates - `ImportJobsController.SuppressedButNotMetsOnly` and the migrator's `_refuse_unless_mets_only` - tolerate exactly `metadata` and `metadata/ad-hoc`, judged relative to the job's Archival Group, and nothing else in `ContainersToAdd`. Everything else (binaries add/delete/rename, container delete/rename, any other container add) is still refused. A job that names no Archival Group gets no allowance.

Rationale: those folders hold nothing, no consumer derives anything from them, and they would be added on the group's next preservation anyway. That is bookkeeping about how the object is recorded, not a change to what it holds - which is what suppression is for.

The alternative (a flag to stop `CreateDepositBase` scaffolding for METS-only deposits) was considered and rejected as a change to the LPII-133-sensitive deposit creation path.

## Tests

- `SuppressionGateTests`: scaffold-only passes; scaffold + another container refused; scaffold under a different object, or on a job with no Archival Group, refused. 8/8.
- `mets-id-migration/tests.py`: same cases for the tool's gate. 51/51.

Docs: migrator README and `docs/issues/188/issue-188-step3-migration.md` record the allowance.

Needs deploying to dev before the trial can be re-run.

### DIFF
--- a/src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
+++ b/src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
@@ -290,7 +290,7 @@ public class ImportJobsController(
         {
             return false;
         }
-        var agPathWithSlash = importJob.ArchivalGroup.LocalPath.TrimEnd('/') + "/";
+        var agPathWithSlash = importJob.ArchivalGroup.LocalPath.TrimEnd('/') + '/';
         if (!container.Id.LocalPath.StartsWith(agPathWithSlash))
         {
             return false;
--- a/src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
+++ b/src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
@@ -98,17 +98,6 @@ public class ImportJobsController(
             return this.StatusResponseFromResult(
                 Result.FailNotNull<ImportJobResult>(ErrorCodes.BadRequest, message));
         }
-        if (!IsPostedDiffReference(importJob, Request.Path)
-            && SuppressedButNotMetsOnly(importJob) is { } refusal)
-        {
-            // The feature flag says WHEN suppression may be used; this says WHAT FOR. Until now
-            // that rule lived only in the migration tool's client-side gate, which left a window
-            // between generating a diff and executing it - and left the UI checkbox trusting the
-            // operator to tick it only on the right kind of job. A diff reference is checked
-            // below instead, once its content exists.
-            return refusal;
-        }
-
         var depositResult = await mediator.Send(new GetDeposit(depositId), cancellationToken);
         if (depositResult.Failure)
         {
@@ -132,19 +121,26 @@ public class ImportJobsController(
                 importJob = diffImportJobResult.Value;
                 importJob.OriginalId = GetDiffUri(depositId);
                 importJob.SuppressActivityStreamEvent = suppressActivityStreamEvent;
-                if (SuppressedButNotMetsOnly(importJob) is { } diffRefusal)
-                {
-                    // A diff reference's content is only known now the diff has been generated -
-                    // and this is also what closes the window between a caller looking at a
-                    // METS-only diff and the deposit changing underneath them before they post it.
-                    return diffRefusal;
-                }
             }
             else
             {
                 logger.LogError("Unable to fetch diff import job for deposit {ErrorDetail}", diffImportJobResult.CodeAndMessage());
                 return this.StatusResponseFromResult(diffImportJobResult);
             }
+        }
+
+        if (SuppressedButNotMetsOnly(importJob, deposit.ArchivalGroup!) is { } refusal)
+        {
+            // The feature flag says WHEN suppression may be used; this says WHAT FOR. Until now
+            // that rule lived only in the migration tool's client-side gate, which left a window
+            // between generating a diff and executing it - and left the UI checkbox trusting the
+            // operator to tick it only on the right kind of job. It runs here, after the deposit
+            // is known, for two reasons: a diff reference's content only exists once the diff has
+            // been generated (which also closes the window between a caller looking at a
+            // METS-only diff and the deposit changing underneath them before they post it), and
+            // the scaffold-folder allowance is judged against the deposit's own Archival Group,
+            // never the one a caller-supplied job claims.
+            return refusal;
         }
 
         if (JobDoesNotBelongToDeposit(importJob, depositId, deposit) is { } mismatch)
@@ -235,10 +231,12 @@ public class ImportJobsController(
     /// a pure METS patch, and refusing it would leave every pre-LPII-9 group unmigrated. They
     /// hold nothing, no consumer derives anything from them, and they would be added on the
     /// group's next preservation anyway: that is bookkeeping about how the object is recorded,
-    /// not a change to what it holds. Nothing else in ContainersToAdd is tolerated.
+    /// not a change to what it holds. Nothing else in ContainersToAdd is tolerated, and the
+    /// folders are judged relative to <paramref name="archivalGroup"/> - the deposit's, which
+    /// the platform knows - not to whatever Archival Group a caller-supplied job claims.
     /// </para>
     /// </remarks>
-    private ActionResult? SuppressedButNotMetsOnly(ImportJob importJob)
+    private ActionResult? SuppressedButNotMetsOnly(ImportJob importJob, Uri archivalGroup)
     {
         if (!importJob.SuppressActivityStreamEvent)
         {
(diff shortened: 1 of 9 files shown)


src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
```diff
@@ -256,13 +271,30 @@ public async Task<IActionResult> GetImportJobResult([FromRoute] string depositId
...
         return this.StatusResponseFromResult(
             Result.FailNotNull<ImportJobResult>(ErrorCodes.BadRequest, message));
     }
 
+    /// <summary>
+    /// Whether a container to add is one of the platform's own scaffold folders (metadata,
+    /// metadata/ad-hoc), judged by its path relative to the job's Archival Group. A job that
+    /// does not say which Archival Group it is for cannot make that claim, so nothing qualifies.
+    /// </summary>
+    private static bool IsPlatformScaffoldFolder(ImportJob importJob, Container container)
```
brianmac1967: This judges a container-add safe by computing its path relative to `importJob.ArchivalGroup` — but that field is caller-supplied and never validated against the deposit's real Archival Group anywhere in this controller. `ExecuteImportJob` accepts a full caller-supplied `ImportJob` body when it isn't a diff reference, and `SuppressedButNotMetsOnly` runs before the deposit is even fetched; `JobDoesNotBelongToDeposit` only checks `importJob.Deposit` and the origin of the binaries, never `ArchivalGroup` or `ContainersToAdd` ids.

A caller with legitimate access to their own deposit could POST a body with a fabricated `ArchivalGroup` (real AG + `/objects`) and a `ContainersToAdd` id ending `.../objects/metadata` — this method computes the relative path as just `metadata`, matches `FolderNames.Metadata`, and the gate passes even though the real path isn't a genuine pre-LPII-9 scaffold folder. That creates a new container with no Activity Stream entry, defeating the invariant the doc comment above claims to enforce ("judged relative to the job's Archival Group").

src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
```diff
@@ -256,13 +271,30 @@ public async Task<IActionResult> GetImportJobResult([FromRoute] string depositId
...
+    {
+        if (importJob.ArchivalGroup is null || container.Id is null)
+        {
+            return false;
+        }
+        var relativePath = container.Id.LocalPath.GetPathBelow(importJob.ArchivalGroup.LocalPath)
```
brianmac1967: Minor: this re-implements the "strip AG prefix, then unescape" pattern that already exists twice in `GetDiffImportJob.cs` (lines 117-126 and 174: `agLocalPathWithSlash` + `.RemoveStart(...).UnEscapePathElementsNoHashes()`). A third near-identical implementation here means a future fix to the escaping/boundary logic (e.g. how BagIt `data/` prefixes or trailing slashes are handled) has to be found and applied in three places instead of one shared helper.

brianmac1967 (commented):
Automated review findings (Claude Code):

**Dry-run log message is now inaccurate for the case this PR allows** — `src/mets-id-migration/app/migrate.py:86` (not touched by this PR, so no diff line to anchor this to). Running `migrate --dry-run` against a pre-LPII-9 Archival Group now passes `_refuse_unless_mets_only` with `containersToAdd = [metadata, metadata/ad-hoc]`, but the log at line 86 still reads "the diff is a single METS patch, as required; not preserving" — misreporting what the dry run actually verified, right at the point an operator is reviewing trial output before a real migration. Worth updating that message (or making it conditional) to reflect the new scaffold-folder allowance.

src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
```diff
@@ -256,13 +269,37 @@ public async Task<IActionResult> GetImportJobResult([FromRoute] string depositId
...
+    /// metadata/ad-hoc), judged by its path relative to <paramref name="archivalGroup"/>.
+    /// </summary>
+    /// <remarks>
+    /// Deliberately literal: same scheme and host as the Archival Group, and the still-escaped
+    /// path below it must be exactly one of the two names. No unescaping (so a single segment
+    /// named <c>metadata%2Fad-hoc</c> is not two), and no BagIt <c>data/</c> stripping - the
+    /// repository never carries that prefix, WorkspaceManager makes the data folder the apparent
+    /// root - so this matches the migration tool's own gate exactly.
+    /// </remarks>
+    private static bool IsPlatformScaffoldFolder(Uri archivalGroup, Container container)
```
brianmac1967, line 292: The platform's scaffold-folder allowance is a hard-coded pair of literal path names here, duplicated independently as `SCAFFOLD_FOLDERS` in `migrate.py` with no shared source of truth between the two. If a third scaffold folder is ever introduced by `CreateDepositBase`/`GetDepositBase`, both gates and both test suites (`SuppressionGateTests.cs` and `tests.py`) need updating in lockstep by hand — miss one and the platform gate and the migration tool's client-side gate silently diverge (one refuses a diff the other would allow). The docs (`docs/issues/188/issue-188-step3-migration.md`) already acknowledge this and defer the real fix to issue #233 — flagging now since it's easy to lose track of once #233 isn't top of mind.

src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
```diff
@@ -146,6 +129,20 @@ public async Task<IActionResult> ExecuteImportJob([FromRoute] string depositId,
             }
         }
 
+        if (SuppressedButNotMetsOnly(importJob, deposit.ArchivalGroup!) is { } refusal)
```
brianmac1967, line 132: Worth a note on the tradeoff this move makes: previously a non-diff-reference job that would always be refused by this gate was rejected before any deposit lookup. Now it always costs a full `GetDeposit` + `ValidateDeposit` first — every malformed/malicious request is more expensive to reject than before. The updated test (`A_Suppressed_Job_That_Changes_Content_Is_Refused_Even_With_The_Flag_On`) reflects this — it now only asserts `ExecuteImportJob` wasn't sent, not that `GetDeposit` was never called. Likely a reasonable tradeoff given *why* it moved (judging the scaffold allowance against the deposit's real Archival Group), just flagging it as a deliberate cost worth being aware of rather than an oversight.

brianmac1967 (commented):
Re-review after the latest push. Good news: the spoofable-gate bug I flagged in the previous review looks fixed — `SuppressedButNotMetsOnly` now takes `deposit.ArchivalGroup` from the fetched deposit rather than trusting a caller-supplied field, and `IsPlatformScaffoldFolder` judges the container path against that. Traced it through and it's sound and well covered by tests. Two smaller things from this pass:

src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
tomcrane: Agreed, and this is a workaround rather than the fix. The doc paragraph ("The allowance is a workaround, not the fix", `docs/issues/188/issue-188-step3-migration.md`) and a comment on #233 (https://github.com/digirati-co-uk/digital-preservation/issues/233#issuecomment-5436947417) both record the real fix: a METS-only deposit should not scaffold folders into a METS it only patches (or `GetDiffImportJob` should skip METS-only empty directories, mirroring its LPII-133 tolerance). Once that lands, the pair goes from both gates and both test suites. Deferred so the migration campaign can run first.

src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
tomcrane: Yes, deliberate. A refused job now pays for `GetDeposit` + `ValidateDeposit` (both reads) before its 400; the flag-off refusal is still pre-fetch. Recorded in the code comment at this point (16727b2) and in the step 3 doc, which also notes the visible effect: an unknown/exporting/busy deposit reports 404/400/409 ahead of the what-for 400.

brianmac1967 (approved):
LGTM

PR #242, opened 2026-08-28
fix(mets-id-migration): a failed read before the deposit exists is not a verdict

tomcrane:
Found on the first dev dry run of the #188 step 3 migrator (started, as it turned out, without the VPN).

Every `GET /repository/{path}?view=mets` hit a 60 s connect timeout, and `migrate_all` recorded each Archival Group as `failed` — which removes it from the campaign, silently, for a reason that says nothing about the document. `survey` and `verify` already follow the rule that a failed read is not a verdict; `migrate` now applies it to the fingerprint read that precedes deposit creation: `ReadFailed` is logged as a warning and the row is left in whatever state it had. A METS that does not parse is still a verdict, as in `verify`.

The three rows this bit on `ledger-dev.sqlite` were reset by hand to `candidate`.

Python only: no image build. 52/52 tests.

### DIFF
--- a/src/mets-id-migration/app/migrate.py
+++ b/src/mets-id-migration/app/migrate.py
@@ -29,12 +29,22 @@ class MigrationRefused(RuntimeError):
     """The migration stopped on purpose, before changing anything preserved."""
 
 
+class ReadFailed(MigrationRefused):
+    """
+    The Archival Group's METS could not be read before anything was created. The same rule as
+    the survey's and verify's: a failed read is not a verdict. Recording it as FAILED would drop a
+    perfectly migratable group from the campaign because the VPN was off.
+    """
+
+
 def migrate_all(ledger: Ledger, candidates, dry_run: bool) -> None:
     for index, row in enumerate(candidates, start=1):
         path = row["path"]
         logger.info("[%s/%s] %s", index, len(candidates), path)
         try:
             migrate_one(ledger, path, dry_run)
+        except ReadFailed as error:
+            logger.warning("%s: leaving it %s for a rerun - %s", path, row["state"], error)
         except (MigrationRefused, api.ApiError, TimeoutError) as error:
             logger.error("%s: %s", path, error)
             ledger.record(path, FAILED, note=str(error))
@@ -182,9 +192,14 @@ def _is_mets_file(slug: str) -> bool:
 def _fingerprint(path: str) -> dict[str, str]:
     """Path to digest for every file the Archival Group's METS lists."""
     try:
-        return ids.file_digests(ids.parse(api.get_archival_group_mets(path)))
-    except (api.ApiError, etree.XMLSyntaxError) as error:
-        raise MigrationRefused(f"Could not read the Archival Group's METS: {error}") from error
+        mets_xml = api.get_archival_group_mets(path)
+    except api.ApiError as error:
+        raise ReadFailed(f"Could not read the Archival Group's METS: {error}") from error
+    try:
+        return ids.file_digests(ids.parse(mets_xml))
+    except etree.XMLSyntaxError as error:
+        # Unlike a failed read, a preserved METS that does not parse IS a verdict.
+        raise MigrationRefused(f"The Archival Group's METS is not well-formed XML: {error}") from error
 
 
 def _verify(before: dict[str, str], after: dict[str, str]) -> None:
--- a/src/mets-id-migration/app/migrate.py
+++ b/src/mets-id-migration/app/migrate.py
@@ -29,11 +29,15 @@ class MigrationRefused(RuntimeError):
     """The migration stopped on purpose, before changing anything preserved."""
 
 
-class ReadFailed(MigrationRefused):
+class ReadFailed(RuntimeError):
     """
     The Archival Group's METS could not be read before anything was created. The same rule as
     the survey's and verify's: a failed read is not a verdict. Recording it as FAILED would drop a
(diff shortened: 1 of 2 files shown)


src/mets-id-migration/app/migrate.py
```diff
@@ -182,9 +192,14 @@ def _is_mets_file(slug: str) -> bool:
 def _fingerprint(path: str) -> dict[str, str]:
     """Path to digest for every file the Archival Group's METS lists."""
     try:
-        return ids.file_digests(ids.parse(api.get_archival_group_mets(path)))
-    except (api.ApiError, etree.XMLSyntaxError) as error:
-        raise MigrationRefused(f"Could not read the Archival Group's METS: {error}") from error
+        mets_xml = api.get_archival_group_mets(path)
+    except api.ApiError as error:
+        raise ReadFailed(f"Could not read the Archival Group's METS: {error}") from error
```
brianmac1967: **Bug: the post-import verification read reuses `ReadFailed`, silently dropping ledger evidence for a completed migration.**

`_fingerprint()` is called from two places in `migrate_one`: the `before` read (line 56, before any deposit exists — the case this PR is fixing) and the `after` read (line 112, called *after* the import job has executed and returned `status == "completed"`, and after `disposable` was set to `False` so the deposit is deliberately kept as evidence).

Both calls go through this same `_fingerprint`, so a transient `api.ApiError` on the *second* call now also raises `ReadFailed` — but by that point the Archival Group has already been rewritten in Fedora/OCFL. `migrate_all`'s new handler for `ReadFailed` only logs a warning and does **not** call `ledger.record(...)`, whereas the old code raised plain `MigrationRefused` here, which *did* record `FAILED`.

Failure scenario: the import job completes successfully, then the verification re-read of the METS hits a network blip. The ledger row is silently left in its pre-migration state (e.g. still `CANDIDATE`), with no record that the Archival Group was actually migrated, no pointer to the kept-but-orphaned deposit for investigation, and a rerun of `migrate_all` will pick the same path up again as a fresh candidate against an Archival Group that's already been changed — losing the original run's rewrite counts/version-transition evidence.

The `ReadFailed` docstring itself says the read happens "before anything was created" — true only for the first call site. The fix should be scoped to that call site (e.g. only wrap the `before = _fingerprint(path)` call's `ApiError` as `ReadFailed`, leaving the `after` read raising `MigrationRefused` as before), not applied uniformly inside `_fingerprint` regardless of caller context.

src/mets-id-migration/app/migrate.py
```diff
@@ -29,12 +29,22 @@ class MigrationRefused(RuntimeError):
...
+    """
+
+
 def migrate_all(ledger: Ledger, candidates, dry_run: bool) -> None:
     for index, row in enumerate(candidates, start=1):
         path = row["path"]
         logger.info("[%s/%s] %s", index, len(candidates), path)
         try:
             migrate_one(ledger, path, dry_run)
+        except ReadFailed as error:
```
brianmac1967, line 50: **Minor/altitude: the ReadFailed vs. MigrationRefused split relies on subclass-and-ordering, and is scoped to the wrong unit (the shared `_fingerprint` helper) rather than the call site that actually justifies it.**

`except ReadFailed` only works here because it's listed *before* `except (MigrationRefused, api.ApiError, TimeoutError)` and `ReadFailed` subclasses `MigrationRefused`. That's correct today, but it's a silent ordering dependency — if someone later merges these clauses or reorders them (e.g. alphabetizing), `ReadFailed` would be swallowed by the general handler and the exemption this PR adds would quietly regress with no test failure pointing at why.

More fundamentally (see the sibling comment on line 197), the real fix belongs at the call site that has the "nothing created yet" precondition, not inside `_fingerprint` itself — `_fingerprint` can't know which of its two callers is invoking it, so raising `ReadFailed` unconditionally from within it is a special case bolted onto a shared helper rather than a fix scoped to where the precondition actually holds.

src/mets-id-migration/app/migrate.py
tomcrane: Confirmed, and fixed in 35b2618. The `ApiError` wrap now lives only at the `before = _fingerprint(path)` call site, where the "nothing created yet" precondition actually holds; `_fingerprint` lets `ApiError` propagate untouched, and its docstring now says why - only the caller knows whether anything has changed yet. A failed verification re-read after a completed import is once again caught by `migrate_all`'s general handler: recorded `FAILED` with the refusal note, deposit kept as evidence.

New regression test `test_a_failed_read_after_the_import_is_recorded_not_exempted` covers exactly your failure scenario: import completes, re-read raises, and the test asserts the row lands `FAILED` and the deposit is not deleted.

Worth noting the dev campaign happened to validate this behaviour by hand: the two groups whose runs were interrupted around the verification stage (`abcd4sqavv`, `abcdmspxcd`) were investigated via their kept deposits and corrected in the ledger - which only worked because the evidence trail existed. Silently dropping it would have hidden them.

src/mets-id-migration/app/migrate.py
tomcrane: Both points taken, fixed in 35b2618. `ReadFailed` now subclasses `RuntimeError` directly rather than `MigrationRefused`, so `migrate_all`'s two except clauses are disjoint and their order no longer carries the behaviour - merging or reordering them can't silently swallow the exemption. And with the wrap moved to the before-read call site (see the sibling thread), `ReadFailed` is raised only where its precondition holds; the class docstring now states that scoping explicitly.

PR #271, opened 2026-09-14
Refuse an Import Job whose archivalGroup is not the Deposit's (#267)

### LINKED ISSUE #267: A hand-written Import Job can name a different Archival Group from its Deposit, and runs against it
Found while verifying the documentation's custom-import-job walk-through.

`ImportJob.ArchivalGroup` carries the comment "The value must match the ArchivalGroup of the deposit". Nothing checks it.

`ImportJobsController.JobDoesNotBelongToDeposit` verifies that the job names a deposit, that it matches the route, and that every added or patched binary's `origin` is under `deposit.Files`. It never compares `importJob.ArchivalGroup` with the deposit's own. `ValidateDeposit` only requires the deposit to *have* one. `ExecuteImportJobHandler` then stores and executes against whatever the job names.

So a job POSTed to `/deposits/{a}/importjobs` whose `archivalGroup` names object **B** will run against B — creating or modifying a version of an object the deposit has nothing to do with — and report success. The binaries come from deposit `a`'s workspace, so the content is real; it simply lands in the wrong object.

A typo, or a copied-and-edited job body, is enough.

**Fix:** compare the two and refuse with a 400, as the neighbouring checks do.

## Related: two other documented requirements that are not enforced

Both have checks that exist but are only called by the UI:

- `ImportJob.AddedBinariesWithInvalidContentTypes()` — called from `ImportJob.cshtml.cs` only. On the API path a null `contentType` is silently dropped (`RequestX.WithContentType` skips the header when empty) and the binary is stored with whatever Fedora defaults to.
- `ImportJob.ItemsWithInvalidSlugs()` — likewise UI-only. The Storage API's `PreProcessValidateImportJob` checks only for a null `id` and a `#` in one.

Worth deciding whether these belong on the API path too, or whether the documentation should simply keep saying "supply it, nothing will stop you if you don't" — which is what it says now.

tomcrane:
Closes #267.

An Import Job's `archivalGroup` is meant to be redundant with the deposit's — the model's own comment says "the value must match the ArchivalGroup of the deposit" — but nothing checked it. `JobDoesNotBelongToDeposit` confirmed the job named the right deposit and that every added or patched binary's `origin` was under the deposit's workspace; it never compared `importJob.ArchivalGroup` with `deposit.ArchivalGroup`. So a hand-written job could name any group in the repository, and the platform would make a new version of it — with content from this deposit's workspace — and report success.

This adds the comparison to `JobDoesNotBelongToDeposit`, alongside the checks that already live there, answering `400` as they do:

- a job with **no** `archivalGroup` is refused ("must declare which Archival Group it is for"). Today it gets past the controller and fails on a null dereference downstream, so this turns a 500 into a 400;
- a job whose `archivalGroup` names a **different** group from the deposit's is refused, with both URIs in the message.

The comparison is by **repository path**, not by URI, so that a job naming the same group on the Storage API host, or with a trailing slash, is not refused for a difference that is not one. The path is compared **as escaped**, because that is the form the Storage API resolves in Fedora and `%` is a legal slug character: a one-segment `foo%2Fbar` and a two-segment `foo/bar` are different objects, and decoding first would have let a job for one through against a deposit for the other (review finding, fixed in f6858dd).

Diff-generated jobs are unaffected: `GetDiffImportJob` sets `archivalGroup` from the deposit, so they always match. The check runs after the suppression gate, so a suppressed job that would fail both still gets the gate's more specific message.

**Why now rather than later**

Today every authenticated caller may write anywhere, so the missing check was a footgun rather than an escalation. Once RFC-0001's per-caller roles and per-caller buckets land, it becomes the check that stops one caller's deposit landing in another caller's object. Better to have it in while it is one comparison and a `400`.

**Tests**

New `ArchivalGroupMatchTests`: different group refused; missing group refused; group that only matches once percent-decoded refused; same group executed; same group on the Storage host with a trailing slash executed.

One existing test changed: `SuppressionGateTests.What_The_Job_Claims_As_Its_Archival_Group_Does_Not_Matter_Only_The_Deposits` demonstrated that the scaffold-folder allowance is judged against the deposit's group by posting a job with **no** `archivalGroup` — which is now refused as malformed before it reaches execution. The test keeps its point, now made with a job that names the same group on a different host, and is renamed accordingly. Its sibling, which posts a job naming `<group>/objects`, is unchanged and still passes: the gate refuses it first.

**Not in this PR**

#267 also notes that `contentType` and slug validity on a hand-written job are checked only by the UI. Whether those belong on the API path is a separate decision; this PR does the one thing the issue is about.

https://claude.ai/code/session_01Jc8dHwb5uEaegAsZoBqMg


src/DigitalPreservation/Preservation.API/Features/ImportJobs/ImportJobsController.cs
```diff
@@ -306,6 +306,17 @@ private static bool IsPlatformScaffoldFolder(Uri archivalGroup, Container contai
...
     /// null when everything belongs.
     /// </summary>
+    /// <summary>
+    /// Compared by repository path rather than by URI, so that a job which names the group on the
+    /// Storage API host, or with different escaping, is not refused for a difference that is not one.
+    /// </summary>
+    private static bool SameArchivalGroup(Uri jobArchivalGroup, Uri? depositArchivalGroup) =>
+        depositArchivalGroup is not null &&
+        string.Equals(
+            jobArchivalGroup.GetPathUnderRoot(decode: true)?.TrimEnd('/'),
```
brianmac1967: **Potential ownership-check bypass via percent-encoding.**

`SameArchivalGroup` compares both URIs with `decode: true` (`GetPathUnderRoot(decode: true)`, which calls `WebUtility.UrlDecode`). But every downstream consumer that actually determines the write target uses the *undecoded* path:

- `Storage.API/Features/Import/Requests/ExecuteImportJob.cs:27` — `importJob.ArchivalGroup.GetPathUnderRoot()!` (default `decode: false`) is what's passed to `fedoraClient.GetValidatedArchivalGroupForImportJob(...)`, i.e. it picks the actual Fedora object to write to.
- `Storage.API/Features/Import/Requests/QueueImportJob.cs:74` and `Storage.API/Features/Import/ImportJobRunner.cs:25` likewise use the default (`decode: false`).

`%` is a legal slug character (`PreservedResource.ValidSlugChar`), so a single, legitimately-created slug segment such as `foo%2Fbar` decodes to `foo/bar`. That means a deposit whose real Archival Group is `.../cc/foo%2Fbar` (one segment) and a job that instead declares `.../cc/foo/bar` (two segments, a *different*, unrelated object) will compare equal here — `SameArchivalGroup` accepts the job — while the raw, undecoded path is what Storage API actually writes to. That reopens exactly the cross-object-write issue #267 was meant to close, just via percent-encoding instead of a bare mismatch.

The sibling method `IsPlatformScaffoldFolder` right above this one already documents the reasoning for staying undecoded ("No unescaping ... so a single segment named `metadata%2Fad-hoc` is not two") — this new check goes the other way and decodes, which seems like the source of the inconsistency.

Worth double-checking whether the comparison should use `decode: false` (matching what actually gets written), or whether downstream should be made to agree with `decode: true` instead.

tomcrane: You're right on every point: Storage resolves the undecoded path, % is a legal slug character, and the scaffold-folder check was already doing the right thing. Fixed in f6858dd by comparing the escaped path, with a test that fails on the old code (the job was executed) and passes now.

brianmac1967 (approved):
lgtm

PR #279, opened 2026-09-15
addressing security issues with pipeline APi as outlined in security advisory

brianmac1967:
addressing security issues with pipeline APi as outlined in security advisory [GHSA-v3wq-qw6c-r3q8](https://github.com/digirati-co-uk/digital-preservation/security/advisories/GHSA-v3wq-qw6c-r3q8)

Preservation UI Playwright tests ran against this branch and all passing.


tomcrane (commented):
Reviewed against the advisory. This closes all three items, including item 3 (the `#/../..` clean-up path), and the fixes are the ones the advisory asked for:

- **Item 1.** `DepositId` must be a valid slug, and `PathX.IsUnderRoot` refuses anything resolving outside the mount - both the rooted-value and `..` cases are tested. `Bash()` is gone; `DriveInfo` needs no process at all.
- **Item 2.** `FixedTimeEquals`, and a null configured key now fails closed, which also removes the `null == null` pass-through when no key is configured and no header is sent.
- **Item 3.** Clean-up and tool-output paths use the resolved deposit's slug; `POST /Pipeline` refuses a non-slug `depositName` before queueing; `IsUnderRoot` guards both deletes as defence in depth.

Two small things:

1. **Line 453 still passes the raw `request.DepositId` into `BagAndCheckObjectFileDigests`**, which builds `ProcessFolderBagit/{depositId}` from it and uses that as the `CopyDirectory` target and the BagIt process argument. Over HTTP the new slug check makes it harmless, and jobs from the Preservation API carry a minted id, so it's a consistency gap rather than a live hole - but it's the same pattern this PR fixes everywhere else. One line: pass `workspaceManager.DepositSlug`.

2. **`ValidSlug` admits `.` and `..`** (they're legal slug characters), so a `depositName` of `..` gets past the controller. It's safe because the deposit lookup then fails before the `try`, and because the clean-up guard exists - two layers, which is fine. A test asserting `..` is refused somewhere would stop a future change quietly removing one of them.

PR #280, opened 2026-09-15
UI: only return a signed-in user to a page on this site

tomcrane:
`AccountController.SignRefresh` takes a `path` to return the user to after they sign in again, and accepted it if it was a well-formed *relative* URI reference. That test is too weak: a reference beginning with two slashes is well-formed and relative, and a browser reads it as another host. The check is now `Url.IsLocalUrl`, which is the one ASP.NET provides for exactly this purpose: a single leading slash, nothing a browser would treat as a different origin. Anything else falls back to `/`, as before.

The UI's other redirects were checked while here; they all go to fixed paths or to a query string built from typed values, none to a caller-supplied target.

One line, no behaviour change for any legitimate link. No UI test project exists to add a test to.

From the September 2026 security re-review (item N10).


brianmac1967 (approved):
LGTM

PR #281, opened 2026-09-15
Keep a job's resources inside its Archival Group; dot segments out of METS paths and slugs

tomcrane:
Three related gaps from the September 2026 security re-review (items N3, 6.3 and N4 there). They are one idea, a name that can point somewhere other than where it appears to, closed at the three places such a name enters the system. No overlap with the advisory fork PR or with #280.

**A posted Import Job** (Preservation API). #271 made the job name the same Archival Group as its Deposit, but each binary and container id inside the job could still point anywhere in the repository, and the Storage API would write there. `JobDoesNotBelongToDeposit` now also requires every id, across all seven lists, to lie *strictly* below the job's Archival Group, compared by repository path as the #271 check is. Strictly, because a job may not name the group itself as a resource, which is what the slug case below used to produce. A resource with no id is refused too. Diff-generated jobs are unaffected.

**A METS document.** An `FLocat` href or `premis:originalName` became a `LocalPath` with no check, and for a METS-only entry that `LocalPath` becomes an S3 key to delete when tool output is refreshed. Paths in a METS are relative to the deposit root, so a dot segment (raw or percent-encoded) or a backslash is refused in `MetsParser` at the point the path is read, with a `NotSupportedException` in the same style as the parser's other "this METS is not usable" refusals. Nothing else about the value is judged: third-party METS carry absolute `https:` references and other shapes the parser already copes with, and those still pass.

**A slug.** `.` and `..` are made entirely of legal slug characters, but as a whole each is a dot segment. `ValidSlug` now refuses them and `MakeValidSlug` never produces them.

**Tests.** Six for the job check (below the group; another group; the group itself; a sibling sharing a name prefix; a container elsewhere; no id). Six for the parser, using the existing `mets-sample-001.xml` with one path corrupted at a time, plus proof the sample still parses as it is and an ordinary rename still passes. Ten for slugs. Every unit suite in the solution is green, including the Pipeline API tests that exercise `ValidSlug`.


brianmac1967 (commented):
## Code review

Two blocking issues, plus several correctness/reuse/efficiency notes. Most severe first.

### 1. `Pipeline.API/Middleware/ApiKeyMiddleware.cs:18` — empty configured API key accepts every request (blocking)

The auth-bypass guard checks `apiKey is null` but not `string.IsNullOrEmpty(apiKey)`. If `ApiKeyOptions.ApiKey` is ever bound to an empty string (e.g. a blank env var), the guard doesn't trip. A request with no `X-API-KEY` header (or an empty one) produces `extractedApiKey.ToString() == ""`, and `CryptographicOperations.FixedTimeEquals(emptyBytes, emptyBytes)` returns `true` — every unauthenticated request gets accepted.

### 2. `Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs:166` — uncaught METS-parse exception leaves the job stuck "Running" (blocking)

`GetWorkspaceManager` — which now triggers the new `MetsParser.RejectDotSegments` — is called *before* `Handle`'s try/catch/finally block starts (line 175). A METS with a dot-segment or backslash path (newly rejected by `MetsParser.cs:42-50`) throws `NotSupportedException` here, uncaught by this handler, and propagates up to `PipelineJobRunner.Execute`'s generic catch, which just logs and returns — never calling `UpdateJobStatus`. The pipeline job is left permanently stuck in "Running" with no failure ever recorded, and the `finally` block's cleanup/lock-release never runs either.

### 3. `DigitalPreservation.Mets/MetsParser.cs:44` — legitimate backslash in a filename now fails ingest

`RejectDotSegments` rejects any backslash appearing anywhere in the path string, not just as a directory separator. A third-party METS with a `premis:originalName`/`FLocat` href like `objects/AC\DC.mp3` (a valid, if unusual, filename character) now throws `NotSupportedException`, aborting the whole METS parse — and per finding #2, that failure is silent, not surfaced.

### 4. `DigitalPreservation.Mets/MetsParser.cs:45` — dot-segment check misses double-encoded traversal

The check unescapes each path segment only once, so a double-percent-encoded dot segment (`%252e%252e` → `%2e%2e`, not `..`) isn't recognized. Verification found no current second-decode consumer downstream, so not exploitable today — but the guard doesn't fully deliver on its stated purpose (a path can't climb out of the deposit) against nested encoding.

### 5. `Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs:1314` — missing null guard can collapse path to the shared root

`BagAndCheckObjectFileDigests` takes a nullable `depositId` (`workspaceManager.DepositSlug`) but, unlike its sibling `CleanupProcessFolder`/`CleanupBagitProcessFolder` added in this same PR, has no null/empty guard before building a path from it. If `DepositSlug` is ever null, `processFolderBagitDeposit` collapses to just the shared `ProcessFolderBagit` root rather than a per-deposit subfolder, and the BagIt process then operates directly on that shared root — risking file collisions across concurrent jobs instead of failing cleanly like the sibling methods now do.

### 6. `Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs:375` — write paths lack the guard delete paths got

The new `PathX.IsUnderRoot` defence-in-depth check was added only to the delete/cleanup paths; the write paths built from the same `workspaceManager.DepositSlug` (`ExecuteBrunnhilde`'s `metadataPathForProcessFilesAndDirectories`, and `processFolderBagitDeposit`) get no equivalent check — leaving the riskier (write) operations less defended than the safer (delete) ones.

### Minor

- `DigitalPreservation.Common.Model/PreservedResource.cs:90` — dot-segment/path-containment rejection is now independently reimplemented three times in this PR (`ValidSlug`, `MetsParser.RejectDotSegments`, `PathX.IsUnderRoot`) with no shared helper, and the three already disagree in strength — a future fix to one won't propagate to the others.
- `Preservation.API/Features/ImportJobs/ImportJobsController.cs:372` — `JobDoesNotBelongToDeposit` calls `ResourceIds(importJob)` (a 7-way `Select().Concat()` over every binary/container) twice instead of once; wasted allocation/iteration that scales with job size.

tomcrane (commented):
Thanks. Items 1, 2, 5 and 6 are Pipeline API defects that this PR's new parser refusal led you to, and they are real, the first two especially; they are fixed in their own PR (#283) so that this one and that one merge independently. On this PR:

**3, backslash in a filename - kept, deliberately.** A backslash can be neither a slug character nor part of a path the platform will ever build, so a METS naming `objects/AC\DC.mp3` describes a file the deposit cannot hold under that name. Refusing it loudly is better than the alternative, which is a METS entry that silently never matches a file. And `System.Uri` turns a backslash into a slash for some schemes, which is exactly the class of ambiguity this check exists to remove. With item 2 fixed the refusal is now recorded against the job rather than lost.

**4, double encoding - by design, and stated as such.** The rule decodes once because that is what every consumer does: a METS path becomes an S3 key literally, and `System.Uri` decodes `%2e` once. `%252e%252e` decodes to `%2e%2e`, which is not a dot segment to anything downstream, so refusing it would refuse a string that cannot traverse. The adversarial review of the advisory fix checked the same question against Kestrel and `Uri` and found no second decode anywhere.

**Minor 1, three implementations - now one.** `UriPathX.IsDotSegment` is the single definition (raw or once-decoded `.` or `..`), used by `IsTraversalSegment`, by the parser's `RejectDotSegments`, and by both slug methods. That also closed a gap you did not name: a slug of `%2e%2e` was accepted and is now refused, and `MakeValidSlug` never produces one. `PathX.IsUnderRoot` stays separate because it judges a resolved filesystem path, not a segment, and its doc comment says so.

**Minor 2** - the id list is built once.

brianmac1967:
From co-pilot:
PR review summary

The PR addresses a genuine security issue and adds useful defence in depth at three boundaries: slug creation, METS parsing, and import-job execution. However, I would request changes because the METS validation can still be bypassed using percent-encoded path separators.

Blocking finding
1. Encoded separators can bypass RejectDotSegments

RejectDotSegments splits the original string on literal / characters and applies IsDotSegment to each resulting segment:

if (path.Contains('\\') || path.Split('/').Any(UriPathX.IsDotSegment))


IsDotSegment only returns true when the entire decoded segment equals . or ..:

public static bool IsDotSegment(string segment) =>
    segment is "." or ".." || Uri.UnescapeDataString(segment) is "." or "..";


That catches:

objects/%2e%2e/x.jpg


But it does not catch values where the separator is encoded inside the same segment, for example:

objects%2f..%2fother-deposit/x.jpg
objects/%2e%2e%2fother-deposit/x.jpg
objects/%2e%2e%5cother-deposit/x.jpg


After decoding, those can become:

objects/../other-deposit/x.jpg
objects/../other-deposit/x.jpg
objects/..\other-deposit/x.jpg


Whether exploitation occurs depends on how downstream code decodes and normalises the path, but this is exactly the kind of alternate representation that path-boundary validation should reject.

There is already a method intended to detect redirecting characters after decoding:

public static bool IsTraversalSegment(string segment)


I suggest using that method for METS path segments rather than the narrower IsDotSegment:

private static void RejectTraversalSegments(string path, string what)
{
    if (path.Split('/').Any(UriPathX.IsTraversalSegment))
    {
        throw new NotSupportedException(
            $"{what} '{path}' contains a traversal segment or invalid path separator; " +
            "METS paths must be relative to the deposit root.");
    }
}


However, this still needs careful testing because a fully encoded path such as objects%2f..%2fx.jpg is presented as one segment. IsTraversalSegment should reject it because the decoded value contains /.

Suggested GitHub review comment

Blocking: This validation only splits on literal / and IsDotSegment only rejects a segment whose complete decoded value is . or ... An input such as objects/%2e%2e%2fother-deposit/x.jpg, objects%2f..%2fx.jpg, or an encoded backslash can survive this check and become a traversal path if it is decoded downstream. Could we use UriPathX.IsTraversalSegment here, or decode once and validate the resulting separators and segments, with tests covering %2f, %2F, %5c, and %5C?

Suggested tests:

[Theory]
[InlineData("objects/%2e%2e%2fother-deposit/x.jpg")]
[InlineData("objects/%2E%2E%2Fother-deposit/x.jpg")]
[InlineData("objects%2f..%2fx.jpg")]
[InlineData("objects/%2e%2e%5cother-deposit/x.jpg")]
[InlineData("objects%5c..%5cx.jpg")]
public void An_Encoded_Traversal_Path_Is_Refused(string href)
{
    var doc = Sample();
    var flocat = doc.Descendants(Mets + "FLocat").First();
    flocat.SetAttributeValue(XLink + "href", href);

    var act = () => Parser().GetMetsFileWrapperFromXDocument(MetsUri, doc);

    act.Should()
        .Throw<NotSupportedException>()
        .WithMessage("*may not climb out of it*");
}

Important non-blocking findings
2. Containment deliberately ignores scheme and authority

IsWithinArchivalGroup compares only paths:

var groupPath = archivalGroup.GetPathUnderRoot()?.TrimEnd('/') + "/";
return resourceId.GetPathUnderRoot()
    ?.StartsWith(groupPath, StringComparison.Ordinal) == true;


This means a resource URI on another scheme or host may pass if its path matches the archival-group path:

Archival group:
https://preservation.test/repository/cc/thing

Potential accepted resource:
https://attacker.test/repository/cc/thing/objects/x.tif


The comment says this is intentional:

// compared by repository path ... so the host does not matter


That may be compatible with existing repository-ID semantics, but it deserves explicit verification because this method is enforcing an authorization boundary. If the full resource URI is subsequently sent to the Storage API, accepting a different authority could have effects beyond simple repository identity comparison.

At minimum, add a test documenting the intended behaviour. Preferably require compatible schemes and authorities unless environment portability genuinely requires host-independent identifiers.

Suggested GitHub review comment

IsWithinArchivalGroup intentionally ignores the URI scheme and authority. Since this is now an authorization boundary, can we confirm that a resource ID on another host but with the same repository path is safe? I would add a test for this case. If cross-host IDs are not explicitly required, compare scheme/authority as well as the normalised path.

3. IsDotSegment should be resilient to malformed escaped input

Uri.UnescapeDataString(segment) is now called from more externally reachable validation paths, including slugs and METS content.

The method should ideally behave as a validator and return false, or expose a safe validation result, rather than allowing malformed external input to generate an unexpected exception. Exactly how malformed % sequences behave can vary by input shape, so add tests for values such as:

%
%2
%GG
..%
%2e%


A defensive implementation could isolate decoding:

public static bool IsDotSegment(string segment)
{
    if (segment is "." or "..")
    {
        return true;
    }

    try
    {
        return Uri.UnescapeDataString(segment) is "." or "..";
    }
    catch (UriFormatException)
    {
        return false;
    }
}


For security-sensitive paths, malformed encoding may instead deserve explicit rejection. The key point is to make that policy deliberate and tested.

4. Add tests for premis:originalName

The production code validates both:

RejectDotSegments(originalName, "premis:originalName");
RejectDotSegments(flocat, "mets:FLocat xlink:href");


But the new parser tests exercise only FLocat. A future refactor could accidentally remove the originalName check without a regression failure.

Add equivalent tests that modify premis:originalName, including a directory entry because this value feeds FindDirectory:

[Theory]
[InlineData("../outside")]
[InlineData("objects/../../outside")]
[InlineData("objects/%2e%2e/outside")]
[InlineData("objects\\outside")]
public void An_OriginalName_That_Could_Leave_The_Deposit_Is_Refused(string name)
{
    var doc = Sample();
    var originalName = doc
        .Descendants()
        .First(e => e.Name.LocalName == "originalName");

    originalName.Value = name;

    var act = () => Parser().GetMetsFileWrapperFromXDocument(MetsUri, doc);

    act.Should().Throw<NotSupportedException>();
}

5. Missing execution assertions in several refusal tests

The first two rejection tests verify that the mediator does not execute the job:

Executed(mediator).MustNotHaveHappened();


These tests do not:

A_Sibling_Group_Sharing_A_Name_Prefix_Is_Refused
A_Resource_With_No_Id_Is_Refused

The returned 400 currently implies early termination, but explicitly asserting that ExecuteImportJob was not sent makes the security expectation clearer and prevents controller-flow regressions.

Refusal(result).Should().Contain("not within its Archival Group");
Executed(mediator).MustNotHaveHappened();

6. Duplicate XML documentation block

Immediately before BuildLookupMaps, the existing documentation and the new documentation produce two consecutive <summary> elements. The new documentation actually describes RejectDotSegments, so it should sit directly above that method without the earlier BuildLookupMaps summary preceding it.

Suggested structure:

/// <summary>
/// Refuses METS paths containing traversal segments.
/// </summary>
private static void RejectDotSegments(string path, string what)
{
    // ...
}

/// <summary>
/// Builds lookup dictionaries for efficient O(1) access to METS elements by ID.
/// </summary>
private static MetsLookupMaps BuildLookupMaps(
    XDocument xMets,
    XElement physicalStructMap)
{
    // ...
}

7. Large METS fixture increases maintenance cost

Adding a genuine parseable sample is sensible, but the 901-line fixture makes the security tests heavier and may contain unrelated metadata. A smaller purpose-built METS fixture would make it clearer which minimum structures the parser requires and reduce test-data maintenance.

This is not a reason to block the fix, especially if the existing sample is valuable as an integration fixture. Consider renaming the tests as integration/parser regression tests if retaining the complete document.

Positive observations
Rejecting . and .. slugs closes the direct parent-resolution problem.
MakeValidSlug has explicit regression coverage for encoded dot segments.
The archival-group containment comparison includes a trailing /, correctly preventing the common prefix mistake where thing2 is treated as a child of thing.
The import-job checks cover binary and container collections rather than only additions.
Null resource IDs are rejected before the null-forgiving operator is used.
The PR adds defence in depth rather than relying on one validation point.
The tests clearly explain the security scenario and expected boundary.
Recommendation

Request changes, based on the encoded-separator bypass in RejectDotSegments.

After that is resolved, I would treat the cross-host URI behaviour as the next most important design decision. The remaining points are worthwhile test coverage and maintainability improvements rather than blockers.

tomcrane:
On the Copilot review - taken 1 to 6, declined 7.

**1, encoded separators.** Not exploitable as it stood: nothing downstream decodes a METS path (an S3 key is literal, and `Uri` never treats `%2f` as a separator), which is also why the review's own suggestion of `IsTraversalSegment` would have been wrong here - it refuses `:`, which every absolute `https:` href in a third-party METS contains. But the principle is right: a boundary check should refuse an alternate spelling rather than reason about what would happen to it. `UriPathX.ContainsEncodedSeparator` is the rule, and the parser and both slug methods use it. All five of the suggested inputs are tests and are refused.

**2, cross-host ids.** Deliberate, and now pinned with a test that says why: ids for the same resource may carry the Storage or the Preservation host (which is why #271 compares by path), and the Storage API resolves an id by path alone, so the host changes nothing about where a resource is written.

**3, malformed escapes.** `Uri.UnescapeDataString` leaves a malformed sequence as it is and does not throw; six cases pinned, none counts as a dot segment.

**4, `premis:originalName`** now has its own parser tests, including a directory entry. **5**, both refusal tests assert non-execution. **6**, the parser's doc comments sit above their own methods again - that one was mine.

**7, the fixture.** Kept. A parser test against a real METS proves the parser still reads real METS after the change, which a minimal fixture would not.

brianmac1967 (approved):
LGTM

PR #282, opened 2026-09-16
Storage API: every resource in an Import Job must lie below its Archival Group

tomcrane:
The Storage API's own copy of the check that #281 adds to the Preservation API: every binary and container id in an Import Job must lie *strictly* below the job's Archival Group, compared by repository path so the host does not matter. A job that declared the right group but named a resource in another would otherwise have been written there, since nothing at this layer compared the two. Strictly below, so the group itself cannot be named as a resource.

It goes into `PreProcessValidateImportJob`, which the advisory fix (6984e3e) made the single pre-check for a job's paths, so a bad job is refused with a 400 by the queue handler and again before the transaction is opened by the executor. Diff-generated jobs are unaffected. This was held out of the advisory fork PR only to avoid a conflict; it is item 5 of the September security re-review's action list.

Tests: another group refused; a sibling sharing a name prefix refused; the group itself refused; the same group on a different host accepted. Storage API suite green.


brianmac1967 (approved):
LGTM and co-pilot

PR #283, opened 2026-09-16
Pipeline API: an empty configured key fails closed; a job whose workspace cannot be resolved is recorded as failed

tomcrane:
Four Pipeline API items from the review of #281, which followed the new METS parser refusal into this service and found what was waiting there. Separate from #281 so each merges on its own.

1. **An empty configured API key accepted every request.** `ApiKeyMiddleware` guarded against a *null* configured key but not an empty one; with `ApiKey` bound to `""` (a blank environment variable, say), an absent header compared equal and every request passed. The guard is `IsNullOrEmpty` now. Three test cases.
2. **A job whose workspace could not be resolved stayed "Running" for ever.** The handler claims the job as Running before it resolves the deposit's workspace. If that failed, it returned without recording anything; if it threw (which a METS the parser now refuses would do), the exception escaped the handler and the runner just logged it. Either way the job never reached a terminal state, and the queue message was already gone. Resolution is now inside a `try`, and both the failed-result and the exception paths record `completedWithErrors` before returning. Two tests.
3. **The tool-output write path** gets the same `PathX.IsUnderRoot` guard the clean-up paths were given in #279.
4. **The BagIt step** refuses to run with no deposit slug, rather than collapsing its working path onto the shared BagIt root where concurrent jobs would bag over each other.

Pipeline API suite green.


brianmac1967:
From co-pilot

PR review summary

This is a useful reliability and security hardening change. The empty API-key fix is sound, and recording workspace-resolution failures avoids leaving claimed jobs permanently Running. I would request changes, primarily because the broad exception handler also swallows cancellation and because raw exception messages are persisted and returned.

The supplied diff is the primary basis for this review. I found internal context confirming that Pipeline API operation and API error investigation are active concerns, but it does not provide additional implementation detail for these changed methods. Steps to run Pipeline API locally.docx confirms the solution is normally exercised with the Storage, Preservation, UI, and Pipeline APIs running together.

Blocking findings
1. Do not convert request cancellation into a failed job

This catches every exception:

catch (Exception e)
{
    logger.LogError(e, ...);
    workspaceResult = Result.FailNotNull<WorkspaceManager>(
        ErrorCodes.UnknownError,
        e.Message);
}


GetWorkspaceManager(...) receives the request’s CancellationToken, so it may legitimately throw OperationCanceledException when the request, host, or worker is shutting down.

The new catch block would:

Swallow the cancellation.
Log it as an unexpected error.
Attempt another remote call through UpdateJobStatus.
Return an ordinary failure instead of preserving cancellation semantics.

That can complicate graceful shutdown and may incorrectly mark work as CompletedWithErrors when it was interrupted rather than actually processed and failed.

Use a filtered catch:

try
{
    workspaceResult = await GetWorkspaceManager(
        request,
        true,
        cancellationToken);
}
catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
{
    throw;
}
catch (Exception exception)
{
    logger.LogError(
        exception,
        "Could not resolve the workspace for pipeline job {JobIdentifier}, deposit {DepositId}",
        request.JobIdentifier,
        request.DepositId);

    workspaceResult = Result.FailNotNull<WorkspaceManager>(
        ErrorCodes.UnknownError,
        "Workspace resolution failed.");
}


Alternatively:

catch (Exception exception) when (exception is not OperationCanceledException)


The first version is more precise because an OperationCanceledException thrown without the supplied token being cancelled may represent an unexpected dependency failure.

Suggested GitHub comment

Blocking: This catches OperationCanceledException from GetWorkspaceManager and converts request or host cancellation into CompletedWithErrors. Could we rethrow when cancellationToken.IsCancellationRequested and only translate unexpected exceptions? That preserves cancellation semantics and avoids making another API call during shutdown.

Add a regression test:

[Fact]
public async Task Cancellation_Is_Not_Converted_Into_A_Failed_Job()
{
    using var cancellation = new CancellationTokenSource();
    cancellation.Cancel();

    A.CallTo(() => preservationApiClient.GetDeposit(
            DepositId,
            cancellation.Token))
        .Throws(new OperationCanceledException(cancellation.Token));

    var act = () => CreateHandler().Handle(
        new ExecutePipelineJob(JobId, DepositId, "tester"),
        cancellation.Token);

    await act.Should().ThrowAsync<OperationCanceledException>();

    A.CallTo(() => preservationApiClient.LogPipelineRunStatus(
            A<PipelineDeposit>.That.Matches(
                d => d.Status == PipelineJobStates.CompletedWithErrors),
            A<CancellationToken>._))
        .MustNotHaveHappened();
}

2. Raw exception messages should not become job status or API output

The exception handler stores e.Message:

workspaceResult = Result.FailNotNull<WorkspaceManager>(
    ErrorCodes.UnknownError,
    e.Message);


That is then incorporated into:

var message =
    $"Could not process pipeline job for job id {request.JobIdentifier} and deposit {request.DepositId}: "
    + workspaceResult.ErrorMessage;


This message is both passed to UpdateJobStatus and returned from the handler.

Exceptions produced during workspace creation, METS loading, S3 access, filesystem handling, XML parsing, or HTTP calls may include:

Internal filesystem paths
Bucket or object names
Internal service addresses
Malformed user-controlled METS content
Dependency implementation details

The structured error log already captures the full exception for operational diagnosis. The persisted and returned result should normally use a controlled message.

For example:

catch (Exception exception)
{
    logger.LogError(
        exception,
        "Could not resolve the workspace for pipeline job {JobIdentifier}, deposit {DepositId}",
        request.JobIdentifier,
        request.DepositId);

    workspaceResult = Result.FailNotNull<WorkspaceManager>(
        ErrorCodes.UnknownError,
        "The deposit workspace could not be resolved.");
}


If known Result failures are already sanitised and useful to clients, preserve their messages. Only replace messages originating from caught exceptions.

Suggested GitHub comment

Blocking/security: The caught exception’s raw Message is persisted through UpdateJobStatus and returned in the failure result. Workspace resolution crosses METS, storage, filesystem, and HTTP boundaries, so exception text may expose internal paths or dependency details. Could we log the complete exception but use a controlled external message such as "The deposit workspace could not be resolved"?

Important non-blocking findings
3. The second exception test does not assert the returned result

This test verifies that the handler does not throw and that the final status was recorded:

await act.Should().NotThrowAsync();


It does not establish that the returned Result is a failure or that its error code is correct. A future regression could return success while still satisfying this test.

Prefer capturing the result:

var result = await CreateHandler().Handle(
    new ExecutePipelineJob(JobId, DepositId, "tester"),
    CancellationToken.None);

result.Failure.Should().BeTrue();
result.ErrorCode.Should().Be(ErrorCodes.UnknownError);


You can then retain the status assertion.

Suggested GitHub comment

Could this test assert the returned Result as well as NotThrowAsync? At present, it would still pass if the handler accidentally returned success after logging CompletedWithErrors.

4. Verify both status transitions and their ordering

The tests assert that CompletedWithErrors is written once, but they do not explicitly prove that:

Running was recorded first.
CompletedWithErrors was recorded second.
No later status overwrote the failure.

Because the defect specifically concerns a claimed job being left Running, ordering is part of the behaviour being fixed.

FakeItEasy can verify ordered assertions:

using (OrderedAssertionsContext _ = Fake.CreateOrderedScope())
{
    A.CallTo(() => preservationApiClient.LogPipelineRunStatus(
            A<PipelineDeposit>.That.Matches(
                d => d.Status == PipelineJobStates.Running),
            A<CancellationToken>._))
        .MustHaveHappenedOnceExactly();

    A.CallTo(() => preservationApiClient.LogPipelineRunStatus(
            A<PipelineDeposit>.That.Matches(
                d => d.Status == PipelineJobStates.CompletedWithErrors),
            A<CancellationToken>._))
        .MustHaveHappenedOnceExactly();
}


The exact FakeItEasy ordered-scope syntax should be aligned with the package version already used by the project.

5. Decide what happens if recording the failure itself fails

The intended guarantee is:

The job must be recorded as CompletedWithErrors, not left Running.

But this call can itself fail:

await UpdateJobStatus(
    request,
    PipelineJobStates.CompletedWithErrors,
    message,
    cancellationToken);


If Preservation API communication fails, the job may still remain Running, and this invocation could throw or replace the original workspace failure.

The PR cannot guarantee persistence during an unavailable dependency, but the behaviour should be explicit. Consider:

Logging a distinct high-severity error if failure-state persistence fails.
Preserving the original workspace failure where possible.
Ensuring the queue/retry policy can retry status persistence safely.
Adding a test for UpdateJobStatus failure.

I would not necessarily block this PR on a full retry mechanism, but the code comment currently promises more than the implementation can guarantee.

6. Confirm whether TryReleaseLock causes duplicate lock release

The new path guard does this inside ExecuteBrunnhilde:

await TryReleaseLock(
    request,
    workspaceManager.Deposit,
    cancellationToken);

return new ProcessPipelineResult
{
    Status = PipelineJobStates.CompletedWithErrors,
    ...
};


Without the full calling flow, I cannot confirm whether the caller also releases the lock after receiving a failed ProcessPipelineResult. If it does, this introduces a duplicate-release attempt.

That may be harmless if lock release is idempotent, but it deserves checking. A cleaner ownership model is usually for the layer that acquires a lock to release it in a finally block rather than having a nested operation selectively release it.

Suggested GitHub comment

Can we confirm that the caller does not also release the deposit lock after this failed result? If it does, this path invokes release twice. Ideally, lock ownership and release would remain with the method that acquired it, preferably through finally.

7. Add targeted tests for both new path guards

The PR adds two security guards but does not include direct tests for either:

PathX.IsUnderRoot(
    processFolder,
    metadataPathForProcessFilesAndDirectories)


and:

PathX.IsUnderRoot(
    pipelineToolOptions.Value.ProcessFolderBagit,
    processFolderBagitDeposit)


Add coverage for:

Empty slug or deposit ID
..
Absolute child replacement
Sibling-prefix paths such as /process2
Alternate separators
A legitimate child path
Trailing separator differences

For example, the tests should prove that:

Root:       /pipeline/process
Accepted:   /pipeline/process/abc123/metadata
Rejected:   /pipeline/process
Rejected:   /pipeline/process2/abc123
Rejected:   /pipeline/process/../outside


This is especially important if DirectorySeparator can differ from the runtime OS separator. The string is assembled using the configured separator, while PathX.IsUnderRoot may use System.IO.Path semantics for the host operating system.

8. Whitespace-only API-key configuration remains accepted

The middleware now rejects null and "":

string.IsNullOrEmpty(apiKey)


A value containing only spaces is still treated as configured and could authenticate a matching whitespace header.

Secrets containing leading or trailing spaces can theoretically be intentional, so silently trimming keys is not always safe. However, a whitespace-only key is almost certainly misconfiguration. Consider:

string.IsNullOrWhiteSpace(apiKey)


Add:

[InlineData(" ")]
[InlineData("   ")]
[InlineData("\t")]


Whether this is required depends on the configuration contract, so I would treat it as non-blocking.

Positive observations
Rejecting an empty configured API key closes a real fail-open authentication case.
The API-key tests cover missing, empty, and genuine supplied header values.
The workspace failure change addresses the inconsistency between a prior Running claim and an early return.
The handler logs exceptions with structured job and deposit identifiers.
Both path guards require strict descent beneath a per-deposit root rather than relying only on a trusted slug.
The BagIt guard also protects concurrent jobs from accidentally sharing the root workspace.
The new tests explain the operational reason behind the change rather than testing only implementation mechanics.

Recommendation

Request changes for:

Preserving OperationCanceledException semantics.
Preventing raw exception text from being persisted or returned.

I would also strongly request tests for the new PathX.IsUnderRoot call sites before approval. The remaining items can be addressed in this PR or recorded as focused follow-up work.

brianmac1967 (approved):
LGTM (and co-pilot)

brianmac1967 (changes requested):
approved by error

tomcrane:
Taken 1 to 5 and 8; answered 6 and 7.

**1, cancellation.** Rethrown when the request's token is cancelled, before the generic catch, so an interrupted job is neither recorded as failed nor the cause of another remote call during shutdown. The suggested regression test is in, as written.

**2, raw exception text.** The exception is logged in full; what is recorded and returned is "the deposit workspace could not be resolved (see the Pipeline API log)". A failed `Result` from `GetWorkspaceManager` keeps its own message, since those are already controlled. The test feeds an exception carrying an internal path and asserts it does not come back.

**3, 4.** Every test now asserts the returned result and its error code; the first asserts the order - Running recorded, then completedWithErrors, and nothing after - with FakeItEasy's `Then`.

**5, recording the failure fails.** Made explicit rather than promised: if the Preservation API cannot record the failure, that is logged at Critical, the original failure is still returned, and the comment says the job does then stay Running because there is no queue message left to retry from. A test drives the recording call to throw and checks nothing escapes and the real failure comes back.

**6, double lock release.** Checked: `Handle` releases the lock only in its exception catch; every failing branch inside `ExecuteBrunnhilde` (mount path missing, object folder missing, and now this guard) releases it itself before returning its result, and `Handle` does not release again on a failed result. So the new guard follows the established pattern and there is no duplicate. Moving ownership of the lock to a single `finally` is a fair ambition but a larger change than this PR.

**7, call-site tests for the guards.** `PathX.IsUnderRoot` has its own tests from #279 covering the root itself, a child, a nested descendant, traversal above, a rooted path elsewhere, and a sibling sharing a name prefix, which is the list asked for. Exercising the two call sites means standing up `ExecuteBrunnhilde` with a real mount path and process folders on disk; that is an integration test the project does not yet have a harness for, so it is not in this PR. On the separator point: the paths are built with the configured `DirectorySeparator`, and `Path.GetFullPath` normalises either separator on Windows and treats `/` as the only one on Linux, where this runs.

**8, whitespace-only keys.** `IsNullOrWhiteSpace`; four blank shapes tested against an absent, empty, identical and real supplied header.

brianmac1967 (approved):

PR #284, opened 2026-09-17
RFC-0001 step 2 mechanics: ResourceUri on AccessTokenProvider

tomcrane:
# RFC-0001 step 2 mechanics: `ResourceUri` on `AccessTokenProvider`

The follow-up PR the [landing sequence](../docs/rfc-0001-landing-sequence.md) schedules after #208 —
"bring across what is worth keeping from the LPII-166 branch in corrected form". Investigation
reduced its scope: the landing sequence's items 1–2 (a guarded `PostConfigure<JwtBearerOptions>`
hoist and the config-location decision) turn out to be **already delivered by #208** —
`AudienceValidationTests` proves `AzureAd:TokenValidationParameters:ValidAudiences` binds through the
stock `AddMicrosoftIdentityWebApi` path both APIs use, so the dual-audience mechanism is pure config
with no code of its own. This PR is item 3 alone, plus one new characterisation pin.

## What it does

**`ResourceUri` (optional) on `IAccessTokenProviderOptions`** — RFC-0001 Phase 2 option (a), with
both defects found on the LPII-166 branch fixed (comparison doc §6):

- **Acquisition is gated on the credential triplet only** (`TenantId`/`ClientId`/`ClientSecret`).
  The old reflection nullCheck required *every* property non-null, so adding the optional property
  would have killed token acquisition on any deploy lacking the new key. Regression-pinned.
- **`ResourceUri` set → the token names the target.** v2.0 endpoint,
  `scope = {ResourceUri}/.default`, no `resource` parameter — the minted token's `aud` is the target
  API. The LPII-166 branch sent `resource = {ClientId}` (the caller's own registration) to the v1
  endpoint, minting a self-token. The outgoing form body is asserted by test.
- **`ResourceUri` unset → the legacy v1 self-token path is unchanged**, pinned field-by-field by
  test. This is what makes the PR deployable to every environment with zero config changes.
- **Token response parsed via `JsonDocument`** — the v2.0 endpoint returns `expires_in` as a JSON
  *number*, which the previous `Dictionary<string, string>` deserialization would have thrown on.
  (Latent defect that would only have surfaced on first v2 use.)

**New pin in `AudienceValidationTests`:** with the singular `Audience` key AND
`TokenValidationParameters:ValidAudiences` both present, acceptance is the **union** — so the
activation-ladder config flip should *replace* the singular key with the list, not accumulate. (In
the real flip the singular value is in the list anyway, so even a forgotten removal changes
nothing; the pin proves that too.)

**Docs/config:** landing sequence step 2 + activation rung 2 updated to match the reduced scope and
the union semantics; both `appsettings.Example.json` files gain the optional
`TokenProvider:ResourceUri` key (empty = legacy behaviour).

## Deploy safety

Deployable to any environment with **no config changes and no Entra changes**: without a
`ResourceUri` key the only executed path is the legacy one, pinned byte-for-byte. All new behaviour
is deploy-then-activate via per-environment config, consistent with the landing sequence's standing
rule (every build supports both models until production completes the ladder).

## Tests

- `AccessTokenProviderTests` (new, 8): both endpoint shapes' form bodies, the nullCheck regression,
  required-option gating, caching.
- `AudienceValidationTests`: 6 including the new union pin.
- Full solution builds clean; Core.Tests 101/101, Storage.API.Tests 73/73 (`Category!=Manual`).


brianmac1967 (commented):
## Code review

One real correctness bug, one documentation self-contradiction, and a couple of smaller notes. Most severe first.

### 1. `DigitalPreservation.Core/Web/Headers/AccessTokenProvider.cs:100` — missing `access_token` silently cached as null for up to 56 minutes

A 2xx Entra AD response missing the `access_token` property now silently returns `null` instead of throwing, and `GetAccessToken()` caches that null for up to 56 minutes.

The old code (`Dictionary<string,string>` indexer, `obj?["access_token"]`) threw `KeyNotFoundException` on a non-null body missing that key, failing loudly on the very first call. The new code (`TryGetProperty` → `: null`) swallows this into a `null` return with no log line, and `GetAccessToken()` then caches that null via `memoryCache.Set(key, token, cacheEntryOptions)` regardless of value.

Confirmed downstream impact: `Preservation.Client/MachineAuthTokenInjector.cs:20-27` treats a null token as "skip the Authorization header" and only logs a warning — so machine-to-machine calls to Preservation API would silently go out unauthenticated (401s) for up to 56 minutes instead of failing fast, e.g. if a misconfigured `ResourceUri` or a transient Entra response shape change occurs.

### 2. `docs/rfc-0001-lpii166-comparison.md:210` — doc sections now contradict each other

This PR rewrites §7.1 to say the guarded `PostConfigure` pattern was "resolved... not adopted — unnecessary", but leaves §5.5 (line 210) and §6.3 (line 249) still asserting it "is a genuine improvement the RFC should adopt" / should be hoisted and pinned per §7.1 — a self-contradiction within the same file created by this diff's edit. A reader landing on §5.5 or §6.3 before reaching the now-rewritten §7.1 would conclude the guarded-rollout pattern is still a recommended TODO and re-propose or re-implement code the PR just declared unnecessary.

### Minor

- `AccessTokenProvider.cs:18` — the new testability seam is a bespoke `HttpMessageHandler?` constructor parameter instead of the `IHttpClientFactory`/`AddHttpClient` + named-handler pattern already used everywhere else in the codebase (Preservation.Client, Storage.Client, Storage.API's `ServiceCollectionX.cs`, per ADR-0000) — a second, divergent DI/testing convention for HTTP clients.
- `AccessTokenProvider.cs:19` — the `MemoryCache` key remains the hardcoded literal `"storageApiAccessToken"` even though `ResourceUri` now makes this class resource-generic (used to mint tokens for Storage, or any configured target API). Not a collision bug today, but misleading in logs/diagnostics for a Preservation-API or Pipeline-API self-token.
- `AccessTokenProvider.cs:74` — `GetBearerToken()` re-asserts non-null on `options.ClientId!`/`options.ClientSecret!` with null-forgiving operators instead of taking the already-validated values as parameters, duplicating the checks `GetAccessToken()` just performed — couples its safety to the caller remembering to gate it first.

tomcrane:
One doc-only commit added: the landing sequence gains **Cutting a production release** (the merge of this PR is the natural cut, the pre-cut validation survey, and the full sequence through to the header-deletion release), and the RFC's Phase 4 reversibility line now records that the two behavioural steps ship as configuration flags (#293), so no tagged release can strand production behind the Entra schedule.

tomcrane (commented):
All five taken, in d058dcc.

**1, the cached null - real, and the worst kind of quiet.** `GetBearerToken` now throws on a 2xx missing `access_token` (and on an empty one), matching the old indexer's fail-fast, and a failure of any kind is never cached - the throw happens before the `Set`. The log line carries property *names* only, since a token endpoint body must never reach a log. Three new tests, including the one that matters: after a token-less success response, a second call reaches the endpoint again (`CallCount == 2`), proving nothing was cached. Your downstream trace was right too - `MachineAuthTokenInjector` would have stripped the header silently - and that consequence is now quoted in the test's comment so nobody weakens it back.

**2, the doc contradiction.** §5.5 and §6.3 now record the §7.1 resolution instead of contradicting it - §5.5 distinguishes the rollout *property* (survived, delivered by the stock binding) from the POC's *code* for it (not adopted), and §6.3's hoisting suggestion is struck through with the same resolution note.

**Minors, all taken:** the client comes from `IHttpClientFactory` (named client; `AddHttpClient()` at the three registration sites) rather than the bespoke handler seam, with the tests going through a small factory stub; the cache key names the resource the token is for - noting the cache is per-instance, so that was legibility rather than a collision, as you said; and `GetBearerToken` takes the validated credential values as parameters, no null-forgiving re-assertions.

Whole solution green, 1,074 tests.

brianmac1967 (approved):
LGTM

tomcrane:
## Deep review ahead of the production cut — two blocking defects found in d058dcc, both fixed in cb86e0a

Five parallel review passes were run over this PR and its surroundings (token lifecycle/concurrency, config binding & deploy order, security adversarial, test adequacy, docs/runbook consistency), with every finding verified against the code before acting on it. The uncomfortable headline: **the commit that addressed the last review round (d058dcc, `IHttpClientFactory` for `AccessTokenProvider`) introduced two independent process-killing defects in Preservation.API.** Neither is reachable by the existing CI shape, both are deterministic in production shape, and both are now fixed and regression-pinned.

### Defect 1 — the token mint recursed through the global correlation filter (StackOverflow)

`AddCorrelationIdHeaderPropagation()` registers an `IHttpMessageHandlerBuilderFilter`, which applies to **every** factory-built client. Once the token mint itself went through the factory, the request to Entra's token endpoint acquired the `PropagateCorrelationIdHandler`, whose `SetMachineToken` calls `GetAccessToken()` — mid-mint, with the cache necessarily still empty, so it mints again, synchronously, with no suspension point: unbounded recursion, `StackOverflowException`, dead process. Preservation.API is the only service with both halves registered (filter + real provider, `Program.cs:103`/`:122`), and `StorageImportJobsService` (no `HttpContext`, polls every 60s) triggers it within a minute of every start whenever auth is on. Unauthenticated inbound requests reach the same path via the handler's `else` branch.

**Fix:** the filter now skips the `"EntraTokenEndpoint"` client by name — platform auth headers have no business on a request to Entra.

### Defect 2 — circular DI resolution hangs the process at the first client build

Found not by review but by the new regression test the review prompted: `AccessTokenProvider` constructor-injects `IHttpClientFactory`; `DefaultHttpClientFactory`'s constructor consumes every registered `IHttpMessageHandlerBuilderFilter`; the correlation filter constructor-injected `IAccessTokenProvider`. A closed triangle — and because the factory is registered via a lambda, .NET DI's creation-time cycle detection never sees it. It manifests as unbounded *runtime* resolution recursion (kept alive by DI's own StackGuard): the process hangs at the first factory client build rather than failing with a clean "circular dependency" error. Same conjunction, so again Preservation.API only, again production shape only. The hang reproduced identically in the test host; the dump shows a single thread ~13,000 frames deep in `CallSiteRuntimeResolver`.

**Fix:** the filter resolves `IAccessTokenProvider` lazily, per handler chain, instead of in its constructor. By the time any handler chain is built the factory singleton exists, so the cycle cannot form; services with no provider registered (Storage API, UI) still resolve null exactly as before.

### Why nothing existing could catch either

CI integration tests run `DisableAuth=true`, so the real provider is never registered; the provider's unit tests use a stub `IHttpClientFactory`, which bypasses the `IHttpMessageHandlerBuilderFilter` pipeline entirely. New test `HeaderPropagationFilterTests` builds clients through the **real** factory with the filter and a real provider registered — the Preservation.API shape — and pins both fixes: the token request must carry no platform auth headers, and other factory clients must still get the machine token end-to-end. `MachineAuthTokenInjectorTests` (new) pins the injector's changed failure contract: a throwing provider propagates and nothing is sent unauthenticated.

### Also fixed in this commit

- **`appsettings.Example.json` audience order** (both APIs): was transitional-pair-first, contradicting the landing sequence, all three terraform environments, and `FourEntryShape`. Order is harmless for acceptance but the retirement instruction says "delete `__2`/`__3`" — against the example order that would have deleted the *permanent* audiences. Now permanent-pair-first everywhere.
- **Landing sequence:** the pre-cut `validation-survey` gate now says it ships in #294, which must merge before the gate can run; the full sequence starts with that merge. Phase-4 flags corrected to future tense (#293 is planned, not in this build). Deposit Archiver added to the Phase-2 repoint list (its `OAUTH_AZURE_SECRET` defaults to the UI secret path in every environment — it mints as the UI registration too). Step 3's summary now points at the four-entry shape rather than implying two.
- **RFC §6 Phase 2:** the `TokenProvider` bullet no longer frames option (a)-vs-(b) as an open choice — (a) shipped in this PR with both LPII-166 defects fixed; the phase is config-only now.

### Verified sound during the same review (the answers the release cut needed)

- **Deploy order between the ops terraform and this release is safe in both directions**, source-verified: v1.2.1 pins the same Microsoft.Identity.Web 3.8.3 and the identical `AddMicrosoftIdentityWebApi(GetSection("AzureAd"))` call, and this PR adds zero C# to the audience-validation path.
- The `ValidAudiences` binding is real (direct `Bind()` onto `JwtBearerOptions`; Microsoft.Identity.Web's fallback audience synthesis explicitly steps aside once the list is bound; nothing in this codebase overwrites `TokenValidationParameters`).
- Missing/empty audience config **fails closed** (validates against one ClientId-derived audience), never fails open. Both-keys-present is the union, per ASP.NET Core's own `JwtBearerPostConfigureOptions` — hence the runbook's replace-don't-accumulate rule.
- Issuer/tenant validation is untouched and AND'd with audience; the widened list cannot admit another tenant's tokens, and no authorization logic branches on `aud`.
- The client secret travels only in the POST body; no path logs the token response body (property names only, pinned by test); the diff is clean for a public repo.

### Non-blocking follow-ups worth an issue or two

A startup warning when both `AzureAd:Audience` and the `ValidAudiences` list are populated (the union is silent otherwise); timeout/retry config and a `CancellationToken` for the `"EntraTokenEndpoint"` client (100s framework default today); single-flight on the token cache miss; reading `expires_in` instead of assuming the 56-minute lifetime; disposing the token `HttpResponseMessage`.

Full solution suite green after the fixes: 1,078 passed, 0 failed (4 new) tests.

brianmac1967 (commented):
## Code review (re-review after the update)

Good progress: the previous blocking bug (a null token from a malformed Entra response getting silently cached for up to 56 minutes) is fixed — it now throws instead. This update also ships regression tests for both original LPII-166 defects, a dedicated recursion-guard test, and a dedicated test for the throw-not-cache behavior.

I traced the actual call chains this time (`StorageImportJobsService` → `StorageApiClient.GetImportJobActivities`/`GetImportJobResult`, and `PipelineJobExecutorService` → `PipelineJobRunner.Execute`) rather than assuming the worst case — both already wrap the relevant calls in broad `catch (Exception)`, so a background service crashing the whole host (the natural worry after last review) doesn't hold up here. The residual concern is narrower, below.

### 1. `DigitalPreservation.Core/Web/Headers/PropagateCorrelationIdHandler.cs:102` — mint-failure throw not audited across all HttpClient callers

`SetMachineToken` (used by the global `IHttpMessageHandlerBuilderFilter` on every non-token `HttpClient`) now lets `AccessTokenProvider.GetAccessToken()` throw instead of degrading to a missing header — this widens the blast radius of a mint failure from one caller to every outbound `HttpClient` call made without an `HttpContext`.

During a transient Entra hiccup or a misconfigured/rotated client secret, every hosted-service or request-path `HttpClient` call that reaches this filter now throws (wrapped in `AggregateException` via `.Result`) instead of proceeding unauthenticated as before in some edge cases. The two call sites verified in this review are protected by broad try/catch, but this wasn't exhaustively audited across every `HttpClient` consumer in Preservation.API/Pipeline.API/Deposit.Archiver — an unwrapped call site would propagate the exception further than intended.

### 2. `AccessTokenProvider.cs:110` — non-string `access_token` throws generic, not the intended diagnostic

`JsonElement.GetString()` throws `InvalidOperationException` with .NET's generic message if Entra ever returns an `access_token` property that isn't a JSON string — before the code reaches its own controlled error path (which logs property names and throws a clear message). A malformed body like `{"access_token": 12345}` gets .NET's generic "target element has type Number" message instead of the intended diagnostic, and the token endpoint response is never safely enumerated for the log.

### 3. `AccessTokenProvider.cs:56` — cache key not normalized for a trailing-slash ResourceUri

The `MemoryCache` key for a resource-scoped token uses the raw configured `ResourceUri` without the trailing-slash trim that `GetBearerToken()` applies when building the scope, so a `ResourceUri` with vs without a trailing slash caches under two different keys for what is the same logical resource/token. Low impact since `ResourceUri` is static per process, but a real inconsistency — e.g. a redeploy that changes only the slash would silently double Entra token requests/cache entries instead of sharing the cache.

### 4. `Preservation.API/Program.cs:121` — AccessTokenProvider DI wiring duplicated in three composition roots

The wiring (bind `TokenProvider` options, `AddHttpClient()`, `AddSingleton<IAccessTokenProvider>`) is duplicated verbatim across `Preservation.API/Program.cs`, `Pipeline.API/Program.cs`, and `DigitalPreservation.Deposit.Archiver/Startup.cs` instead of being hoisted into one shared `DigitalPreservation.Core` extension method. Not a runtime bug, but a future change to how `AccessTokenProvider` is registered must be applied identically in three places — exactly the kind of drift the PR's own docs warn about for the `ValidAudiences` secret duplication.

tomcrane (commented):
All four taken, in 74fffe5.

1. **Blast-radius audit (your residual concern) — done, and it comes out clean.** The mint can only run where a request goes out with no Authorization to propagate, so I audited exactly those paths:
   - *Preservation API background:* `StorageImportJobsService` → `GetImportJobActivities`/`GetImportJobResult` — `catch (Exception)` in the client (as you found), so a mint failure (including the `AggregateException` from `.Result`) becomes an error `Result` → logged FAILED read + 30-minute backoff, not a crash.
   - *Preservation API request paths:* the three `StorageApiClient` methods **without** a catch (`TestArchivalGroupPath`, `GetArchivalGroupName`, `ExportArchivalGroup`) are all reached only from authenticated handlers — there the filter propagates the caller's bearer and never mints, so the new throw path is unreachable; any other exception surfaces as a 500 exactly as pre-PR.
   - *Anonymous endpoints* (`iiif-token` GET/POST, media): all behind `EnableIiifMediaEndpoints`, off in every environment and gated on #290 before it is ever on. When it is, their downstream calls are the caught ones (`GetBinaryStream` etc.).
   - *Pipeline API:* the runner's broad catch, as you verified.
   - *Deposit Archiver (Lambda):* `Functions.cs` has no catch, deliberately left that way — an unhandled mint failure fails the invocation and Lambda's retry/DLQ takes over, which is the right fail-closed shape for that host; swallowing it would archive nothing and report success.
   - *Health checks:* DB-context only, no HttpClient — an Entra outage cannot fail `/health` and cycle the containers.
2. **`ValueKind` guard**: `{"access_token": 12345}` now stays on the controlled path (property-names log + clear message) instead of .NET's generic `GetString()` throw. Test added.
3. **Cache key trims the trailing slash**, matching the scope construction — with-slash and without-slash spellings of the same resource now share one cache entry.
4. **Wiring hoisted** to `AddAccessTokenProvider(options)` in Core; the three composition roots keep only their own options construction (they genuinely differ — the Archiver builds options from its secrets provider, the APIs bind `TokenProvider`), so the mechanical trio can no longer drift.

Affected projects green (Core.Tests 122, Pipeline.API.Tests 38, Preservation.API.Tests 197); CI will confirm the full matrix.

brianmac1967:
From Co-pilot:

I found one likely blocking functional gap, plus several reliability and test-quality concerns. The core ResourceUri implementation is otherwise well structured, particularly the v1 compatibility path, controlled token parsing, named HTTP client guard, and audience-binding coverage.

Findings
1. Blocking: Deposit Archiver never receives ResourceUri

Location: DigitalPreservation.Deposit.Archiver/Startup.cs

The Archiver still constructs AccessTokenProviderOptions manually:

var accessTokenProviderOptions = new AccessTokenProviderOptions
{
    ClientId = authProvider.ClientId,
    ClientSecret = authProvider.ClientSecret,
    TenantId = authProvider.TenantId
};


It does not copy or bind ResourceUri.

That conflicts directly with the landing-sequence documentation, which says:

Preservation API, Pipeline API and the Deposit Archiver ... must be repointed here too.

Even if the deployment adds something equivalent to:

TokenProvider__ResourceUri=api://84c62880…


the Archiver will continue taking the legacy v1 self-token path because ResourceUri remains null.

Suggested fix

Ensure the Archiver configuration model exposes ResourceUri, then propagate it:

var accessTokenProviderOptions = new AccessTokenProviderOptions
{
    ClientId = authProvider.ClientId,
    ClientSecret = authProvider.ClientSecret,
    TenantId = authProvider.TenantId,
    ResourceUri = authProvider.ResourceUri
};


Alternatively, bind AccessTokenProviderOptions directly from the appropriate configuration section, which would reduce the chance of future option drift.

A regression test should assert that the Archiver composition root preserves a configured ResourceUri.

2. Medium: cache expiration ignores the token’s actual lifetime

Location: AccessTokenProvider.GetAccessToken()

The token remains cached for a fixed 56 minutes:

var cacheEntryOptions = new MemoryCacheEntryOptions()
    .SetAbsoluteExpiration(TimeSpan.FromMinutes(56))
    .SetSlidingExpiration(TimeSpan.FromMinutes(55));


The new parsing code already has access to expires_in, but does not use it. If Entra returns a token valid for less than 56 minutes, the provider can serve an expired token until the fixed cache entry expires.

This was partly pre-existing behavior, but the PR is substantially changing token acquisition and now explicitly parses v2 responses. This is a good point to stop assuming a one-hour lifetime.

Suggested approach

Return the token and lifetime from GetBearerToken():

private sealed record TokenResult(string AccessToken, int ExpiresIn);


Then cache it with a safety margin based on expires_in, for example:

var lifetime = TimeSpan.FromSeconds(result.ExpiresIn);
var safetyMargin = TimeSpan.FromMinutes(5);
var cacheDuration = lifetime > safetyMargin
    ? lifetime - safetyMargin
    : TimeSpan.FromSeconds(Math.Max(1, lifetime.TotalSeconds / 2));

memoryCache.Set(key, result.AccessToken, cacheDuration);


Also add tests for numeric and malformed expires_in values.

3. Medium: configuration validation accepts whitespace-only values

Location: AccessTokenProvider.GetAccessToken() and GetBearerToken()

The code uses string.IsNullOrEmpty:

if (string.IsNullOrEmpty(options.TenantId)
    || string.IsNullOrEmpty(options.ClientId)
    || string.IsNullOrEmpty(options.ClientSecret))


and:

if (!string.IsNullOrEmpty(resourceUri))


Values such as " " therefore pass validation. A whitespace ResourceUri produces an invalid scope:

 /.default


Whitespace credentials similarly result in a request being sent rather than an immediate configuration failure.

Use string.IsNullOrWhiteSpace and normalize the validated values:

if (string.IsNullOrWhiteSpace(options.TenantId)
    || string.IsNullOrWhiteSpace(options.ClientId)
    || string.IsNullOrWhiteSpace(options.ClientSecret))
{
    logger.LogWarning("AccessTokenProvider options are not configured correctly");
    return null;
}

var resourceUri = string.IsNullOrWhiteSpace(options.ResourceUri)
    ? null
    : options.ResourceUri.Trim().TrimEnd('/');


Tests should include whitespace-only credentials and ResourceUri.

4. Medium: the recursion regression test is designed to crash the test process

Location: HeaderPropagationFilterTests.MintingAToken_DoesNotRecurseThroughTheFilter

The comment explicitly states:

If the guard regresses, the first test dies by stack overflow rather than a failed assert; that is deliberate.

A regression test should report a bounded, understandable failure. A stack overflow can terminate the test host, potentially obscure the responsible test, and prevent the remainder of the suite from running.

The present implementation is correct, but the test failure mode is unnecessarily destructive.

Suggested improvement

Add a bounded recursion detector, such as a custom provider or handler that throws after a second attempted mint. Another option is to test the built handler chain directly and assert that the Entra named client does not contain PropagateCorrelationIdHandler.

The desired failure should look like:

Expected one token endpoint request, but found a second recursive request.


rather than terminating the runner.

5. Low: request and response disposal should be explicit

Location: AccessTokenProvider.GetBearerToken()

Neither HttpRequestMessage nor HttpResponseMessage is disposed:

var response = await client.SendAsync(request);


The factory-created HttpClient can be short-lived safely because the factory owns handler reuse, but messages and responses should still be disposed.

For example:

using var request = new HttpRequestMessage(HttpMethod.Post, endpoint);
request.Content = new FormUrlEncodedContent(collection);

using var response = await client.SendAsync(request);
response.EnsureSuccessStatusCode();

var json = await response.Content.ReadAsStringAsync();


This is not likely to block the PR, but it is worth correcting while the method is being rewritten.

6. Low: simultaneous callers can mint duplicate tokens

The cache flow is:

if (memoryCache.TryGetValue(key, out string? token))
{
    return token;
}

token = await GetBearerToken(...);
memoryCache.Set(key, token, cacheEntryOptions);


Two concurrent requests after startup or expiry can both miss the cache and mint separate tokens. That is not usually functionally incorrect, but it can unnecessarily hit Entra and could worsen an outage or throttling event.

A keyed SemaphoreSlim, cached Lazy<Task<string>>, or similar single-flight mechanism would make token acquisition deterministic.

This is arguably pre-existing behavior, so I would not block this PR on it unless the hosted services can produce a burst immediately after startup.

7. Check the literal source for injected HTML

The supplied diff contains values such as:

$"<a href="https://login.microsoftonline.com/...


and malformed Markdown links containing generated <a> elements.

This appears likely to be formatting introduced while copying the diff into chat. If those tags are literally present in the branch, the C# will not compile and several documentation links are corrupted.

Particularly check:

Token endpoint strings in AccessTokenProvider.cs
URLs in the new tests
Issue #293
Issue #287
PR #294
Positive observations
The named client exclusion is a sensible fix for the IHttpClientFactory handler recursion.
The v1 path is explicitly regression-tested, including its current form body.
The v2 path correctly uses scope without resource.
JsonDocument is appropriate here because mixed JSON value types make Dictionary<string, string> unsuitable.
Missing, empty and non-string access_token responses fail loudly and are not cached.
The tests clearly characterize the union behavior of singular Audience plus ValidAudiences.
The four-entry audience test covers both URI and bare-GUID forms and rejects an unrelated audience.
Centralizing registration in AddAccessTokenProvider reduces composition-root drift, although the Archiver option-population gap shows that option construction also needs centralizing.
Recommendation

Request changes for the Deposit Archiver ResourceUri omission. I would also strongly recommend fixing whitespace validation and making the recursion test fail safely before merging. The cache-lifetime and disposal items could be addressed here or recorded as tightly scoped follow-ups.

tomcrane (commented):
Taken in 619eba5, with calibration notes — this round had one finding that mattered and a tail I took only because they were cheap:

1. **Archiver `ResourceUri` — the real catch, thank you.** `AuthProviderModel` had no `ResourceUri`, so the Archiver's Phase-2 repoint would have needed a *code* change while the landing sequence promises config-only — precisely the stranded-release trap this build exists to close. Property added, propagated in `Startup`; at Phase 2 the key rides in the same `oauth_azure` secret as the credentials. (No automated pin: the Archiver has no test project and a Lambda composition root isn't worth standing one up for; the propagation is a two-line object initializer.)
2. **Whitespace validation** — taken (`IsNullOrWhiteSpace` + trim, two tests), though honestly marginal: a `" "` credential already failed closed, just at Entra instead of at the gate.
3. **Bounded test failure** — fair test-hygiene point, taken with both suggested mechanisms: a chain-shape test asserts the filter handler's absence from the token-endpoint client outright, and the filter tests mint through a reentrancy-detecting provider, so a regressed guard now fails as `"Recursive mint: …"` rather than killing the test host.
4. **Disposal** — taken; one-line `using`s.
5. **`expires_in` cache + single-flight** — deliberately *not* here; split to #295 with the named-client timeout/CancellationToken work, so this PR can converge. Both are pre-existing behaviours, not regressions of this diff.
6. **"Injected HTML"** — a paste artifact of the diff-into-chat route, not in the source (`grep '<a href'` over the branch: zero hits); the solution builds clean.

Core.Tests 125/125. From here I'd apply the stopping rule that a further round needs a finding that changes behaviour in a deployed configuration — this round had exactly one, and it's in.

tomcrane:
Re the Slack question — should the Archiver map `authProvider.ScopeUri` into `ResourceUri` rather than a new `ResourceUri` secret key? **Deliberately not**, for three reasons that are the same reason:

1. **They are different concepts that only coincide under the legacy model.** `ScopeUri` today holds the UI registration's own App ID URI (the legacy shared audience) and its consumer is `AuthTokenInjector` — the *delegated* flow (`ITokenAcquisition`), where it is the scope requested on behalf of a signed-in user. `ResourceUri` is the *machine* mint's target resource (Phase 2: `api://84c62880…`). Overloading one key to drive both mechanisms means a Phase-2 edit to the delegated scope silently repoints the machine mint, and vice versa.
2. **`ScopeUri` is populated in every environment today, and `ResourceUri`'s absence is the off-flag.** The entire release-safety story is that this build is config-inert: under existing config the machine mint takes the byte-identical v1 self-token path, and the v2 path activates only when an operator deliberately sets `ResourceUri` (rung 4). Wiring `ScopeUri` in would flip the Archiver to the v2 endpoint on deploy day — an uncontrolled behaviour change smuggled into the production cut, precisely what the landing sequence forbids. (It would also buy nothing: under today's config the resulting token has the same effective audience the self-token path already produces.)
3. **At Phase 2 the Archiver's repoint is a new secret for its own registration** — `ClientId`/`ClientSecret`/`TenantId`/`ResourceUri`, the same four-key motion as `TokenProvider__ResourceUri` on the APIs. Symmetric, documented in the landing sequence, and `ScopeUri` keeps meaning what it has always meant.

So Copilot's amended framing ("if `ScopeUri` appears likely to be the source, map it") has the inference backwards: the source *must not* be a key that is already populated, or the flag stops being a flag. The `AuthProviderModel.ResourceUri` property added in 619eba5, absent from every current secret, is the intended shape.

brianmac1967 (approved):

PR #285, opened 2026-09-17
Refuse import jobs carrying relative URIs with 400, not an unhandled 500

tomcrane:
Fixes the 500 behind the two failing e2e tests (`archivalGroup.spec.ts`, BagIt and non-BagIt): a posted import job whose `containersToAdd[0].id` is the bare string `'invalid-id'` crashes the API instead of being refused.

## Cause

System.Text.Json deserialises a bare string into a **relative** `Uri`. The #267 validation chokepoint (`ImportJobsController.JobDoesNotBelongToDeposit`) then addresses ids by path — `AbsolutePath`, `GetPathUnderRoot`, `IsBaseOf` — and every one of those throws `InvalidOperationException` on a relative `Uri`. An absolute-but-foreign id gets the intended 400; a relative one became an unhandled 500. The same latent hole existed for a relative `deposit`, a relative `archivalGroup`, and a null or relative binary `origin`.

(Context: before #267 this job was accepted with 201 and failed during execution — that's what the e2e test asserted. Refusing it at submission is #267 working as intended; this PR only fixes the refusal arriving as a crash. The e2e test needs its own update for the new contract, separately.)

## Fix

`IsAbsoluteUri` guards on all four caller-controlled surfaces in `JobDoesNotBelongToDeposit`, each landing in the branch whose message already describes the problem:

- `Deposit` relative → "does not match the Deposit it was submitted to"
- `ArchivalGroup` relative → "does not match the Deposit's Archival Group"
- binary `Origin` null/relative → "is not a child of deposit file location"
- any resource id relative → "must have an id, as an absolute URI" (message extended; the existing no-id test still passes unchanged)

A comment at the top of the chain records that the guards are load-bearing, not tidiness.

## Tests

Five regression tests (each asserting 400 + handler not invoked, via the existing #267 harnesses):

- `A_Relative_Resource_Id_Is_Refused_Rather_Than_Crashing` — the e2e scenario verbatim
- `A_Relative_Binary_Origin_Is_Refused_Rather_Than_Crashing`
- `A_Binary_With_No_Origin_Is_Refused_Rather_Than_Crashing`
- `A_Job_With_A_Relative_Archival_Group_Is_Refused_Rather_Than_Crashing`
- `A_Job_With_A_Relative_Deposit_Is_Refused_Rather_Than_Crashing`

Preservation.API.Tests: 197/197 (`Category!=Manual`).


brianmac1967 (approved):
LGTM

PR #286, opened 2026-09-17
Release the deposit lock when a pipeline job fails before its workspace exists

tomcrane:
Closes the lock leak on the workspace-resolution-failure path that PR #283 introduced alongside its fix: a job whose workspace cannot be resolved (deposit unreachable, or a METS the parser refuses) is correctly recorded as `completedWithErrors` — but that early exit, alone among the handler's exit paths, released no lock. The deposit stayed locked until someone noticed and released it by hand.

Credit: spotted in external review of yesterday's e2e failure. (It turned out not to be that failure's cause — the job there *completed*, and on the success path `Completed` is only returnable after a successful release; the e2e assertion raced the release-then-record transition, fixed in digirati-co-uk/digital-preservation-e2e-tests@a64084f. But the leak is real.)

**Fix:** after recording the failure, fetch the deposit again (this path has no workspace to take it from) and `TryReleaseLock` — best effort, same conventions as `RecordFailureBestEffort`: cancellation propagates, any other exception is logged and the original failure is still returned. When the deposit itself cannot be fetched there is nothing to unlock, and nothing is attempted.

**Tests** (in `ExecutePipelineJobWorkspaceFailureTests`, the #283 suite):
- `A_Resolution_Failure_After_The_Deposit_Was_Fetched_Still_Releases_The_Lock` — the METS-refusal shape: deposit fetched OK, workspace build throws, lock released.
- The deposit-not-found case now also pins that `ReleaseDepositLock` is *not* called.

Pipeline.API.Tests: 35/35.


brianmac1967 (commented):
## Code review

This correctly fixes the deposit-lock-never-released bug (the workspace-resolution failure path in `Handle` previously recorded the job as failed but never called `TryReleaseLock`). The fix itself introduces a subtler issue around ordering, plus some redundant work and a test gap. Most severe first.

### 1. `Pipeline.API/Features/Pipeline/Requests/ExecutePipelineJob.cs:249` — status-then-unlock ordering lets a retry lose its lock

`RecordFailureBestEffort` makes the job's `CompletedWithErrors` status visible to the caller *before* `ReleaseLockBestEffort` releases the deposit lock, and the lock is user-scoped (not job-scoped) and released unconditionally.

Scenario: a user locks a deposit and calls `RunPipeline`; job1 fails at workspace resolution (METS refused). `Handle` records job1 as `CompletedWithErrors` — visible to the caller immediately. The same user, seeing the failure, retries `RunPipeline` right away; `RunPipelineHandler.Handle` allows this because `LockedBy` still equals the caller's identity, and it does not re-lock — job2 starts running against the still-held lock. Meanwhile job1's `ReleaseLockBestEffort` does its own `GetDeposit` + `ReleaseDepositLock` round-trip and unconditionally clears the lock (`DeleteDepositLockHandler` has no job/token check), unlocking the deposit while job2's Brunnhilde scan is still in flight and letting a third party lock/modify it concurrently. This exact interleaving was impossible before this PR, since the lock was never released on this path at all.

### 2. `ExecutePipelineJob.cs:109` — deposit re-fetched even when it's already known missing

`ReleaseLockBestEffort` always re-fetches the deposit via `GetDeposit`, even in the common case where `GetWorkspaceManager` already proved the deposit doesn't exist (deposit-not-found failure) — doubling a guaranteed-to-fail remote call on every such failure.

### 3. `ExecutePipelineJob.cs:105` — discarded deposit forces a re-fetch that can itself fail

The already-fetched `Deposit` from `GetWorkspaceManager` is discarded on workspace-construction failure, forcing `ReleaseLockBestEffort`'s second fetch — which can itself transiently fail right when the release is most needed (immediately after a successful first fetch). If the deposit were propagated through the failed `Result` instead of re-fetched, this avoidable failure mode wouldn't exist.

### 4. `Pipeline.API.Tests/ExecutePipelineJobWorkspaceFailureTests.cs:74` — no test covers `ReleaseLockBestEffort`'s own failure path

The new test only asserts `ReleaseDepositLock` was called, not that `ReleaseLockBestEffort`'s own try/catch correctly swallows a throw from `ReleaseDepositLock`, and no test pins the number of `GetDeposit` calls. A future regression that makes the catch block re-throw, or that silently drops lock release in the METS-refusal case some other way, wouldn't be caught by this suite.

tomcrane:
All four taken, in 1190c14 — responses point by point:

**1. Ordering** — accepted and fixed: the release now runs *before* `RecordFailureBestEffort`, so by the time `completedWithErrors` is visible the lock work is already done, and the retry interleaving cannot arise. Worth being precise about the window that existed: the release began the moment the recording call returned, so the exposure was one HTTP round-trip (~100-300ms) against a retry that must observe the status, re-POST, and have SQS deliver + claim job2 inside it — but the fix costs nothing and the ordering is now also pinned by a FakeItEasy `.Then` assertion, so it can't silently flip back.

**2 + 3. The re-fetch** — accepted, fixed with one change: `ResolveWorkspace` now carries the already-fetched `Deposit` out alongside the failed result (null exactly when the deposit could not be fetched at all). No second `GetDeposit` anywhere: the not-found case no longer repeats a guaranteed failure, and the METS-refusal case cannot lose its release to a transient on a redundant fetch. `GetWorkspaceManager` is untouched for its mid-run call sites.

**4. Test gaps** — accepted: new `A_Lock_Release_Failure_Is_Swallowed_And_The_Failure_Still_Recorded` (throwing `ReleaseDepositLock` → original failure returned, nothing escapes, failure still recorded — which the new ordering makes possible to assert at all); `GetDeposit` call counts pinned to exactly one in both the not-found and METS-refusal tests; the release-then-record order pinned as above. Pipeline.API.Tests 36/36.

brianmac1967 (approved):
LGTM

